PR #43, opened 2015-03-12
fix underscore not showing on Ubuntu

ysangkok:
Without this change, underscore (being under the baseline in Liberation Mono and Courier New), is not showing.


copy:
This also changes the height of the screen, I wonder if that has any unwanted side-effects. Anyway, it also looks better.

ysangkok:
I just noticed I forgot to change it in all the docs/samples. It should be sufficient just to remove it there, since line-height:normal is the default (we need to override line-height:1.5 from somewhere else in the style sheet, so this change is still correct).

PR #75, opened 2015-11-11
Set longer timeout for test cases

aaron-elkins:
Set longer timeout to prevent some test failures on some slower operating systems.
Additionally change .gitignore


copy:
Thanks!

PR #79, opened 2015-12-12
Change emulator.settings to settings

warpcoil:
closure compiler fails.

I'm assuming it should be settings, in either case closure compiler compiles and it works.


copy:
Thanks.

PR #96, opened 2016-04-23
fix typo

CapsAdmin:
I'd just like to interject for a moment. What you’re referring to as LUA, is in fact, Lua.


copy:
Good to know.

Issue #134, opened 2017-03-05
Build fails with latest closure compiler


becosta:
Following the steps in the Readme file, i had to manually download the [closure compiler v20170124](http://dl.google.com/closure-compiler/compiler-20170124.zip) to build libv86.js.

Using any latest release or the latest compiler gave me the following [errors](http://pastebin.com/fQFHzQhs).

PR #135, opened 2017-03-05
copy/v86#134 - Changes closure compiler url from latest to v20170124

becosta:
A quick and dirty workaround which allows the project to build until a real solution has been implemented.


copy:
Thanks for the PR. I've implemented a proper fix in  cf55f14.

PR #149, opened 2017-04-27
fix unzip command in readme

vdloo:
Looks like the file is now called `closure-compiler-v20170423.jar`
instead of `compiler.jar`. This appears to already have been corrected in
the Makefile [here](https://github.com/vdloo/v86/blob/master/Makefile#L142).


copy:
Thanks!

Issue #41, opened 2015-02-25
Provide bootable image with 9p support


ysangkok:
I want to use the 9p filesystem and inject files into it. To inject files, I need to get the emulator instance. The Arch Linux demo is, AFAIK, the only one that features 9p support, however, it only works with the production build, which uses the closure compiler, which means I cannot get the emulator instance (since I don't know how it was mangled).

I can also build the image myself, maybe you have recommendations on how to do this? I imagine that I only need to build a new kernel, since I can use the userland from ttylinux or the linux26 image.

copy:
The easiest way to do this is to build everything from scratch using [busybox](http://www.busybox.net/). This will include building userland, but it also handles making the boot loader and the disk image.

When you're done and the image is reasonably small, upload it somewhere and I'll make it available on the website.

ysangkok:
I found the kernel image in use in the Arch Linux demo: http://104.131.53.7:8086/arch/var/cache/pacman/pkg/linux-3.17.6-1-i686.pkg.tar.xz

I put it in on a file system and loaded it with GRUB (from a floppy). I used the flags from /proc/cmdline: `root=host9p rootfstype=9p rootflags=trans=virtio`. However, it is unable to mount the root file system. The backtrace is too large to be seen, but on kernels I compile myself (I tried 3.2 and 3.17.8), I'm probably getting the same problem: `no channels available`. Do you have any ideas why 9p is not working for me?

copy:
Do you want to boot from the 9p filesystem or just have a separate 9p partition?

In either case, you'll need to specify a JSON file with the filesystem infos (created using [this](https://github.com/copy/fs2json) script) and a path where the files are stored. See [filesystem.md](https://github.com/copy/v86/blob/master/docs/filesystem.md). If you want to boot from 9p, you put the root file system in there. 

The default front end only creates a 9p interface for the Archlinux profile at this point. If you're interested, I can make these specifiable via URL for custom profiles.

ysangkok:
I want to Linux to use the 9p filesystem as its root filesystem, this is why I tried specifying `root=host9p`.

I'm trying to mount the same filesystem, I just copied the JSON file to my local server (and correctly passed it to V86Starter). If the filesystem mounts, it may not load files due to CORS. But it's not even mounting yet. Once I get that working, I'll start playing with custom filesystems. I guess something is problematic with my boot process, this is what I'm trying to figure out. But I don't see how I'm doing it differently from you, though I must be doing some differently.

[...]

PR #153, opened 2017-04-30
update linux 9p image mount hook doc

vdloo:
The "$root" should always be host9p if you want to boot with the root fs mounted over 9p, otherwise the initcpio hook will try to mount /dev/sda1 (or whatever you have configured) over /new_root as a 9p fs (which it maybe isn't) and the system will complain with "no channels available" and drop you in the emergency shell without any real clues about what is going on. I think you can probably achieve this with "$root" as well by grub installing 'host9p' as the root fs or doing what @ysangkok does https://github.com/copy/v86/issues/41 but I think that in the case of the specific linux-9p-image doc it makes more sense to hardcode it in  the hook since it should never be anything other than host9p. Took me a while to figure out this was what was causing my machine to fail to boot since the 'no channels available' error doesn't really give you much to go on.

Before:
before

After:
after

### DIFF
--- a/docs/linux-9p-image.md
+++ b/docs/linux-9p-image.md
@@ -31,7 +31,7 @@ run_hook() {
 
 mount_9p_root() {
     msg ":: mounting '$root' on real root (9p)"
-    if ! mount -t 9p "$root" "$1"; then
+    if ! mount -t 9p host9p "$1"; then
         echo "You are now being dropped into an emergency shell."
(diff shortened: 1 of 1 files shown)


copy:
Thanks, indeed the documentation was missing the kernel command line.

Note that `host9p` is just a name we've hardcoded, it could be variable in the future. But for now, this is fine.

PR #154, opened 2017-05-01
Improve fprem instruction implementation to update FPU status word's condition codes

AmaanC:
This PR only aims to allow the `fprem` instruction to update the condition codes in the FPU status word. This allows the `fprem` instruction to actually complete instead of looping waiting for `C2` to be reset to `0`.

**Note:** Doesn't take into account exceptions, nor does it emulate the way `fprem` actually calculates the partial remainder (by iteration).

### DIFF
--- a/src/fpu.js
+++ b/src/fpu.js
@@ -876,26 +876,20 @@ FPU.prototype.op_D9_reg = function(imm8)
                 case 0:
                     // fprem
                     var st1 = this.get_sti(1);
-                    // Math.trunc truncates towards 0 as we want
                     var fprem_quotient = Math.trunc(st0 / st1);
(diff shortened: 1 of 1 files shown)


AmaanC:
Could use something like `this.status_word |= (fprem_quotient >> 0 & 1) * FPU_C1;` instead of the branching. Let me know if you prefer that!

copy:
Looks good. Just a minor thing: Remove the comments, the code is rather straightforward and more readable without them.

> Could use something like this.status_word |= (fprem_quotient >> 0 & 1) * FPU_C1; (replacing 0 with the bit number) instead of the branching. Let me know if you prefer that!

No, it's good as is.

AmaanC:
Done :+1:

PR #155, opened 2017-05-06
expand Archlinux doc with scripted install example

vdloo:


copy:
Very nice work, thanks!

PR #168, opened 2017-09-10
Fix docs

BenNottelling:
It seems at some point the way API calls are made got change and left these docs outdated:
`options.bios =` becomes `bios: `

---
This change is [Reviewable](https://reviewable.io/reviews/copy/v86/168)


copy:
Thanks!

PR #176, opened 2017-12-10
Emulate PC Speakers and Sound Blaster 16 Card

ErnWong:
### This PR contains the following goodness

- DMA supports all 8 channels.
- DMA now supports reading the channel registers. According to osdev.org, these IO ports should be write only, but for some reason the logs show that the Windows Sound Blaster driver uses it to keep track of the progress - see https://gist.github.com/ErnWong/e3edd265c7dbce8d9b48f41661854f3f for example.
- DMA `do_write` now supports 16 bit incomplete transfers.
- Emulate PC Speaker sounds by listening to changes in PIT registers. The PIT sends `pcspeaker-enable` and `pcspeaker-update` bus messages to the `SpeakerAdapter`.
- Emulate SB16. The win98 stock driver is able to recognise this device. Direct-mode/DMA 8bit/16bit mono/stereo transfers are implemented.
- Poor man's method of sample rate conversion (currently leads to pitch distortion).
- SB16 auto-initialize modes and double buffering appears to work fine.

### Video Demos

- SB16 demo: https://drive.google.com/file/d/1FSt14j7ayKnaT7AS8lF_4S1i4ZX9GlEk/view (note: the demo also includes VGA modifications not included with this PR)
- PC Speaker demo: https://drive.google.com/file/d/11Udg1PGL6lPNjCgu4ApKlSwjs0v1sZAC/view

### Things that aren't implemented yet for SB16

- Mixer Chips
- ADPCM transfer modes
- FM Synthesizers
- MIDI (MPU-401 UART Mode and SB-MIDI Mode)
- Input device (like microphones and stuff)
- Windows Media Player only works with audio hardware acceleration disabled

### Caution - Allergen Information

- Because new entries are now added to the state save file, the `STATE_VERSION` is bumped from 3 to 4. This PR will currently break existing state files.
- `DMA.prototype.do_write` was rewritten to only transfer the given `len` parameter in order for double buffering to work. This seems to be undoing one of your previous commits: https://github.com/copy/v86/commit/8996c72eb2d0de0153fff687bad5f3ff06dc57be Maybe I've misunderstood how `do_write` is supposed to work?
- I didn't modify the `DMA.prototype.do_read` function the same way I modified `do_write`, because I haven't fully understood it yet.
- Yeah, pay close attention to how I've modified `dma.js`. I have heavily butchered the original code... you might not like it :stuck_out_tongue_winking_eye:

### DIFF
--- a/src/browser/speaker.js
+++ b/src/browser/speaker.js
@@ -9,7 +9,7 @@
  */
 function SpeakerAdapter(bus)
 {
-    if (typeof window === "undefined")
+    if(typeof window === "undefined")
     {
         return;
     }
@@ -86,7 +86,7 @@ SpeakerAdapter.prototype.beep_update = function()
 
 SpeakerAdapter.prototype.dac_process = function(event)
 {
-    this.bus.send("speaker-samplerate", this.audio_context.sampleRate);
+    this.bus.send("speaker-tell-samplerate", this.audio_context.sampleRate);
     this.bus.send("speaker-process", event);
     if(this.debug_dac)
     {
--- a/src/browser/speaker.js
+++ b/src/browser/speaker.js
@@ -34,7 +34,7 @@ function SpeakerAdapter(bus)
     this.beep_frequency = 440;
     this.pit_enabled = false;
 
-    this.dac_processor = this.audio_context.createScriptProcessor(DMA_BLOCK_SAMPLES, 0, 2);
+    this.dac_processor = this.audio_context.createScriptProcessor(2048, 0, 2);
     this.dac_processor.onaudioprocess = this.dac_process.bind(this);
     this.dac_processor.connect(this.audio_context.destination);
 
--- a/src/browser/speaker.js
+++ b/src/browser/speaker.js
@@ -96,8 +96,11 @@ SpeakerAdapter.prototype.beep_update = function()
 
 SpeakerAdapter.prototype.dac_process = function(event)
 {
+    var out = event.outputBuffer;
+    var out0 = out.getChannelData(0);
+    var out1 = out.getChannelData(1);
     this.bus.send("speaker-tell-samplerate", this.audio_context.sampleRate);
-    this.bus.send("speaker-process", event);
+    this.bus.send("speaker-process", [out0, out1, out.length], [out0.buffer, out1.buffer]);
 
     if(DEBUG)
     {
--- a/src/browser/speaker.js
+++ b/src/browser/speaker.js
@@ -56,16 +56,19 @@ function SpeakerAdapter(bus)
         this.beep_update();
     }, this);
 
-    this.debug_dac = false;
-    this.debug_dac_out = [];
-    window["speaker_debug_dac_out"] = this.debug_dac_out;
-    window["speaker_debug_start"] = () =>
+    if(DEBUG)
     {
-        this.debug_dac = true;
-        setTimeout(() =>
+        this.debug_dac = false;
+        this.debug_dac_out = [];
+        window["speaker_debug_dac_out"] = this.debug_dac_out;
+        window["speaker_debug_start"] = () =>
         {
-            this.debug_dac = false;
-        },250);
+            this.debug_dac = true;
+            setTimeout(() =>
+            {
+                this.debug_dac = false;
+            },250);
+        }
     }
 }
 
@@ -93,8 +96,12 @@ SpeakerAdapter.prototype.dac_process = function(event)
 {
     this.bus.send("speaker-tell-samplerate", this.audio_context.sampleRate);
     this.bus.send("speaker-process", event);
-    if(this.debug_dac)
+
+    if(DEBUG)
     {
-        this.debug_dac_out.push(event.outputBuffer.getChannelData(0).slice());
+        if(this.debug_dac)
+        {
+            this.debug_dac_out.push(event.outputBuffer.getChannelData(0).slice());
+        }
     }
 }
--- a/src/browser/speaker.js
+++ b/src/browser/speaker.js
@@ -50,6 +50,18 @@ function SpeakerAdapter(bus)
         this.beep_frequency = OSCILLATOR_FREQ * 1000 / pit.counter_reload[2];
         this.beep_update();
     }, this);
+
+    this.debug_dac = false;
+    this.debug_dac_out = [];
+    window["speaker_debug_dac_out"] = this.debug_dac_out;
+    window["speaker_debug_start"] = () =>
+    {
+        this.debug_dac = true;
+        setTimeout(() =>
+        {
+            this.debug_dac = false;
+        },250);
+    }
 }
 
 SpeakerAdapter.prototype.beep_update = function()
@@ -76,4 +88,8 @@ SpeakerAdapter.prototype.dac_process = function(event)
 {
     this.bus.send("speaker-samplerate", this.audio_context.sampleRate);
     this.bus.send("speaker-process", event);
+    if(this.debug_dac)
+    {
+        this.debug_dac_out.push(event.outputBuffer.getChannelData(0).slice());
+    }
 }
--- a/src/browser/speaker.js
+++ b/src/browser/speaker.js
@@ -66,6 +66,13 @@ function SpeakerAdapter(bus)
         this.dac_buffer1 = data[1];
     }, this);
 
+    bus.register("speaker-request-samplerate", function()
+    {
+        bus.send("speaker-tell-samplerate", this.audio_context.sampleRate);
+    }, this);
+
+    bus.send("speaker-tell-samplerate", this.audio_context.sampleRate);
+
     if(DEBUG)
     {
         this.debug_dac = false;
@@ -108,7 +115,6 @@ SpeakerAdapter.prototype.dac_process = function(event)
     out.copyToChannel(this.dac_buffer0, 0);
     out.copyToChannel(this.dac_buffer1, 1);
 
-    this.bus.send("speaker-tell-samplerate", this.audio_context.sampleRate);
     this.bus.send("speaker-request-data", out.length);
 
     if(DEBUG)
--- a/src/browser/speaker.js
+++ b/src/browser/speaker.js
@@ -42,6 +42,8 @@ function SpeakerAdapter(bus)
     this.dac_processor = this.audio_context.createScriptProcessor(2048, 0, 2);
     this.dac_processor.onaudioprocess = this.dac_process.bind(this);
     this.dac_processor.connect(this.audio_context.destination);
+    this.dac_buffer0 = new Float32Array(this.dac_processor.bufferSize);
+    this.dac_buffer1 = new Float32Array(this.dac_processor.bufferSize);
 
     bus.register("pcspeaker-enable", function(yesplease)
     {
@@ -56,6 +58,12 @@ function SpeakerAdapter(bus)
         this.pit_enabled = counter_mode == 3;
         this.beep_frequency = OSCILLATOR_FREQ * 1000 / counter_reload;
         this.beep_update();
+    }, this);
+
+    bus.register("speaker-update-data", function(data)
+    {
+        this.dac_buffer0 = data[0];
+        this.dac_buffer1 = data[1];
     }, this);
 
     if(DEBUG)
@@ -97,10 +105,11 @@ SpeakerAdapter.prototype.beep_update = function()
 SpeakerAdapter.prototype.dac_process = function(event)
 {
     var out = event.outputBuffer;
-    var out0 = out.getChannelData(0);
-    var out1 = out.getChannelData(1);
+    out.copyToChannel(this.dac_buffer0, 0);
+    out.copyToChannel(this.dac_buffer1, 1);
+
     this.bus.send("speaker-tell-samplerate", this.audio_context.sampleRate);
-    this.bus.send("speaker-process", [out0, out1, out.length], [out0.buffer, out1.buffer]);
+    this.bus.send("speaker-request-data", out.length);
 
     if(DEBUG)
     {
--- a/src/browser/speaker.js
+++ b/src/browser/speaker.js
@@ -34,7 +34,7 @@ function SpeakerAdapter(bus)
     this.beep_frequency = 440;
     this.pit_enabled = false;
 
-    this.dac_processor = this.audio_context.createScriptProcessor(DSP_DACSIZE, 0, 2);
+    this.dac_processor = this.audio_context.createScriptProcessor(1024, 0, 2);
     this.dac_processor.onaudioprocess = this.dac_process.bind(this);
     this.dac_processor.connect(this.audio_context.destination);
 
--- a/src/browser/speaker.js
+++ b/src/browser/speaker.js
@@ -90,7 +90,7 @@ SpeakerAdapter.prototype.beep_update = function()
         this.beep_gain.gain.setValueAtTime(0, current_time);
         this.beep_playing = false;
     }
-}
+};
 
 SpeakerAdapter.prototype.dac_process = function(event)
 {
@@ -104,4 +104,4 @@ SpeakerAdapter.prototype.dac_process = function(event)
             this.debug_dac_out.push(event.outputBuffer.getChannelData(0).slice());
         }
     }
-}
+};
--- a/src/browser/speaker.js
+++ b/src/browser/speaker.js
@@ -49,10 +49,12 @@ function SpeakerAdapter(bus)
         this.beep_update();
     }, this);
 
-    bus.register("pcspeaker-update", function(pit)
+    bus.register("pcspeaker-update", function(data)
     {
-        this.pit_enabled = pit.counter_mode[2] == 3;
-        this.beep_frequency = OSCILLATOR_FREQ * 1000 / pit.counter_reload[2];
+        var counter_mode = data[0];
+        var counter_reload = data[1];
+        this.pit_enabled = counter_mode == 3;
+        this.beep_frequency = OSCILLATOR_FREQ * 1000 / counter_reload;
         this.beep_update();
     }, this);
 
--- a/src/browser/speaker.js
+++ b/src/browser/speaker.js
@@ -13,6 +13,11 @@ function SpeakerAdapter(bus)
     {
         return;
     }
+    if(!window.AudioContext && !window.webkitAudioContext)
+    {
+        console.warn("Web browser doesn't support Web Audio API");
+        return;
+    }
 
     /** @const @type {BusConnector} */
     this.bus = bus;
--- a/src/browser/speaker.js
+++ b/src/browser/speaker.js
@@ -29,7 +29,7 @@ function SpeakerAdapter(bus)
     this.beep_gain.connect(this.audio_context.destination);
 
     this.beep_oscillator = this.audio_context.createOscillator();
-    this.beep_oscillator.type = 'square';
+    this.beep_oscillator.type = "square";
     this.beep_oscillator.frequency.value = 440;
     this.beep_oscillator.connect(this.beep_gain);
     this.beep_oscillator.start();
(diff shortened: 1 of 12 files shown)


ZaDarkSide:
Amazing work!

src/pit.js
```diff
@@ -230,6 +234,8 @@ PIT.prototype.counter_write = function(i, value)
     {
         this.counter_next_low[i] ^= 1;
     }
+
+    this.bus.send("pcspeaker-update", this);
```
copy: Send only the values that are needed, not the whole object. The bus is designed for messages that can be send to a web worker, which works for arrays but won't for objects (at least not if the store functions).

src/pit.js
```diff
@@ -305,4 +311,6 @@ PIT.prototype.port43_write = function(reg_byte)
 
     this.counter_mode[i] = mode;
     this.counter_read_mode[i] = read_mode;
+
+    this.bus.send("pcspeaker-update", this);
```
copy: Same as above.

src/sb16.js
```diff
@@ -0,0 +1,1607 @@
...
+
+// IRQ Select.
+register_mixer_read(0x80, function()
+{
+    switch(this.irq)
+    {
+        case 2: return 0x1;
+        case 5: return 0x2;
+        case 7: return 0x4;
+        case 10: return 0x8;
```
copy: Can you make constant for the irqs?

src/sb16.js
```diff
@@ -0,0 +1,1607 @@
...
+});
+
+
+
+//
+// Mixer Handlers
+//
+
+
+
```
copy: Minor: Don't use more than 2 newlines.

src/sb16.js
```diff
@@ -0,0 +1,1607 @@
...
+    this.asp_registers.fill(0);
+    this.asp_registers[5] = 0x01;
+    this.asp_registers[9] = 0xF8;
+}
+
+SB16.prototype.get_state = function()
+{
+    var state = [];
+
+    // state[ 0] = this.write_buffer              ;
```
copy: Please no spaces before the semicolon :-)

src/sb16.js
```diff
@@ -0,0 +1,1607 @@
...
+    for(var i = 0; i < addresses.length; i++)
+    {
+        fm_handlers[addresses[i]] = handler;
+    }
+}
+
+function between(start, end)
+{
+    if(start === end) return [start];
+    return [start].concat(between(start + 1, end));
```
copy: While I like functional programming, this style is massively inefficient in JS and not even more readable, so just use a for loop.

src/sb16.js
```diff
@@ -0,0 +1,1607 @@
...
+/** @const */ SB_IRQ_MPU   = 0x4;
+
+
+// Probably less efficient, but it's more maintainable, instead
+// of having a single large unorganised and decoupled table.
+var DSP_command_sizes = new Uint8Array(256);
+var DSP_command_handlers = [];
+var mixer_read_handlers = [];
+var mixer_write_handlers = [];
+var fm_handlers = [];
```
copy: These should be capitalised if they are constants.

src/sb16.js
```diff
@@ -0,0 +1,1607 @@
...
+
+SB16.prototype.get_channel_count = function()
+{
+    return this.dsp_stereo? 2 : 1;
+}
+
+SB16.prototype.dma_transfer_size_set = function()
+{
+    this.dma_sample_count = 1
+        + (this.write_buffer.shift() << 0)
```
copy: Super-nitpick, but the operator should be at the end of the previous line to make it more obvious that the line continues. Also, this code is not easily verifiable to be correct in the presence of automatic semicolon insertion.

src/sb16.js
```diff
@@ -0,0 +1,1607 @@
...
+    state[34] = this.audio_samplerate          ;
+
+    return state;
+};
+
+SB16.prototype.set_state = function(state)
+{
+    // this.write_buffer               = state[ 0];
+    // this.read_buffer                = state[ 1];
+    this.read_buffer_lastvalue      = state[ 2];
```
copy: Follow the general style of code in this project: Only a single space before `=`.

src/sb16.js
```diff
@@ -0,0 +1,1607 @@
...
+    dbg_log("e2 dma identification. (not understood) input: " + input, LOG_SB16);
+
+    this.e2_value += (input & 0x01) * SB_E2_SIGNS[(this.e2_count + 0) % 4];
+    this.e2_value += (input & 0x02) * SB_E2_SIGNS[(this.e2_count + 2) % 4];
+    this.e2_value += (input & 0x04) * SB_E2_SIGNS[(this.e2_count + 1) % 4];
+    this.e2_value += (input & 0x08) * SB_E2_SIGNS[(this.e2_count + 3) % 4];
+    this.e2_value += (input & 0x10) * SB_E2_SIGNS[(this.e2_count + 2) % 4];
+    this.e2_value += (input & 0x20) * SB_E2_SIGNS[(this.e2_count + 0) % 4];
+    this.e2_value += (input & 0x40) * SB_E2_SIGNS[(this.e2_count + 3) % 4];
+    this.e2_value += (input & 0x80) * SB_E2_SIGNS[(this.e2_count + 1) % 4];
```
copy: Use bitwise operators for the `*` and `%`.

src/sb16.js
```diff
@@ -0,0 +1,1607 @@
...
+
+    this.e2_value += SB_E2_EXTRA[this.e2_count % 4];
+
+    this.e2_count++;
+    this.e2_value %= 256;
+
+    this.dma.on_unmask(this.dma_channel_8bit, () =>
```
copy: Functions cannot be stored in the state since they can't be serialised. Instead of storing `on_unmask` a function in `on_unmask`, can we store the variables that are captured in the closure of this inline function, and move it to the prototype of SB16?

src/sb16.js
```diff
@@ -0,0 +1,1607 @@
...
+
+// Get DSP version number.
+register_dsp_command([0xE1], 0, function()
+{
+    this.read_buffer.clear();
+    this.read_buffer.push(4);
+    this.read_buffer.push(5);
+});
+
+// DMA identification - based completely off dosbox
```
copy: We can't copy or port code from dosbox, because their license (GPL) isn't compatible with ours (BSD2). Please delete this code and write it again from scratch.

copy:
Very impressive work, I appreciate your contribution.

> DMA.prototype.do_write was rewritten to only transfer the given len parameter in order for double buffering to work. This seems to be undoing one of your previous commits: 8996c72 Maybe I've misunderstood how do_write is supposed to work?
> I didn't modify the DMA.prototype.do_read function the same way I modified do_write, because I haven't fully understood it yet.
> Yeah, pay close attention to how I've modified dma.js. I have heavily butchered the original code... you might not like it 😜

I wrote this code a long time ago, never worked on again and I don't recall the details of the dma stuff. It's only used by the floppy disk controller though, so I'll test my floppy disk images and if they work it's good enough. Don't be scared to butcher it more even you need.

I gave you an initial review, but it looks mostly good. I haven't tested it yet, but the videos already look promising. Are the remaining bits a work-in-progress or can I merge it with the current feature set?

src/sb16.js
ErnWong: :astonished:

src/sb16.js
ErnWong: Yikes. I'll be careful with licenses in the future. My implementation of this DMA identification routine is already a bit different to the one dosbox's uses, but I'll remove this routine entirely because:
1. I haven't encountered a situation that uses this routine yet.
2. I can't find any documentation about this routine.
3. I don't really understand this routine myself so I can't rewrite it from scratch anyway :stuck_out_tongue:

src/sb16.js
ErnWong: Like a constant array that maps the irq values directly to the return values? In that case, would you like me to do the same for the subsequent functions up to line 1235? (DMA Select and IRQ status)

copy: No, I was thinking of using `case SB16_INT_FOO:` instead of the magical values.

What you're suggesting might also be a good idea, but I wouldn't do that if the array is sparse.

ErnWong: Ah I see what you mean.

ErnWong: Hmmm, the best that I can come up with is `SB16_IRQ2 = 2`, `SB16_IRQ5 = 5`, etc, because the SB16 device only uses one IRQ line for everything. In the perspective of the device, these different IRQ lines have no special meaning.

ErnWong:
Thanks for reviewing this!

Saving and restoring the state while music is playing is buggy, so I'll try and fix that first.

But after that, if you're happy with it, I'm happy for you to merge it with the current feature set. I'll be working on the remaining features later on, but those probably won't be ready anytime soon.

ErnWong:
That last commit seems to have resolved the bug related to saving and restoring states while music is playing.

So yes, thank you very much. I'm happy for it to be merged at its current state if you're happy. I'm also happy to fix anything you want me to fix.

src/browser/speaker.js
```diff
@@ -0,0 +1,109 @@
...
+    {
+        this.beep_gain.gain.setValueAtTime(0, current_time);
+        this.beep_playing = false;
+    }
+};
+
+SpeakerAdapter.prototype.dac_process = function(event)
+{
+    this.bus.send("speaker-tell-samplerate", this.audio_context.sampleRate);
+    this.bus.send("speaker-process", event);
```
ErnWong: Now I think about it, I think the `speaker-process` bus messages needs to be rewritten with web workers in mind. Instead of passing the entire `event` object for the sb16 device to modify, is it better for the sb16 device to message back to the speaker adaptor with the data?

I'm not too sure. If we look at `vga.js` and `screen.js`, `vga.js` seems to be receiving the destination buffer (on which vga.js modifies) via the `screen-tell-buffer` bus message: https://github.com/copy/v86/blob/master/src/vga.js#L258-L261

If web workers were to be used, would modifying the buffer within `vga.js` still get updated in `screen.js`?

copy: > If web workers were to be used, would modifying the buffer within vga.js still get updated in screen.js?

Not exactly. If the third argument if omitted, the buffer is copied, so this won't work for us. If the underlying buffer is passed as the third argument (see transfer in [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage)), the buffer becomes unusable on the sender, but is transferred without copying.

For vga this works, because you can still use `putImageData` even if the image data buffer has been send to another worker.

I think in this case the same trick as for vga might work. If not, we'll need to send the data from sb16 to the speaker.

But to honest, I don't mind if this code is broken in web workers. I haven't tested the web worker in a long time, so it's probably broken anyway. And as I mentioned on gitter, if I ever try web workers again, I'd probably just put the CPU in the worker and let everything else run in the main thread.

ErnWong: Gotcha. I was wondering why there was a third unused argument in `BusConnector.prototype.send`, but now it makes sense.

Yup, I'll try to use the same technique of transferring the buffer ownership in sb16, and see if it works.
I'll still try and make sb16 bus calls compatible with web workers, so if we ever change our minds and try putting other devices in the worker, there's one less obstacle to go through.

ErnWong:
Alrighty. When you're ready `^_^`

cyberplant:
Guys, I don't know the details about what you're talking, but I must say that I really enjoy reading your conversations! It reminds me of my first PC when I was a teenager, playing with pascal on direct hardware.. my first sound card, etc!

Great work!!
Hope you keep adding features to the project!

src/pit.js
```diff
@@ -132,11 +138,9 @@ PIT.prototype.get_counter_value = function(i, now)
...
     dbg_log("diff=" + diff + " dticks=" + diff_in_ticks + " value=" + value + " reload=" + this.counter_reload[i], LOG_PIT);
 
-    if(value < 0)
-    {
-        var reload = this.counter_reload[i];
-        value = value % reload + reload;
-    }
+    // could be too large after restore_state
+    var reload = this.counter_reload[i];
+    value = (value % reload + reload) % reload;
```
copy: We should emit a warning when the value is larger than the reload, and we can also get rid of the double-modulus.

```javascript
    var reload = this.counter_reload[i];

    if(value >= reload)
    {
        dbg_log("Warning: Counter" + i + " value " + value  + " is larger than reload " + reload, LOG_PIT);
        value %= reload;
    }
    else if(value < 0)
    {
        value = value % reload + reload;
    }
```

src/lib.js
```diff
@@ -265,6 +265,103 @@ function ByteQueue(size)
...
+ * @constructor
+ *
+ * Queue wrapper around Float32Array
+ * Used by devices such as the sound blaster sound card
+ */
+function FloatQueue(size)
+{
+    var data = new Float32Array(size),
+        start,
+        end;
```
copy: For new code, use prototype and properties instead of closures and variables (like the `CircularQueue` below).

src/lib.js
```diff
@@ -265,6 +265,103 @@ function ByteQueue(size)
...
+    {
+        var slice = new Float32Array(count);
+
+        if(count > this.length)
+        {
+            count = this.length;
+        }
+        var slice_end = start + count;
+
+        var partial = data.slice(start, slice_end);
```
copy: Don't use `.slice`, it makes an unnecessary copy. Instead, use `.subarray` (also below).

src/sb16.js
```diff
@@ -0,0 +1,1620 @@
...
+    var amplitude = this.dsp_16bit? 32767.5 : 127.5;
+    var offset = this.dsp_signed? 0 : -1;
+    var repeats = (this.dsp_stereo? 1 : 2) * this.dac_rate_ratio;
+
+    var buffer;
+    if(this.dsp_16bit)
+    {
+        buffer = this.dsp_signed ?
+            this.dma_buffer_int16 :
+            this.dma_buffer_uint16;
```
copy: This can go on one line (line length is 100).

src/sb16.js
```diff
@@ -0,0 +1,1620 @@
...
+        // Keep transfering until dac_buffer contains enough data.
+        setTimeout(() => { this.dma_transfer_next(); }, 0);
+    });
+};
+
+SB16.prototype.dma_to_dac = function(sample_count)
+{
+    var amplitude = this.dsp_16bit? 32767.5 : 127.5;
+    var offset = this.dsp_signed? 0 : -1;
+    var repeats = (this.dsp_stereo? 1 : 2) * this.dac_rate_ratio;
```
copy: Space before `?`.

src/sb16.js
```diff
@@ -0,0 +1,1620 @@
...
+            this.dma_buffer_uint8;
+    }
+
+    var channel = 0;
+    for(var i = 0; i < sample_count; i++)
+    {
+        for(var j = 0; j < repeats; j++)
+        {
+            this.dac_buffers[channel].push(audio_normalize(buffer[i], amplitude, offset));
+            channel ^= 1;
```
copy: I believe `audio_normalize(buffer[i], amplitude, offset)` is a constant and should be moved out of the loop.

src/sb16.js
```diff
@@ -0,0 +1,1620 @@
...
+};
+
+SB16.prototype.audio_send = function(size)
+{
+    this.dac_process_samples = size;
+
+    if(this.cpu_paused || this.dma_paused)
+    {
+        // Silence when paused.
+        var silence = new Float32Array(size);
```
copy: I don't think should continuously allocate when the emulator is paused. Can we instead send a different message, which, when received, sets `dac_buffer0` and `dac_buffer1` to null, and we check for this case in the processing function?

src/sb16.js
```diff
@@ -0,0 +1,1620 @@
...
+    {
+        dbg_log("dac_buffer contains only " +
+            (this.dac_buffers[0].length / 2) +
+            " samples out of " + this.dac_process_samples + " needed", LOG_SB16);
+    }
+
+    this.bus.send("speaker-update-data",
+    [
+        this.dac_buffers[0].shift_block(size),
+        this.dac_buffers[1].shift_block(size),
```
copy: The underlying buffer of these should be passed as the third argument to avoid another copy.

src/browser/speaker.js
```diff
@@ -0,0 +1,127 @@
...
+    this.audio_context = new (window.AudioContext || window.webkitAudioContext)();
+
+    this.beep_gain = this.audio_context.createGain();
+    this.beep_gain.gain.value = 0;
+    this.beep_gain.connect(this.audio_context.destination);
+
+    this.beep_oscillator = this.audio_context.createOscillator();
+    this.beep_oscillator.type = 'square';
+    this.beep_oscillator.frequency.value = 440;
+    this.beep_oscillator.connect(this.beep_gain);
```
copy: I'm getting two scary deprecation warnings in the code above, can we fix them?

```
[Deprecation] GainNode.gain.value setter smoothing is deprecated and will be removed in M64, around January 2018. Please use setTargetAtTime() instead if smoothing is needed. See https://www.chromestatus.com/features/5287995770929152 for more details.
SpeakerAdapter @ speaker.js:28
22:50:34.877 speaker.js:33 [Deprecation] OscillatorNode.frequency.value setter smoothing is deprecated and will be removed in M64, around January 2018. Please use setTargetAtTime() instead if smoothing is needed. See https://www.chromestatus.com/features/5287995770929152 for more details.
SpeakerAdapter @ speaker.js:33
```

src/browser/speaker.js
```diff
@@ -0,0 +1,127 @@
...
+    this.bus = bus;
+
+    this.audio_context = new (window.AudioContext || window.webkitAudioContext)();
+
+    this.beep_gain = this.audio_context.createGain();
+    this.beep_gain.gain.value = 0;
+    this.beep_gain.connect(this.audio_context.destination);
+
+    this.beep_oscillator = this.audio_context.createOscillator();
+    this.beep_oscillator.type = 'square';
```
copy: Use double quotes.

copy:
@ErnWong Thanks. I have a few more things to address, but then it's ready to go.

I tested it and it worked quite well. The Windows startup sound sounded even better than in qemu. When I find time I'll update the profiles on the website to include some audio files. It would be interesting to see if this also works in Linux, but I couldn't get it to recognize the sb16 in qemu. KolibriOS also has a driver and detects the device successfully, but I wasn't able to get sound in v86 or qemu.

I also checked the performance, in particular I'd like there to be close-to-zero overhead if no sound is playing or the emulator is paused. It looks like the processing goes on even if no audio is played, but from profiling the overhead seems to be almost zero. Oddly, was a higher overhead when the emulator was paused, but I think we can get rid of that.

@cyberplant I'm glad you're enjoying it and also very glad about every contribution.

src/sb16.js
ErnWong: You're right. Good spotting!

ErnWong:
Thanks @copy! I've pushed the next set of changes.

src/lib.js
```diff
@@ -273,93 +273,93 @@ function ByteQueue(size)
...
-        end;
+    this.size = size;
+    this.data = new Float32Array(size);
+    this.start = 0;
+    this.end = 0;
+    this.length = 0;
 
     dbg_assert((size & size - 1) === 0);
 
     this.length = 0;
```
copy: This line is duplicated above.

ErnWong: Yikes :) I'm sorry about that.

ErnWong:
Many, many thanks @copy for your time! (and of course, for this wonderful project)

MailYouLater:
@ErnWong: Is there any particular reason for keeping the [dev-sound](/ErnWong/v86/tree/dev-sound) branch now that it's been merged into [master](/copy/v86/tree/master)?
If you're keeping it because you're planning to do more work on sound (please do) then I would recommend deleting it and creating a new branch (which can have the same name if desired) for that work, so that it's based on the commit history that's in [master](/copy/v86/tree/master) and can be more easily merged next time.

ErnWong:
Thanks @MailYouLater ! That's a good idea. Tomorrow I'll double check for any loose ends on that branch before deleting it.

PR #177, opened 2017-12-13
Fix Source Maps: Reference source files relative to build directory.

ErnWong:
**Problem:** Do the source maps work on your machines? On mine, Google Chrome tries to load files from the nonexistent directory: e.g. `build/src/vga.js`, etc.

**Change:** The Makefile build recipes are modified: they will first `cd` into the `build/` directory before running the closure compiler. The paths to the source files are also rewritten to be relative to the `build/` directory. (I.e. `src/vga.js` becomes `../src/vga.js`)

**Result:** Google Chrome now loads the source files correctly.

Ideally, it'll be better if the Closure Compiler handled the relative paths correctly instead of us having to change directory and change all our input paths. I haven't found an equivalent `--source_map_root` compiler option though.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,9 +26,9 @@ CLOSURE_SOURCE_MAP=\
 		#--jscomp_error newCheckTypes\
 
 CLOSURE_FLAGS=\
-	        --js lib/closure-base.js\
+	        --js ../lib/closure-base.js\
 		--generate_exports\
-		--externs src/externs.js\
+		--externs ../src/externs.js\
 		--warning_level VERBOSE\
 		--jscomp_error accessControls\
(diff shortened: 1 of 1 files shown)


copy:
Good point, but I think we should look for another solution:

1. Add a symlink from `build/src` to `src`
2. `--source_map_include_content`, which will store the contents of the JavaScript files in the source map
3. Fiddle around with `--source_map_location_mapping`

I think 2. is a nice solution, as it also prevents mismatches between the dev files and the build version. Not sure about browser support, but it works at least in Chromium.

copy:
I added `--source_map_include_content`, feel free to reopen if you think anything else should be done.

ErnWong:
Thanks! I like your solution.

Issue #110, opened 2016-09-26
Double click "dos prompt" in windows 98 makes the screen look all glitchy


stuaxo:
http://imgur.com/a/2Fmoc

DaniellPlasmmer:
+1

http://image.prntscr.com/image/f07bb4057f8041b49869259eaa136da2.png

copy:
It's an issue with the video driver (and the same thing happens when I load the image in QEMU): https://web.archive.org/web/20130726072031/http://www.bearwindows.boot-land.net/vbe9x.htm

> Installation note 0: Current driver unsolved problems:
> 
>    Locking and garbage problems when entering DOS WDM or console session.

The reason that we're using this driver instead of the stock driver is that we don't support all VGA features yet, see #90. This driver also has a better performance and supports bigger resolutions.

stuaxo:
Ah, interesting :)

Bitplanes always confused me back in the day.

Issue #90, opened 2016-03-05
VGA Glitches


Creeper20428:
Note: This is probably not related to #26, since I'm dealing with all of VGA.

VGA is kind of glitchy: 
![w95-0](https://cloud.githubusercontent.com/assets/8443268/13545482/5dfab1d0-e242-11e5-990a-26978088f098.png)
![w98-0](https://cloud.githubusercontent.com/assets/8443268/13545485/7ec8495e-e242-11e5-8bc0-2ce8fef5f370.png)

Steps for reproducing problems:
 1a) Open Windows 95. (it's glitchy, downloaded from here: [http://www.mediafire.com/download/85w636eic9sv12h/W95.rar](W95.rar) since I'm lazy)
 2a) Open any application. I just opened IE3 since I wanted to see how Internet Explorer looked like in 1995.

 1b) Open Windows 98
 2b) Open the MS-DOS prompt

fibo:
I guess it is a feature, also Windows glitches.

Creeper20428:
> I guess it is a feature
This doesn't happen in Bochs or any other x86 emulator

copy:
First of all – pretty cool that Windows 95 works at all. It's one of the versions I didn't test. I might put it one the website at some point.

Indeed our vga implementation isn't great. It's better to use a driver that supports vbe, which is implemented correctly, faster and supports bigger resolutions. A driver for Windows can be found here: http://www.navozhdeniye.narod.ru/vbe9x.htm (the same driver is used on Windows 98).

I'll leave this issue open since we might want to fix this at some point (for DOS games or so).

computernewb:
I think it may be due to how it handles 16-colored programs and operating systems. 

Windows 2.03 has some bad graphical glitches as shown here.

<img src="http://i.imgur.com/E5bUETz.png"></img>

Creeper20428:
Yeah... I tried Windows 2 but it's weird like that too.
Did anyone try Windows 3.11 or "Windows 3.X for Workgroups"?

But most of the operating systems I used look good, though.

computernewb:
[...]

Issue #26, opened 2014-10-13
Implement VGA Mode X


ociebieda:
Hello.
I gived a try with this emulator.
I created many bootabla IMG disks with both, FreeDOS and MS-DOS.
FreeDOS and MS-DOS 3.3/6.x are starting up correctly, when Win 95/98 boot disks are not working.
So, I decided to check out some dos games.
On created before FreeDOS and MS-DOS bootable images I added following games: Ninja, Bruce Lee, Golden Axe, Lotus 3 - all of them are simple and work even on real 286 machine.
When I run every of these games, then screen goes 320x200 graphic mode, shows black screen and nothing more happens.
I am very surprised, because like I said, these are simple games which run on 286 computers, and you mention that this emulator is near Pentium 1 machine.
I wonder if there is something to do with emulated graphic hardware?
Thanks for any reply.

ErnWong:
I had a look at some [vga register dumps](https://docs.google.com/spreadsheets/d/16AqaixJ2ttqWM8mCvq7Yx3JhJShTXs8jYvDspzt_eL8/edit?usp=sharing) for some video modes and programs, including these games. It turns out that Ninja and Bruce Lee uses Video Mode 4h/5h, a 4-color 320x200 resolution used when CGA were around. These two games don't work on v86 yet because our VGA implementation doesn't handle the Odd/Even Chain Modes and the Interleaved Shift Mode yet.

(I'm just throwing words around to make it sound like I know what I'm talking about :stuck_out_tongue: but in reality I'm still learning how VGA works)

Golden Axe appears to be using a flavour of the infamous Mode X.

Lotus III uses mode 13h, which has been mostly implemented correctly. If we wait long enough, you can get into the menu screen and into the game if you don't mind the slowness.

![image](https://user-images.githubusercontent.com/2609018/34538852-05ec3ff2-f133-11e7-9e20-87d996f887d7.png)

[...]

PR #178, opened 2017-12-17
Implement VGA graphical behaviour on per-register basis

ErnWong:
### This package contains the following goodness

- VGA write modes 0, 1, 2, 3 with set/reset, planar rotation.
- VGA write using single dword to represent all 4 plane bytes.
- VGA read mode 1.
- Provides a workaround solution to issues #110 and #90 (sacrificing the performance and resolution of the vbe driver for normal vga)

**Update:** this PR now also includes:
- VGA chain odd/even and interleaved shift - implements modes 4h and 5h.
- VGA unchained, chain 4 256 color shift - implements modes 13h, mode X.
- VGA panning, page flipping, split screen.
- VGA palette redraws.

### Demo

(This is the same video as the one for sb16 PR)

https://drive.google.com/file/d/1FSt14j7ayKnaT7AS8lF_4S1i4ZX9GlEk/view
(the demo also contains code for 320x400 display mode that is not included with this PR)

### Notes

- I'm curious to see if you like how the modes are implemented here. The calculations (bitmasks, ALU, etc.) got moved out from inside `vga_memory_write_graphical_planar` into their own, separate functions. This way, ~~the code looks more beautiful and satisfies my OCD~~ the circuitry intentions for each mode is much clearer. I'm not sure about the performance aspects though (I haven't profiled them yet).

### Todo

- [x] Find a way to test out each mode. So far, I've only tested by booting up Windows 98 with VBE drivers removed, which worked well even before read mode 1 was implemented.
- [x] Test with Windows 98: windows, screen savers, paint, solitaire
- [x] Test some of Michael Abrash's Graphics Programming Black Book sample code
- [x] Test some DOS games on the write modes
- [x] Merge routines for mode 12h and 13h into single function.
- [x] Implement unchained 256 color shift
- [x] Implement odd/even chain and interleaved shift
- [x] Implement page flipping, panning, and split screens
- [x] Try out using linked lists to keep track of what pixels use what color, so when the palette changes the pixels can be updated
- [x] Fix the svga and text mode regressions on other OSes
- [x] Fix / optimize / consider keeping or removing the linked list.
- [x] Add new states to `get_state` and `set_state` functions.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+root=true
+
+[*.js]
+charset = utf-8
+indent_style = space
+indent_size = 4
--- a/src/browser/screen.js
+++ b/src/browser/screen.js
@@ -537,24 +537,33 @@ function ScreenAdapter(screen_container, bus)
             {
                 graphic_context.strokeRect(
                     layer.buffer_x,
-                    layer.buffer_y,
+                    layer.buffer_min_y,
                     layer.buffer_width,
-                    layer.buffer_height
+                    layer.buffer_max_y - layer.buffer_min_y
                 );
             });
             graphic_context.lineWidth = 1;
             return;
         }
+
+        var min_y = min / graphical_mode_width | 0;
+        var max_y = max / graphical_mode_width | 0;
+
         layers.forEach((layer) =>
         {
+            var buffer_min_y = Math.max(min_y, layer.buffer_min_y);
+            var buffer_max_y = Math.min(max_y, layer.buffer_max_y);
+
+            var buffer_height = Math.max(0, buffer_max_y - buffer_min_y);
+
             graphic_context.putImageData(
                 graphic_image_data,
                 layer.screen_x - layer.buffer_x,
-                layer.screen_y - layer.buffer_y,
+                layer.screen_y - layer.buffer_min_y,
                 layer.buffer_x,
-                layer.buffer_y,
+                layer.buffer_min_y,
                 layer.buffer_width,
-                layer.buffer_height
+                buffer_height
             );
         });
     };
--- a/src/browser/screen.js
+++ b/src/browser/screen.js
@@ -176,6 +176,11 @@ function ScreenAdapter(screen_container, bus)
         layers_has_changed = true;
     }, this);
 
+    bus.register("screen-clear", function()
+    {
+        this.clear_screen();
+    }, this);
+
     bus.register("screen-set-size-text", function(data)
     {
         this.set_size_text(data[0], data[1]);
@@ -514,7 +519,6 @@ function ScreenAdapter(screen_container, bus)
 
         layers_has_changed = false;
 
-        this.clear_screen();
         if(DEBUG_SCREEN_LAYERS)
         {
             // Draw the entire buffer. Useful for debugging
--- a/src/browser/screen.js
+++ b/src/browser/screen.js
@@ -175,7 +175,7 @@ function ScreenAdapter(screen_container, bus)
     }, this);
     bus.register("screen-set-size-graphical", function(data)
     {
-        this.set_size_graphical(data[0], data[1], data[3], data[4]);
+        this.set_size_graphical(data[0], data[1], data[2], data[3]);
     }, this);
 
 
--- a/src/browser/screen.js
+++ b/src/browser/screen.js
@@ -64,9 +64,6 @@ function ScreenAdapter(screen_container, bus)
         // number of rows
         text_mode_height;
 
-    var layers = [];
-    var layers_has_changed = false;
-
     var screen = this;
 
     // 0x12345 -> "#012345"
@@ -149,10 +146,7 @@ function ScreenAdapter(screen_container, bus)
 
     bus.register("screen-fill-buffer-end", function(data)
     {
-        var min = data[0];
-        var max = data[1];
-
-        this.update_buffer(min, max);
+        this.update_buffer(data);
     }, this);
 
     bus.register("screen-put-char", function(data)
@@ -168,12 +162,6 @@ function ScreenAdapter(screen_container, bus)
     bus.register("screen-update-cursor-scanline", function(data)
     {
         this.update_cursor_scanline(data[0], data[1]);
-    }, this);
-
-    bus.register("screen-update-layers", function(data)
-    {
-        layers = data;
-        layers_has_changed = true;
     }, this);
 
     bus.register("screen-clear", function()
@@ -336,8 +324,6 @@ function ScreenAdapter(screen_container, bus)
 
         graphical_mode_width = width;
         graphical_mode_height = height;
-
-        layers = [];
 
         this.bus.send("screen-tell-buffer", [graphic_buffer32], [graphic_buffer32.buffer]);
         update_scale_graphic();
@@ -510,15 +496,8 @@ function ScreenAdapter(screen_container, bus)
         row_element.parentNode.replaceChild(fragment, row_element);
     };
 
-    this.update_buffer = function(min, max)
+    this.update_buffer = function(layers)
     {
-        if(max < min && !layers_has_changed)
-        {
-            return;
-        }
-
-        layers_has_changed = false;
-
         if(DEBUG_SCREEN_LAYERS)
         {
             // Draw the entire buffer. Useful for debugging
@@ -537,33 +516,24 @@ function ScreenAdapter(screen_container, bus)
             {
                 graphic_context.strokeRect(
                     layer.buffer_x,
-                    layer.buffer_min_y,
+                    layer.buffer_y,
                     layer.buffer_width,
-                    layer.buffer_max_y - layer.buffer_min_y
+                    layer.buffer_height
                 );
             });
             graphic_context.lineWidth = 1;
             return;
         }
-
-        var min_y = min / graphical_mode_width | 0;
-        var max_y = max / graphical_mode_width | 0;
-
         layers.forEach((layer) =>
         {
-            var buffer_min_y = Math.max(min_y, layer.buffer_min_y);
-            var buffer_max_y = Math.min(max_y, layer.buffer_max_y);
-
-            var buffer_height = Math.max(0, buffer_max_y - buffer_min_y);
-
             graphic_context.putImageData(
                 graphic_image_data,
                 layer.screen_x - layer.buffer_x,
-                layer.screen_y - layer.buffer_min_y,
+                layer.screen_y - layer.buffer_y,
                 layer.buffer_x,
-                layer.buffer_min_y,
+                layer.buffer_y,
                 layer.buffer_width,
-                buffer_height
+                layer.buffer_height
             );
         });
     };
--- a/src/browser/screen.js
+++ b/src/browser/screen.js
@@ -65,6 +65,7 @@ function ScreenAdapter(screen_container, bus)
         text_mode_height;
 
     var layers = [];
+    var layers_has_changed = false;
 
     var screen = this;
 
@@ -171,7 +172,8 @@ function ScreenAdapter(screen_container, bus)
 
     bus.register("screen-update-layers", function(data)
     {
-        this.layers = data;
+        layers = data;
+        layers_has_changed = true;
     }, this);
 
     bus.register("screen-set-size-text", function(data)
@@ -180,15 +182,6 @@ function ScreenAdapter(screen_container, bus)
     }, this);
     bus.register("screen-set-size-graphical", function(data)
     {
-        if(DEBUG_SCREEN_LAYERS)
-        {
-            data[0] = data[3];
-            data[1] = data[4];
-        }
-        if(!data[0]) data[0] = 1;
-        if(!data[1]) data[1] = 1;
-        if(!data[3]) data[3] = data[0];
-        if(!data[4]) data[4] = data[1];
         this.set_size_graphical(data[0], data[1], data[3], data[4]);
     }, this);
 
@@ -312,6 +305,15 @@ function ScreenAdapter(screen_container, bus)
 
     this.set_size_graphical = function(width, height, buffer_width, buffer_height)
     {
+        if(DEBUG_SCREEN_LAYERS)
+        {
+            // Draw the entire buffer. Useful for debugging
+            // panning / page flipping / screen splitting code for both
+            // v86 developers and os developers
+            width = buffer_width;
+            height = buffer_height;
+        }
+
         graphic_screen.style.display = "block";
 
         graphic_screen.width = width;
@@ -330,7 +332,7 @@ function ScreenAdapter(screen_container, bus)
         graphical_mode_width = width;
         graphical_mode_height = height;
 
-        this.layers =
+        layers =
         [{
             screen_x: 0, screen_y: 0,
             buffer_x: 0, buffer_y: 0,
@@ -510,31 +512,29 @@ function ScreenAdapter(screen_container, bus)
 
     this.update_buffer = function(min, max)
     {
-        /*if(max < min)
+        if(max < min && !layers_has_changed)
         {
             return;
         }
 
-        var min_y = min / graphical_mode_width | 0;
-        var max_y = max / graphical_mode_width | 0;
+        layers_has_changed = false;
 
-        graphic_context.putImageData(
-            graphic_image_data,
-            0, 0,
-            0, min_y,
-            graphical_mode_width, max_y - min_y + 1
-        );
-        */
         this.clear_screen();
         if(DEBUG_SCREEN_LAYERS)
         {
+            // Draw the entire buffer. Useful for debugging
+            // panning / page flipping / screen splitting code for both
+            // v86 developers and os developers
             graphic_context.putImageData(
                 graphic_image_data,
                 0, 0
             );
+
+            // For each visible layer that would've been drawn, draw a
+            // rectangle to visualise the layer instead.
             graphic_context.strokeStyle = "#0F0";
             graphic_context.lineWidth = 4;
-            this.layers.forEach((layer) =>
+            layers.forEach((layer) =>
             {
                 graphic_context.strokeRect(
                     layer.buffer_x,
@@ -546,7 +546,7 @@ function ScreenAdapter(screen_container, bus)
             graphic_context.lineWidth = 1;
             return;
         }
-        this.layers.forEach((layer) =>
+        layers.forEach((layer) =>
         {
             graphic_context.putImageData(
                 graphic_image_data,
--- a/src/browser/screen.js
+++ b/src/browser/screen.js
@@ -64,6 +64,8 @@ function ScreenAdapter(screen_container, bus)
         // number of rows
         text_mode_height;
 
+    var layers = [];
+
     var screen = this;
 
     // 0x12345 -> "#012345"
@@ -167,13 +169,27 @@ function ScreenAdapter(screen_container, bus)
         this.update_cursor_scanline(data[0], data[1]);
     }, this);
 
+    bus.register("screen-update-layers", function(data)
+    {
+        this.layers = data;
+    }, this);
+
     bus.register("screen-set-size-text", function(data)
     {
         this.set_size_text(data[0], data[1]);
     }, this);
     bus.register("screen-set-size-graphical", function(data)
     {
-        this.set_size_graphical(data[0], data[1]);
+        if(DEBUG_SCREEN_LAYERS)
+        {
+            data[0] = data[3];
+            data[1] = data[4];
+        }
+        if(!data[0]) data[0] = 1;
+        if(!data[1]) data[1] = 1;
+        if(!data[3]) data[3] = data[0];
+        if(!data[4]) data[4] = data[1];
+        this.set_size_graphical(data[0], data[1], data[3], data[4]);
     }, this);
 
 
@@ -294,7 +310,7 @@ function ScreenAdapter(screen_container, bus)
         update_scale_text();
     };
 
-    this.set_size_graphical = function(width, height)
+    this.set_size_graphical = function(width, height, buffer_width, buffer_height)
     {
         graphic_screen.style.display = "block";
 
@@ -307,12 +323,19 @@ function ScreenAdapter(screen_container, bus)
         // Make sure to call this here, because pixels are transparent otherwise
         //screen.clear_screen();
 
-        graphic_image_data = graphic_context.createImageData(width, height);
+        graphic_image_data = graphic_context.createImageData(buffer_width, buffer_height);
         graphic_buffer = new Uint8Array(graphic_image_data.data.buffer);
         graphic_buffer32 = new Int32Array(graphic_image_data.data.buffer);
 
         graphical_mode_width = width;
         graphical_mode_height = height;
+
+        this.layers =
+        [{
+            screen_x: 0, screen_y: 0,
+            buffer_x: 0, buffer_y: 0,
+            buffer_width: width, buffer_height: height
+        }];
 
         this.bus.send("screen-tell-buffer", [graphic_buffer32], [graphic_buffer32.buffer]);
         update_scale_graphic();
@@ -487,7 +510,7 @@ function ScreenAdapter(screen_container, bus)
 
     this.update_buffer = function(min, max)
     {
-        if(max < min)
+        /*if(max < min)
         {
             return;
         }
@@ -501,6 +524,40 @@ function ScreenAdapter(screen_container, bus)
             0, min_y,
             graphical_mode_width, max_y - min_y + 1
         );
+        */
+        this.clear_screen();
+        if(DEBUG_SCREEN_LAYERS)
+        {
+            graphic_context.putImageData(
+                graphic_image_data,
+                0, 0
+            );
+            graphic_context.strokeStyle = "#0F0";
+            graphic_context.lineWidth = 4;
+            this.layers.forEach((layer) =>
+            {
+                graphic_context.strokeRect(
+                    layer.buffer_x,
+                    layer.buffer_y,
+                    layer.buffer_width,
+                    layer.buffer_height
+                );
+            });
+            graphic_context.lineWidth = 1;
+            return;
+        }
+        this.layers.forEach((layer) =>
+        {
+            graphic_context.putImageData(
+                graphic_image_data,
+                layer.screen_x - layer.buffer_x,
+                layer.screen_y - layer.buffer_y,
+                layer.buffer_x,
+                layer.buffer_y,
+                layer.buffer_width,
+                layer.buffer_height
+            );
+        });
     };
 
     this.init();
--- a/src/browser/screen.js
+++ b/src/browser/screen.js
@@ -332,12 +332,7 @@ function ScreenAdapter(screen_container, bus)
         graphical_mode_width = width;
         graphical_mode_height = height;
 
-        layers =
-        [{
-            screen_x: 0, screen_y: 0,
-            buffer_x: 0, buffer_y: 0,
-            buffer_width: width, buffer_height: height
-        }];
+        layers = [];
 
         this.bus.send("screen-tell-buffer", [graphic_buffer32], [graphic_buffer32.buffer]);
         update_scale_graphic();
--- a/src/browser/screen.js
+++ b/src/browser/screen.js
@@ -524,6 +524,7 @@ function ScreenAdapter(screen_container, bus)
             graphic_context.lineWidth = 1;
             return;
         }
+
         layers.forEach((layer) =>
         {
             graphic_context.putImageData(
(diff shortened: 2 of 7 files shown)


copy:
It would be interesting to test some DOS games as mentioned in #26. We should also test [DexOS](https://dex-os.github.io/), which is also currently broken due to missing vga features.

BenNottelling:
> It would be interesting to test some DOS games as mentioned in #26. We should also test DexOS, which is also currently broken due to missing vga features.
screen shot 2017-12-17 at 10 30 10 am
Looks like it's still broken

copy:
>  Looks like it's still broken

Ah, right. 800x600 is not one of the old vga modes.

BenNottelling:
> Ah, right. 800x600 is not one of the old vga modes.

I'm not sure what you're saying/what you mean.

src/vga.js
```diff
@@ -185,6 +181,22 @@ function VGAScreen(cpu, bus, vga_memory_size)
...
+    this.color_compare = 0;
+    this.color_compare0 = 0;
+    this.color_compare1 = 0;
+    this.color_compare2 = 0;
+    this.color_compare3 = 0;
+    this.color_dont_care = 0;
+    this.color_care0 = 0;
+    this.color_care1 = 0;
+    this.color_care2 = 0;
+    this.color_care3 = 0;
```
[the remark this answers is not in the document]
ErnWong: Good idea.

src/vga.js
```diff
@@ -629,6 +583,117 @@ VGAScreen.prototype.vga_memory_write_graphical_planar = function(addr, value)
...
+ * @return {number} 32-bit number representing the bytes for each plane.
+ */
+VGAScreen.prototype.apply_feed = function(data_byte)
+{
+    var dword = 0;
+    dword |= data_byte << 0;
+    dword |= data_byte << 8;
+    dword |= data_byte << 16;
+    dword |= data_byte << 24;
+    return dword;
```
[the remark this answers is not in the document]
ErnWong: *Oh no my ocd* 🙈 *Goodbye to my dreams of aligning everything perfectly.*

But nah, I agree that things like ` data_byte << 0` are just silly.

copy:
> I'm not sure what you're saying/what you mean.

DexOS uses a VGA mode that wasn't changed in this PR, so I couldn't expect this change to fix DexOS.

BenNottelling:
Ohhhh okay. That makes sense

ErnWong:
win98-2color-glitch

Setting Windows 98 to 2-Color appears to leave some black regions undrawn, but that happens to VirtualBox as well so it's not too much of a concern.

ErnWong:
The following code samples from Michael Abrash's Black Book (http://www.jagregory.com/abrash-black-book/) chapters 24 to 28 does not appear to work:

## L26-1

Coloured text should be drawn on the blue background.
image

## L26-2

White text should be drawn on the blue background.
image

## L27-3

After pressing key twice, the screen should return back to graphics mode with the previous graphics:
image

## L28-3

Should be a white diagonal line:
image

I'll go over the code and see where it went wrong. It shouldn't be *too* difficult to debug, as we've got the source code for the programs.

ErnWong:
Fixed the bug with write mode 3 (L26-1, L26-2, L28-3).

In sample program L27-3, I believe that the bios is incorrectly clearing the display memory while changing modes. The program uses the fact that setting bit 7 of AL during video mode change will instruct the bios to not clear the memory, yet from the debugger I see `vga_memory_write_graphical_planar` being called to clear the memory when changing to mode 90h (10h with bit 7 set).

In sample program L28-2, occasionally, some pixels are missing and those pixels are different each time:

image

But otherwise, all the code samples from chapters 24 to 28 are now working to a more or less satisfactory degree. I'll go and investigate L28-2.

ErnWong:
Turns out that `screen_fill_buffer` does not copy over the very last byte of display memory that was changed. Since `diff_addr_max` is inclusive, I'm modifying the `for` loop: from `i < end` into `i <= end`

https://github.com/copy/v86/blob/b22826525b7831b7f5b5ccffa286d283917e5af4/src/vga.js#L1426-L1435

ErnWong:
Completed testing using Abrash's Book's sample programs. The next step is the fun part: play around with old DOS games.

src/vga.js
```diff
@@ -1502,7 +1502,7 @@ VGAScreen.prototype.screen_fill_buffer = function()
             var start_pixel = start - offset;
             var end_pixel = end - offset + 1;
 
-            for(var i = start; i < end; i++)
+            for(var i = start; i <= end; i++)
```
copy: Nice find!

ErnWong:
Current situation for a small selection of old dos games:

## Skyroads (playable)

 - Uses 13h video mode - working.
 - Changes the 256 color palette for transition - changes are not redrawn automatically, so parts of the screen remain black at times.
 - Uses sb16's 8-bit single cycle dma transfer mode - working.

## Wolfenstein 3d (unplayable)

 - Custom "high res 13h mode" (320x400 256 color) in the beginning.
   Miscellaneous Output Register: `0x63`, Sequencer memory mode: `0x06`
 - Stuck in black screen, constantly reading `3DA`.

## Mario (unplayable)

 - Custom "high res 13h mode" (320x400 256 color)
   Miscellaneous Output Register: `0x63`, Sequencer memory mode: `0x06`
   Right colors, but heavily distorted pixel locations similar but not exactly identical to Windows 98 startup screen.

## Simcity (unplayable)

 - "Program too big to fit in memory" for even when 2Gb of memory has been configured in v86. Same with virtualbox, so it's probably a faulty installation/build of the SimCity demo that I'm using.

## Doom (unplayable)

 - Custom "high res 13h mode" (320x400 256 color)
   Miscellaneous Output Register: `0x63`, Sequencer memory mode: `0x06`
   Similar to Mario, but slightly less distorted
 - Failed asserts: "Overlapped read16 0x0224 (undefined)" . That appears to be the sb16 mixer read address read port - potential bug with sb16.

## Syndicate (unplayable)

 - Mode 13h splash screen, but stuck at that very same splash screen. Esc works to quit. There is a cursor drawn but it is not responsive.

## Lemmings (unplayable)

 - Custom "high res 13h mode" (320x400 256 color)
   Miscellaneous Output Register: `0x63`, Sequencer memory mode: `0x06`
 - Stuck in black screen, constantly reading `3DA`.

## Duel 2000 (a little star battle game) (playable)

 - Mode 12h (write modes 0,2) - working
   Miscellaneous Output Register: `0xE3`, Sequencer memory mode: `0x06` (the game didn't appear to modify this register)

## Capture the flag (playable)

 - Mode 12h (write mode 0) - Yikes! Wrong colors! `dac_map` or `palette`?
   Miscellaneous Output Register: `0xE3`, Sequencer memory mode: `0x06`
 - Sound working.

## Brucelee (unplayable)

 - Custom "high res 13h mode" (320x400 256 color)
   Miscellaneous Output Register: `0x63`, Sequencer memory mode: `0x06`
 - Sound working.

## Golden Axe (unplayable)

 - Custom "high res 13h mode" (320x400 256 color)
   Miscellaneous Output Register: `0x63`, Sequencer memory mode: `0x06`
 - Stuck in black screen, constantly reading `3DA`.

## Lotus (unplayable)

 - Mode 13h
   Miscellaneous Output Register: `0x63`, Sequencer memory mode: `0x0E`
 - Stuck in black screen, constantly reading `3DA`.

## Ninja (unplayable)

 - Miscellaneous Output Register: `0x63`, Sequencer memory mode: `0x02`?
 - Blank screen, and quits?

## The Incredible Machine (unplayable)
 - Mode 12h
   Miscellaneous Output Register: `0xE3`, Sequencer memory mode: `0x06`?
 - Stuck in black screen, constantly reading `3DA`.

## Todo

- [ ] Try the games out using the "high res 13h mode" crudely implemented in branch [pr-vga-320x400](https://github.com/ErnWong/v86/tree/pr-vga-320x400).
- [ ] Investigate the `3DA` reads.
- [ ] Investigate the incorrect colors in Capture the Flag.
- [ ] I just realised that the sequencer memory mode register values for mode 12h seems to be different to what [osdev wiki tables](http://wiki.osdev.org/VGA_Hardware#List_of_register_settings) says (yet it works?). I will try to consolidate these inconsistencies. Maybe the sequencer memory mode register is not the determining factor.
- (unrelated to this PR) Investigate I/O read16 port 224 (sb16 mixer address port)

ErnWong:
Oops, I've misidentified a lot of the video modes above.

BenNottelling:
The PR cleans up the reactOS boot text which is very nice.
after
before

ErnWong:
I've just pushed a pretty big set of changes that implements unchained modes, odd/even modes, page flipping, panning, split screen, and added some mechanism to update the screen when the palettes get updated. The overall pipeline for vanilla VGA now looks like [this](https://gist.github.com/ErnWong/7db67164224aaa3be5f6fcc2a6f5dd21).

At the moment, the "Color Use" linked list is still a bit buggy and causes flashes of black patches on the screen on very few programs. Enabling this linked list appears to make redraws faster on average, but slower during transitions where most of the color palettes are modified.

I thought it'll be more efficient to rely on built-in Canvas API functions to handle the page flipping/panning, so in this PR, the `ScreenAdapter` is currently responsible for assembling the final picture based on an array of 'layers' information passed on from `VGAScreen`. Maybe there's a better way of implementing this, but this is what I've come up with so far.

The new VGAScreen states have yet to be included in the `save_state` and `get_state` functions.

ErnWong:
Apologies. I referenced the incorrect issue in one of the commit messages, and in a panic I rebased and forced pushed to correct it.

MailYouLater:
@BenNottelling: It also (at least now) clears up the visual glitch in the MSDOS 6.22 setup you mentioned on #179.
image

ErnWong:
Hey @copy, do I need to bump the `STATE_VERSION` number?

copy:
> Hey @copy, do I need to bump the STATE_VERSION number?

Yes.

Let me know when this is ready for review (but give me some time to review it, since it's a large patch). I'll leave a few initial comments now.

.editorconfig
```diff
@@ -0,0 +1,6 @@
+root=true
+
+[*.js]
+charset = utf-8
+indent_style = space
+indent_size = 4
```
copy: Not sure if this was intentional, but it's a good idea to have it, so leave it.

src/vga.js
```diff
@@ -1333,113 +2212,349 @@ VGAScreen.prototype.svga_register_read = function(n)
...
+    var dac_next = this.dac_color_use_next;
+
+    this.dac_color_has_changed.fill(0);
+
+    var color_use_disabled = this.dac_color_use_disabled;
+
+    // If half, or more than half the colors have changed, better to
+    // do a complete redraw than using the linked list.
+    color_use_disabled |= (this.attribute_mode & 0x40) && dac_changed.length > 127;
+    color_use_disabled |= !(this.attribute_mode & 0x40) && dac_changed.length > 7;
```
copy: Is the dac linked list solely an optimisation? Can you explain the optimisation to me? It seems to complicate the logic significantly, if it is solely an optimisation what applications benefit from it?

src/browser/screen.js
```diff
@@ -167,13 +169,27 @@ function ScreenAdapter(screen_container, bus)
...
-        this.set_size_graphical(data[0], data[1]);
+        if(DEBUG_SCREEN_LAYERS)
+        {
+            data[0] = data[3];
+            data[1] = data[4];
+        }
+        if(!data[0]) data[0] = 1;
+        if(!data[1]) data[1] = 1;
+        if(!data[3]) data[3] = data[0];
+        if(!data[4]) data[4] = data[1];
```
copy: Move these checks into `set_size_graphical` and add a comment to what the debug block does.

If these values are controlled by the guest VM, they should probably be checked earlier.

src/vga.js
ErnWong: Yes. The linked list was initially created for these reasons:

- There are many games (and probably many GUI programs out there) that relies on changing the DAC palette to create smooth transitions, fade in/out, and animations. Therefore, we need to do some kind of `redraw` at some point to update the destination buffer with the correct values (or else the screen remains blank/incorrect and unusable).
- I saw this comment:
https://github.com/copy/v86/blob/153c50f257b0ab8fa5d8707d9f4680f209d5a817/src/vga.js#L969-L974
- There was a significant amount of lag (about 1 second) when doing a complete redraw during the Windows 98 startup screen (e.g. notice the black gap at the top of the screen between 5 seconds and 6 seconds in [this video](https://drive.google.com/file/d/1FSt14j7ayKnaT7AS8lF_4S1i4ZX9GlEk/view)), so I thought that a complete redraw would be too expensive for smooth transitions.
- ~~I think~~ I'm obsessed with overengineering :stuck_out_tongue:

I thought it'll be quite inefficient to redraw the entire buffer if only a relatively small number of pixels needs to be updated. The linked list thing is effectively 256 independent linked list, each keeping track of the pixel addresses that use a given color.

During a redraw, we loop through the colors that have changed. For a given `color256`, we can get the first pixel address that uses that color using `dac_color_use_root[color256]`.

The subsequent pixel addresses that also uses that color is found using `dac_color_use_next[pixel_addr]`. We go through the list until we reach the end, with the value `VGA_UNUSED_COLOR` as the terminator.

image

To keep the linked lists up to date, whenever we modify the `pixel_buffer`, we move the pixel address from the old list to the new list. We join the pixel's previous neighbour nodes together in the previous list (to remove the pixel), and insert the pixel into the front of the new list. This remove and insertion step takes constant time.

However, `complete_redraw`s are no longer that expensive. In non-svga modes, I've limited the redraw to only the pixel buffer (512KiB) instead of the entire svga buffer (defaulted to 8MiB)

https://github.com/ErnWong/v86/blob/be4e3ee2f54ac0b363071643c7fea3b906ef715c/src/vga.js#L983-L998

Yeah, I agree that this piece linked list adds quite a fair bit of complexity to the VGA controller, and I'm happy to remove it until we find a use case for it. In fact, I've been thinking about removing it, since `complete_redraw`s for small VGA resolutions are already pretty fast (I don't have quantitative figures though). The linked lists consumes about 4MiB of memory. Compare this with the 512MiB pixel buffer. Moreover, when the guest os / programs change the DAC palette, they usually change all of the pixels anyway.

I think that the only situation that will greatly benefit from this are with larger SVGA resolutions using 8bpp that do transitions for a small number of pixels.

What do you think?

src/vga.js
```diff
@@ -273,6 +386,34 @@ function VGAScreen(cpu, bus, vga_memory_size)
     this.plane1 = new Uint8Array(this.svga_memory.buffer, 1 * VGA_BANK_SIZE, VGA_BANK_SIZE);
     this.plane2 = new Uint8Array(this.svga_memory.buffer, 2 * VGA_BANK_SIZE, VGA_BANK_SIZE);
     this.plane3 = new Uint8Array(this.svga_memory.buffer, 3 * VGA_BANK_SIZE, VGA_BANK_SIZE);
+    this.pixel_buffer = new Uint8Array(this.svga_memory.buffer,
+        VGA_PIXEL_BUFFER_START, VGA_PIXEL_BUFFER_SIZE);
+
+    if(this.vga_memory_size >= VGA_MIN_MEMORY_SIZE + VGA_COLOR_USE_LIST_SIZE)
```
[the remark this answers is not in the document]
ErnWong: Yeah, the idea was to not use the dac linked list if vga_memory_size is not large enough, so I can keep the memory footprint of it down. `VGA_MIN_MEMORY_SIZE` is also the svga memory address where the dac linked lists start.

`VGA_MIN_MEMORY_SIZE + VGA_COLOR_USE_LIST_SIZE` is thus the total memory size of: the vga planes, the vga pixel buffer, and the dac linked lists (see the memory layout at the bottom of this [gist](https://gist.github.com/ErnWong/7db67164224aaa3be5f6fcc2a6f5dd21 for some pictures).

I'll add some comments, or rename some variables to make this clearer for any future reader if we decide to keep the dac linked lists (see my comment below).

ErnWong:
Thanks @copy! There's one last thing I want to fix before it's ready for review: I still want to investigate why these colors are incorrect below.

image

src/browser/screen.js
ErnWong: Good point

ErnWong:
Done. @copy I think it's ready for review!

ErnWong:
Wait, sorry, it appears that one of the last couple of commits had led to a regression.

ErnWong:
Ok, I think it's ready.

MailYouLater:
I don't know if this is necessarily relevant to this PR, but while the current state of this PR shows great improvements over what's already in the repo, some of the colors are still incorrect for ReactOS' boot screen in debug mode.

Version currently hosted on copy.sh/v86:
image
This PR:

QEMU:

ErnWong:
Interesting. Probably is relevant to this PR.

I'll start with something more important though:

image

The 4 bit color `0xF` is mapped by the attribute controller palette to the 8 bit color `0x0F`, but this is then incorrectly mapped by the DAC palette to the color `0x0` (black) instead of white.

ErnWong:
I haven't found out the source of the above problem yet. ReactOS gradually changes the DAC palette to make a smooth transition from black. However, when it reaches the end of the transition, it sets all the colors correctly except for `0xF`. See the final line in the [logs](https://gist.github.com/ErnWong/1cf81d6128784c2bb77274964271cfb5). It does not appear to be calling other vga ports while this happens, or maybe it's not logged.

I will continue this investigation tomorrow. The easiest way might be to look at the corresponding ReactOS source code. Thanks everyone for posting the glitchy screenshots aye.

ErnWong:
This might be the problem:

We're treating the DAC palette red/green/blue bytes as 6-bit values (ignoring bits 6 and 7, and scaling it up by a factor of 255/63), as per [page 105 of the IBM VGA reference](http://www.mcamafia.de/pdf/ibm_vgaxga_trm2.pdf) and the [FreeVGA Resource](http://www.osdever.net/FreeVGA/vga/colorreg.htm#3C9). This, AFAIK is the original VGA behaviour.

However, it appears that these values can also be 8-bit values on some chips/cards, ([page 61 of Intel Legacy VGA features](https://01.org/sites/default/files/documentation/intel-gfx-prm-osrc-hsw-display_0.pdf)).

By rotating the bits instead of scaling, we seem to get to correct color.

Prior to change:
https://github.com/copy/v86/blob/153c50f257b0ab8fa5d8707d9f4680f209d5a817/src/vga.js#L947-L953

If we change one of the lines to this:
```
olor_byte = color_byte << 2 | (color_byte >>> 7 & 0x3);
```

We get the correct colors:

image

image

However, I'm not sure if the bits are supposed to be rotate like this. I'll see if I can find some documentation for this, and see if it still displays correctly on other programs/oses.

ErnWong:
The bit rotation is not an accurate approach because the whitest 18-bit color will result in not the whitest 24-bit color.

I think a better approach would be to ignore the last two bits first, before doing the scale of `255/63`, then finally convert to integer:

```
color_byte = (color_byte & 0x3F) * 255 / 63 | 0
```

src/vga.js
copy: Thanks for the explanation. Yes, I think we should get rid of it for the reasons you stated (a complete redraw being fast enough anyway in non-svga modes). Could you still quickly get a number on how long complete redraw takes?

> I think that the only situation that will greatly benefit from this are with larger SVGA resolutions using 8bpp that do transitions for a small number of pixels.

I'm actually not aware of any OSes that use this mode, unless you explicitly configure them so. But even then, the usefulness of this mode over 16bpp svga is doubtful.

src/vga.js
```diff
@@ -740,115 +985,521 @@ VGAScreen.prototype.complete_redraw = function()
...
+        this.text_mode_redraw();
+    }
 
-    if(is_graphical)
+    if(this.svga_enabled || !this.virtual_width || !this.svga_width)
     {
-        this.svga_width = width;
-        this.svga_height = height;
-        this.set_size_graphical(width, height, 8);
+        // Skip for svga mode and avoid division by zero
```
copy: Should we reinitialise the layers to a single layer here, in case the controller has previously been in vga mode?

src/browser/screen.js
```diff
@@ -307,13 +325,20 @@ function ScreenAdapter(screen_container, bus)
...
 
         graphical_mode_width = width;
         graphical_mode_height = height;
 
+        layers =
+        [{
+            screen_x: 0, screen_y: 0,
+            buffer_x: 0, buffer_y: 0,
+            buffer_width: width, buffer_height: height
+        }];
```
copy: This is the configuration for svga, right? In the reset state, shouldn't we start without layers?

src/vga.js
ErnWong: > ...the usefulness of this mode over 16bpp svga is doubtful

Agree.

> Could you still quickly get a number on how long complete redraw takes?

Sure thing. Here's some very rough readings, which are absolute times from my laptop [after adding a few lines](https://gist.github.com/ErnWong/27be47ac4f8f35ea6b45a1b1c08291bd#file-vga-js-L2572).

- 8-bit color modes start at about 3ms/redraw and speeds up to about 2ms/redraw (probably due to v8's optimizations).

- 4-bit color modes start at about 6ms/redraw and quickly speeds up to about 3.5ms/redraw. It's probably slower because it has to go through both `dac_map` and `vga256_palette`

Are these the kind of numbers you're looking for?

copy: Yes, these numbers are good. Let's throw the linked list out.

src/browser/screen.js
ErnWong: > This is the configuration for svga, right?

Yup, this is the configuration so that `update_buffer` behaves the same way as it used to.

> In the reset state, shouldn't we start without layers?

Hmm, good point. Starting without layers will help suppress drawing incorrect graphics when the video mode hasn't been fully configured yet, and it's more intuitive to draw nothing than to draw something potentially incorrect.

src/vga.js
ErnWong: Good idea. If I'm thinking what you're thinking, that probably means that we should call this `update_vga_panning` function every time the port 1CF dispi write function gets called. That way, when the controller changes between svga <--> vga mode, the layers are setup correctly.

That means that we could rename the `update_vga_panning` function to `update_layers` or something.

src/vga.js
```diff
@@ -252,11 +344,17 @@ function VGAScreen(cpu, bus, vga_memory_size)
...
+    this.diff_plot_max = 0;
 
     this.dest_buffer = undefined;
 
     bus.register("screen-tell-buffer", function(data)
     {
+        if(this.dest_buffer && data[0])
+        {
+            data[0].set(this.dest_buffer.subarray(0,data[0].length));
+        }
```
copy: This copies the raw pixels from the old to the new buffer after a resize. Won't the result look distorted?

(minor: Missing space after `,`)

ErnWong: > Won't the result look distorted?

For a brief moment and for most situations, probably yes. It's there just in case some programs decide to draw into the video memory before configuring the resolution, or decide to reuse the graphics already stored in the video memory, which I think should work fine with real VGA hardware.

src/browser/screen.js
```diff
@@ -487,20 +507,52 @@ function ScreenAdapter(screen_container, bus)
...
-        var max_y = max / graphical_mode_width | 0;
+        layers_has_changed = false;
 
-        graphic_context.putImageData(
-            graphic_image_data,
-            0, 0,
-            0, min_y,
-            graphical_mode_width, max_y - min_y + 1
-        );
+        this.clear_screen();
```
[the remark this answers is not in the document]
ErnWong: > Is this needed?

Because we currently draw layers that fill the entire screen, probably not. I'll have a check to see if removing `clear_screen()` impacts any test programs. There is a case where we want to blank the screen:

https://github.com/ErnWong/v86/blob/932bdecf63ff2ed81e5b6259bf20dafd869d9c5d/src/vga.js#L1242-L1249

But I think we can implement a 'clear-screen' bus message instead of clearing the screen consistently.

> It certainly shouldn't be called in svga mode.

I agree.

src/browser/screen.js
```diff
@@ -487,20 +507,52 @@ function ScreenAdapter(screen_container, bus)
...
+                    layer.buffer_width,
+                    layer.buffer_height
+                );
+            });
+            graphic_context.lineWidth = 1;
+            return;
+        }
+        layers.forEach((layer) =>
+        {
+            graphic_context.putImageData(
```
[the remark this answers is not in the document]
ErnWong: > Can we restore the old behaviour at least for svga modes?

Sure. Since there are only a couple of layers at most, adding some logic to restrict the drawing of each layer shouldn't leave much impact.

> If possible, we should move the layer handling to vga.js, since this file is for platform-independent rendering (but since the layer logic is quite simple we could leave it here).

I think I get what you mean. I'll first take account of the `min` and `max` in `screen.js` while keeping `screen.js` device-agnostic and platform-agnostic. After that, I think I'll try implementing the layering in `vga.js` to compare the performances.

src/vga.js
copy: Sounds good.

src/browser/screen.js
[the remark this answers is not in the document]
copy: a `clear-screen` message sounds reasonable if we need it, but I believe in this case we could just check if `layers` is empty in this function.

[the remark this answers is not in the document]
ErnWong: If each layer object represents a patch of graphics to be drawn on top of the existing canvas, I think an empty `layers` array would make more sense to represent a "don't draw over the existing canvas, and keep whatever's currently drawn" instead of a "clear everything". What do you think?

src/browser/screen.js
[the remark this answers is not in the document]
ErnWong: >Sure. Since there are only a couple of layers at most, adding some logic to restrict the drawing of each layer shouldn't leave much impact.

The logic required to update the right pixels, given that the layer sizes can change over time, is a bit more complicated than I thought. Not only do we need to update the pixels between `min` and `max` in the large image buffer, but also the pixels on the canvas where the layer boundaries have moved or changed.

For svga, I'm thinking that we can use the layers to implement the `min` and `max`. I.e., instead of sending the `min` and `max` values, we can send a layer object representing the pixels that needs updating for svga.

ErnWong:
The last couple of commits changes the "screen.js api". Instead of having both `min`/`max` and `layers`, it  uses `layers` to do the job of `min`/`max` as well.

This way, it helps justify putting the layer handling code in `screen.js`. What do you think?

src/browser/screen.js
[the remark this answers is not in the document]
copy: Yeah, that makes sense.

src/vga.js
```diff
@@ -385,253 +514,342 @@ VGAScreen.prototype.set_state = function(state)
...
 
     if(this.graphical_mode)
     {
-        // TODO: Consider non-svga modes
-        this.set_size_graphical(this.svga_width, this.svga_height, this.svga_bpp);
+        if(this.svga_enabled)
+        {
+            this.set_size_graphical(this.svga_width, this.svga_height,
+                8, this.svga_width, this.svga_height);
+            this.update_layers();
```
copy: Does this restore the bpp correctly?

src/vga.js
```diff
@@ -740,115 +953,467 @@ VGAScreen.prototype.complete_redraw = function()
...
+{
+    this.screen_width = width;
+    this.screen_height = height;
+
+    this.stats.bpp = bpp;
+    this.stats.is_graphical = true;
+    this.stats.res_x = width;
+    this.stats.res_y = height;
+
+    this.bus.send("screen-set-size-graphical", [width, height, bpp, virtual_width, virtual_height]);
```
copy: bpp is ignored by the screen adapter, could you remove it?

copy: This is an expensive operation (reallocating a potentially large buffer), could we check if any of parameters have actually changed?

src/vga.js
ErnWong: Oops, nope it doesn't. I'll fix that up.

src/vga.js
ErnWong: Good idea. Btw, what's the original intention for the `screen_width` and `screen_height` variables, vs `svga_width` and `svga_height`? I've been using `svga_width/height` to represent the actual size of the screen area.

copy: `svga_width` is the current value of the corresponding register, it should be ignored if `svga_enabled` is false.

ErnWong: Ah, I see I see. I'll change `svga_width` to `screen_width` at places where I've used it wrong then.

ErnWong: It appears that `bpp` is listened for here:

https://github.com/copy/v86/blob/6440f7f9bcbf04535af8e3d8d0044a411d5640f9/src/browser/main.js#L972-L976

Should we keep `bpp` or still remove it from the bus message?

ErnWong:
For now, I've put back the `bpp` in the `screen-set-size-graphical` message because it was depended upon by `src/browser/main.js` for the runtime info. Let me know if you want me to take it back out.

copy:
@ErnWong Greatly appreciate your contribution, thanks!

ErnWong:
@copy It was a great pleasure :smile:

PR #180, opened 2018-01-02
Update copyright year to 2018

AmaanC:
Happy 2018!

IANAL, so I'm not sure how this stuff works, but I did manually verify that the project has had commits in every year (except 2012, which is the year the project started, so that's probably okay).


copy:
Cheers!

MailYouLater:
@AmaanC: You can delete your [patch-2](/AmaanC/v86/patch-2) branch now that it's been merged in.

PR #181, opened 2018-01-14
Allow the QEMU test to fail

BenNottelling:
I haven't seen this one work yet so it makes sense to me to allow this one to fail. (now we can add the badge without v86 looking bad because of builds "failing")

You see my issue here? Both PRs will build and work fine, but the VGA PR will be shown as failing because of a test that isn't affected by it.
screen shot 2018-01-14 at 1 45 59 am

---
This change is [Reviewable](https://reviewable.io/reviews/copy/v86/181)


copy:
That makes sense, thanks.

Issue #182, opened 2018-01-18
Pressing a key makes the mouse stop working


1j01:
Permanently, as far as I can tell.

BenNottelling:
Dumb question incoming, but... modifier keys? Like alt or control or option (mac key)? Are those modifier keys?

1j01:
Yes

BenNottelling:
Maybe a duplicate of #165?

1j01:
I don't think so; the keyboard still works, and blurring the page doesn't make the mouse start working again.

ErnWong:
Potentially related, but also potentially not.

@1j01 Mind telling us what browser you're using, and what OSes these problems appeared in?

ErnWong:
Gottit, I can reproduce it in Windows 98 with Chrome, pressing alt freezes the mouse.

1j01:
Yeah, Chrome and I tried several OSes, it seems to happen in any that have mouse support
Happens in Firefox as well...

And actually, I think it's just *any* key.

BenNottelling:
Can confirm it happens with any key, but only if you try to move the mouse while holding a key, otherwise it works fine

1j01:
I wonder if it's interrupting the mouse packets with keyboard packets.

ErnWong:
I'm not too familiar with the PS/2 controller and how interrupts work, so I might be wrong and saying silly things.

I think the problem is because there's nothing to lower the mouse interrupt other than reading port 0x60, so when the mouse interrupt handler decides to skip the read, the interrupt is left in a permanently raised state.

I think this is what's happening:

1. We first hold down a key while waving the mouse like Harry Potter's wand.

2. Both the [keyboard adapter](https://github.com/copy/v86/blob/master/src/browser/keyboard.js) and the [mouse adapter](https://github.com/copy/v86/blob/master/src/browser/mouse.js) start sending data to the [PS/2 controller](https://github.com/copy/v86/blob/master/src/ps2.js).

[...]

PR #183, opened 2018-01-18
Detect and trigger Alt-Key keyup, fixes #165

### LINKED ISSUE #165: ALT+Function keys freeze keyboard
(Browsers: Firefox 54, IE11)

Steps to reproduce:

- From the demo page, start MSDOS
- Press ALT-F2
- Try typing anything; you will notice nothing registers

Similarly, you can use the Linux 2.6 demo. Press ALT-F2 and then ALT-F1 again to return to the first terminal. Nothing can then be typed in this first terminal.

ErnWong:
Firefox 57 and Edge 41 on Windows (tested on Windows 10) do not emit keyup event for the alt key when the alt key is pressed with other keys, although Firefox on Ubuntu *does* appears to emit the event. Interesting how older versions of Firefox *do* seem to emit the event too. See #165.

This patch is tested on Firefox (Ubuntu Xenial in a VM and Windows 10) and Chrome. Haven't tested on Apple and with other browsers.

Possible bug with the web browsers? If it is, it might be better to close this PR and wait until they fix it rather add noisy code that adapts to their bugs... if we don't mind keeping v86 in a buggy state while we wait that is.

**Update:** Edge Alt keyup sometimes fires, sometimes doesn't. Strange. https://jsbin.com/vocehicipe/6/edit?html,js,output

Related: https://connect.microsoft.com/IE/feedback/details/2513563/no-keyup-event-for-alt-when-used-as-the-only-modifier-key

### DIFF
--- a/src/browser/keyboard.js
+++ b/src/browser/keyboard.js
@@ -328,6 +328,7 @@ function KeyboardAdapter(bus)
         if(!e.altKey && keys_pressed[0x38])
         {
             // trigger ALT keyup manually - some browsers don't
+            // see issue #165
             handle_code(0x38, false);
         }
         return handler(e, false);
@@ -338,6 +339,7 @@ function KeyboardAdapter(bus)
         if(!e.altKey && keys_pressed[0x38])
         {
             // trigger ALT keyup manually - some browsers don't
+            // see issue #165
             handle_code(0x38, false);
         }
         return handler(e, true);
--- a/src/browser/keyboard.js
+++ b/src/browser/keyboard.js
@@ -325,11 +325,21 @@ function KeyboardAdapter(bus)
 
     function keyup_handler(e)
     {
+        if(!e.altKey && keys_pressed[0x38])
+        {
+            // trigger ALT keyup manually - some browsers don't
+            handle_code(0x38, false);
+        }
         return handler(e, false);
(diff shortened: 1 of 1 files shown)


ErnWong:
Turns out it is a filed bug with Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1335347
Potentially this one too: https://bugzilla.mozilla.org/show_bug.cgi?id=809410

copy:
The patch looks good and safe, I think we should merge it even if it's a browser bug. Does this also fix #182?

src/browser/keyboard.js
```diff
@@ -325,11 +325,21 @@ function KeyboardAdapter(bus)
 
     function keyup_handler(e)
     {
+        if(!e.altKey && keys_pressed[0x38])
+        {
+            // trigger ALT keyup manually - some browsers don't
```
copy: Add a reference to #165

ErnWong:
> Does this also fix #182?

Nope, it does not appear to fix the mouse problem.

ErnWong:
> The patch looks good and safe, I think we should merge it even if it's a browser bug.

Sounds good!

copy:
Thanks!

PR #184, opened 2018-01-20
 Handle simultaneous keyboard and mouse streams

### LINKED ISSUE #182: Pressing a key makes the mouse stop working

ErnWong:
This fixes #182, by handling the following situations:

- Not all mouse/keyboard IRQ handlers check the status port to see whether the next port 0x60 read is going to return the mouse byte or keyboard byte. For example: [KolibriOS](http://websvn.kolibrios.org/filedetails.php?repname=Kolibri+OS&path=%2Fdrivers%2Fmouse%2Fcommouse.asm)

- Not all interrupts may result in reading port 0x60 to clear the interrupt itself. For example, the mouse IRQ handler for Windows 1 and 98 appears to skip reading port 0x60 if it finds out that the next byte
to be returned is a keyboard byte and not a mouse byte. The mouse IRQ thus never gets a chance to get cleared, and subsequent `device_raise_irq` do not actually trigger the handler.

- Between the time:
  ... after a mouse interrupt gets raised, or the OS reads the status port, and
  ... before the OS reads port 0x60,
more keyboard data may be received, and this race condition will make the OS read a keyboard byte when it is expecting a mouse byte, and vice versa. Thus, this patch ensures that incoming keyboard data will not mess up the `next_byte_is_aux` state, and vice versa with incoming mouse data. Instead, `next_byte_is_aux` is only toggled when the current buffer being read has been fully cleared.

Do you like this approach? I don't mind scrapping the changes I've made if you think there's a better method.

### DIFF
--- a/src/ps2.js
+++ b/src/ps2.js
@@ -53,6 +53,9 @@ function PS2(cpu, bus)
     /** @type {boolean} */
     this.next_read_resolution = false;
 
+    /** @type {boolean} */
+    this.next_byte_is_aux = false;
+
     /**
      * @type {ByteQueue}
      */
@@ -71,6 +74,9 @@ function PS2(cpu, bus)
 
     /** @type {number} */
     this.last_mouse_packet = -1;
+
+    /** @type {number} */
+    this.last_notify = 0;
 
     /**
      * @type {ByteQueue}
@@ -128,15 +134,17 @@ PS2.prototype.get_state = function()
     state[11] = this.next_handle_scan_code_set;
     state[12] = this.next_read_rate;
     state[13] = this.next_read_resolution;
-    //state[14] = this.kbd_buffer;
+    state[14] = this.next_byte_is_aux;
     state[15] = this.last_port60_byte;
     state[16] = this.sample_rate;
     state[17] = this.resolution;
     state[18] = this.scaling2;
-    //state[19] = this.mouse_buffer;
+    state[19] = this.last_notify;
     state[20] = this.command_register;
     state[21] = this.read_output_register;
     state[22] = this.read_command_register;
+    //state[23] = this.kbd_buffer;
+    //state[24] = this.mouse_buffer;
 
     return state;
 };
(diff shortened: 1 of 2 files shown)


ErnWong:
It still appears to freeze up the mouse in Windows 98 in certain conditions, leading to very jagged cursor movements. I'll have a look at it.

ErnWong:
Strange... this PR works in one windows 98 image, but not in another windows 98 image. The first image is the original image from https://copy.sh/v86. The second image is the result of putting the first image through VirtualBox to disable the svga driver and adding the sound card driver. The current problem it seems to face is that the keyboard irq is not being handled at all.

The code for the ps2 controller before this PR works fine with the second image (when key is pressed separately from mouse movement), so this PR currently has a regression.

ErnWong:
Everything looks fine now, except for one keyboard-affecting bug that I'm not sure how to fix. During Windows 98 startup, it regularly gets stuck on a black screen (similar to the Windows Protection Error before, but now without the words).

If, from this black screen, you reset the machine once, the keyboard does not appear to work anymore even after Windows 98 fully boots up. Upon closer inspection, it appears that the [configuration byte](https://wiki.osdev.org/%228042%22_PS/2_Controller#PS.2F2_Controller_Configuration_Byte) (aka the `command_register`) is not being reset/reconfigured properly (I think?) and the keyboard-interrupt-enable-bit (bit-0) remains disabled for the rest of the session.

However, if from the black screen, you press 'reset', and press 'reset' again after the bios screen loads halfway, the keyboard appears to work fine.

Looking at the logs, resetting once does not appear to write to port 0x64 of the PS/2 controller at all, whereas if you reset twice, the logs show that port 0x64 is being written to (during the BIOS stage I think, but I'm not sure).

I don't think I understand enough of the system to debug this further, so I think I'll leave this PR to where it is now and call it a day.

ErnWong:
Hi @copy I'd be keen for your thoughts on this PR if you're available :)

copy:
Sorry, I pushed this back several times. I am a bit torn about this PR. I agree that this is a bug that needs to be solved, but I don't think we should use timers to fix it (as hardware doesn't clear the buffers after a certain time – does it?). The regression also needs to be fixed before we can merge this.

That said, I don't know how to properly fix this and I don't have time to look into this right now. If you have another fix that'd be great. If not, I'm inclined to test and merge the current version (minus the regression), and look for a proper solution another day.

ErnWong:
> I am a bit torn about this PR. 

I understand this feeling, don't worry :) I'm sorry for creating this situation. Thanks for telling me it.

> but I don't think we should use timers to fix it (as hardware doesn't clear the buffers after a certain time – does it?)

I agree.

> The regression also needs to be fixed before we can merge this.

I believe the regression mentioned [here](https://github.com/copy/v86/pull/184#issuecomment-359996318) and [here](https://github.com/copy/v86/pull/184#issuecomment-360013201) has been fixed via b0cd300 so that's cool. The [other keyboard-affecting bug](https://github.com/copy/v86/pull/184#issuecomment-360048384) has already existed in the master branch before this PR and appears unrelated to this PR, so I think that's cool too.

> ...I don't have time to look into this right now. If you have another fix that'd be great.

I don't have an alternative fix just yet, but would you like me to look into it? I'd be happy to.

~~*...and then I can send you another hacky PR and slowly destroy the entire v86 project with badly written code without you knowing... [insert evil laugh]*~~

copy:
> I don't have an alternative fix just yet, but would you like me to look into it? I'd be happy to.

That would be great.

copy:
Closed as superseded by #189.

PR #185, opened 2018-01-24
Added an ACPI test for travis

BenNottelling:
This pull request adds..

- An ACPI test on travis
- The build badge for travis to the readme
- The build option to enable ACPI `useacpi=true` ex: `make build/libv86.js useacpi=true`

Thanks for the help with this guys, I'm still fairly new to developing. 
Side note: I committed this to my master branch, how can I undo this / have it be even with the master branch from @copy's branch? thanks for any help.

---
This change is [Reviewable](https://reviewable.io/reviews/copy/v86/185)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,6 @@ env:
     - TEST_SUITE=integration
     - TEST_SUITE=unit-qemu
     - TEST_SUITE=nasm
+matrix:
+  allow_failures:
+    - env: TEST_SUITE=unit-qemu


MailYouLater:
Unfortunately, since you referenced your master branch in this PR it'll keep updating this PR with any changes you make to your master branch.
There may be another way to deal with this issue (like... IDK, maybe try renaming your master branch? but that may have other consequences... I'm not sure) but my recommendation would be to use your master branch for this PR for the time being, then when done with this PR, you delete your master branch (on GitHub, not just locally), which will break this PR's connection with it, and then you recreate it (it won't magically reconnect to this PR) at whatever commit @copy's master branch is at, and you should be back to a more normal workflow.

PR #186, opened 2018-01-24
ACPI test for travis

BenNottelling:
This pull request adds..

- An ACPI test on travis
- The build option to enable ACPI `useacpi=true` ex: `make build/libv86.js useacpi=true`

- .DS_Store added to the .gitignore because it was annoying me

Thanks for the help with git, you guys are great

---
This change is [Reviewable](https://reviewable.io/reviews/copy/v86/186)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ images/
 *.img
 *.fixture
 *.fuse_hidden*
+*.DS_Store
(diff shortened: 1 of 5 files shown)


copy:
Thanks.

MailYouLater:
> * .DS_Store added to the .gitignore because it was annoying me

Well, we know someone has a Mac. 😉

On a more serious note, a change like this should really be in it's own commit, but oh well. It is what it is.

PR #189, opened 2018-02-09
Handle simultaneous keyboard and mouse streams - Alternative 2

ErnWong:
Alternative to PR #184 for fixing #182.

Two main changes to master branch:

1. Pulse interrupt lines to ensure rising edge.
    Supports for pulse:
    - [Under *Interrupt Processing*, paragraph 2, from here](http://www.os2museum.com/wp/ibm-ps2-model-50-keyboard-controller/)
    - [Paragraph 1 from here](https://wiki.osdev.org/8259_PIC#What_does_the_8259_PIC_do.3F)

    Against pulse:

    - [Under *Reading Keyboard Input*, paragraph 2, from here](http://www.computer-engineering.org/ps2keyboard). It says that the interrupt is cleared after reading port `0x60`. It might be referring to AT controllers instead of PS/2.

2. Before data in port `0x60` gets read, don't switch between `kbd` and `mouse` queues.
    In hardware terms: inhibit keyboard and mouse communication and don't overwrite the `DBBOUT` output buffer until `OBF` gets cleared.
    Useful refs: 
    - [Under *Reading Keyboard Input*, paragraph 1, from here](http://www.computer-engineering.org/ps2keyboard)
    - [Under *Main Entry*, line `0102`, from here](http://www.halicery.com/8042/8042_1503033.TXT)
    - [VirtualBox](https://www.virtualbox.org/browser/vbox/trunk/src/VBox/Devices/Input/DevPS2.cpp#L175) behaves this way and Kolibri OS works fine on it.
    - [Qemu](https://github.com/qemu/qemu/blob/0748b3526e8cb78b9cd64208426bfc3d54a72b04/hw/input/pckbd.c#L153-L181) doesn't seem to behave this way. Mouse can jump around and hang in Kolibri OS running on qemu due to the keyboard controller giving keyboard values to the mouse driver.
    - Testing KolibriOS on real machine: works :ok_hand:.

Difference between this PR and PR #184:
- Keyboard data gets priority over mouse data
- Works without timers! :)

### DIFF
--- a/src/ps2.js
+++ b/src/ps2.js
@@ -77,6 +77,14 @@ function PS2(cpu, bus)
      */
     this.mouse_buffer = new ByteQueue(1024);
 
+    /**
+     * @type {boolean}
+     * Also known as DBBOUT OBF - Output Buffer Full flag
+     */
+    this.next_byte_is_ready = false;
+
+    /** @type {boolean} */
+    this.next_byte_is_aux = false;
 
     this.bus.register("keyboard-code", function(code)
     {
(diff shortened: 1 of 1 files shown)


copy:
Thanks! This looks good to me, I tested it and it fixes the problem.

PR #191, opened 2018-02-14
Return correct VGA CRTC index value in 3D4 read

### LINKED ISSUE #190: FreeBSD text not being drawn
PR #178 introduced this regression.

### The problem

I can't find the root cause yet, but the immediate problem is that FreeBSD had set the map mask register (`plane_write_bm`) to `0` during some point during the boot process, so any write to the display memory doesn't get applied to any of the four planes.

### Why the bug showed up after PR #178, and not before

I checked: FreeBSD was already setting the map mask register to `0` before PR #178, but it did not lead to detectable problems because the `plane_write_bm` checking code was added after PR #178.

https://github.com/copy/v86/blob/ea5d1f54efe025902d3472a4b2bdfb501dfb6f52/src/vga.js#L662-L667

The reason why `plane_write_bm` checking code got added was because font data (which supposed to live in plane 2) would otherwise override the text data (which supposed to live in plane 0).

The reason why font data didn't override text data before PR #178 is because: (1) the old VGA controller code did not obey the [Memory Map Select field](http://www.osdever.net/FreeVGA/vga/graphreg.htm#06), and (2) the font data was written when Memory Map Select field was `0x0` (meaning the display buffer was mapped to memory address `0xA0000`) before it changed to `0x3` (where the display buffer was mapped to memory address `0xB8000`).

### tl;dr

Improvements to vga.js made an old bug surface.

We'll need to find out why the map mask register (`plane_write_bm`) is being set to `0x0` instead of `0x3`.

I'll start looking into these files:

https://github.com/freebsd/freebsd/blob/master/sys/dev/fb/vga.c
https://github.com/freebsd/freebsd/blob/master/sys/dev/fb/vgareg.h

ErnWong:
Fixes #190

### DIFF
--- a/src/vga.js
+++ b/src/vga.js
@@ -317,9 +317,9 @@ function VGAScreen(cpu, bus, vga_memory_size)
     io.register_read(0x3CC, this, this.port3CC_read);
 
     io.register_write_consecutive(0x3D4, this, this.port3D4_write, this.port3D5_write);
+    io.register_read(0x3D4, this, this.port3D4_read);
     io.register_read(0x3D5, this, this.port3D5_read);
 
-    io.register_read(0x3D4, this, function() { dbg_log("3D4 read", LOG_VGA); return 0; });
     io.register_read(0x3CA, this, function() { dbg_log("3CA read", LOG_VGA); return 0; });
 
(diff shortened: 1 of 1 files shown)


copy:
Thanks!

PR #196, opened 2018-02-18
Some SB16 and PC Speaker improvements

ErnWong:
- Reduces hearing loss. The PC speaker sounded much louder than everything else.
   (*Volume will be adjustable later when the mixer chip gets implemented.*)
- Addresses exceptions: overlapped IO reads and non-finite / out-of-range oscillator frequencies.
- Addresses some bugs and some sound stuttering.


copy:
Thanks!

PR #204, opened 2018-04-06
Add basic ANSI escape character control support

nununoisy:
Add support for readability in serial terminal

Commands:
^[J - normally clears screen to the right, simply ignored here
^[H^[J - home cursor, clear screen to the right - clears textarea

### DIFF
--- a/src/browser/serial.js
+++ b/src/browser/serial.js
@@ -53,7 +53,8 @@ function SerialAdapter(element, bus)
                 this.text = this.text.slice(0, -1);
                 this.update();
             }
-            else if(chr === "\x1b") {
+            else if(chr === "\x1b")
+            {
                this.control_mode = true;
                this.control_buffer = "";
             }
--- a/src/browser/serial.js
+++ b/src/browser/serial.js
@@ -13,6 +13,8 @@ function SerialAdapter(element, bus)
     this.bus = bus;
     this.text = "";
     this.text_new_line = false;
+    this.control_mode = false;
+    this.control_buffer = "";
 
     this.last_update = 0;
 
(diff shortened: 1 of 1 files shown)


src/browser/serial.js
```diff
@@ -42,28 +44,57 @@ function SerialAdapter(element, bus)
...
         else
         {
-            this.text += chr;
-
-            if(chr === "\n")
+            this.control_buffer += chr;
+            var cmds = {
+                bksp: /$\[.J^/,
+                clr: /$\[.H\x1B\[.J^/,
+                invalid: /$*[^HJ]^/
```
copy: These don't look quite right:

- `$` matches the end of the text and `^` the beginning
- The `*` in `invalid` probably requires something before it
- I believe  `invalid` would match before the other two; it needs to be more specific
- Instead of the `.`, we should probably use a more specific match. [This page](https://www.gnu.org/software/screen/manual/html_node/Control-Sequences.html) says it can be 1, 2 or 3, so `[123]` might be more appropriate

nununoisy: Whoops! I'm still getting the hang of regex. Silly mistakes.

src/browser/serial.js
```diff
@@ -42,28 +44,57 @@ function SerialAdapter(element, bus)
...
     {
-        if(chr === "\x08")
-        {
-            this.text = this.text.slice(0, -1);
-            this.update();
-        }
-        else if(chr === "\r")
-        {
-            // do nothing
+        if(this.control_mode === false) {
```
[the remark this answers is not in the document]
nununoisy: I tried so hard to match the code style, it's nothing like my own. Will fix.

nununoisy:
Alright. I've made the requested changes. Let me know what you think.

src/browser/serial.js
```diff
@@ -42,28 +44,57 @@ function SerialAdapter(element, bus)
...
+            }
+            else if(cmds.clr.test(this.control_buffer))
+            {
+                this.text = "";
+                this.control_mode = false;
+                this.update();
+            }
+            else if(cmds.invalid.test(this.control_buffer))
+            {
+                this.control_mode = false;
```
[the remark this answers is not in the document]
nununoisy: I can see that. Alright.

src/browser/serial.js
```diff
@@ -42,28 +44,59 @@ function SerialAdapter(element, bus)
...
-        else if(chr === "\r")
+        if(this.control_mode === false)
         {
-            // do nothing
+            if(chr === "\x08")
+            {
+                this.text = this.text.slice(0, -1);
+                this.update();
+            }
+            else if(chr === "\x1b") {
```
copy: Missing newline before `{`.

src/browser/serial.js
```diff
@@ -42,28 +44,59 @@ function SerialAdapter(element, bus)
...
         else
         {
-            this.text += chr;
-
-            if(chr === "\n")
+            this.control_buffer += chr;
+            var cmds = {
+                bksp: /^\[[0-2]?J$/,
+                clr: /^\[[0-9;]{0,5}H\x1b\[[0-2]?J$/,
+                invalid: /[hl=>0-2M-OmrA-Efg3-8KnRc8qy]/
```
copy: This is really hard to understand, can you find another way to find the end of the control buffer? If not, I can give this a shot.

Maybe a single regular expression could be used to match any sequence as described [here](https://en.wikipedia.org/wiki/ANSI_escape_code#CSI_sequences):

> The ESC [ is followed by any number (including none) of "parameter bytes" in the range 0x30–0x3F (ASCII 0–9:;<=>?), then by any number of "intermediate bytes" in the range 0x20–0x2F (ASCII space and !"#$%&'()*+,-./), then finally by a single "final byte" in the range 0x40–0x7E (ASCII @A–Z[\]^_`a–z{|}~).

nununoisy: We could go off of that list. I created that for compatibility with VT-100 controls as well. If need be I will fix it.

nununoisy: I might just wait for this PR to go through and write a new version of this file with screen memory so I can implement more commands

nununoisy:
Alright. Maybe this time. I’m working on a new branch with screen memory so I can implement complex commands.

copy:
Closing as no progress in 3 months.

Issue #216, opened 2018-06-22
[Safari] crash in speaker.js due to version issues in Web Audio API


humphd:
v86 crashes early during startup due to a Web Audio API issue in Safari:

```
TypeError: undefined is not an object (evaluating 'this.node_treble_left
        .connect(this.node_bass_left)
        .connect')
```

The problem is due to the method chaining here:

```js
    this.node_treble_left
        .connect(this.node_bass_left)
        .connect(this.node_gain_left)
        .connect(this.node_merger, 0, 0);
```

https://googlechrome.github.io/samples/webaudio-method-chaining/ also exhibits the same behaviour.

According to https://developer.mozilla.org/en-US/docs/Web/API/AudioNode/connect#Syntax:

> If the destination is a node, connect() returns a reference to the destination AudioNode object, allowing you to chain multiple connect() calls. In some browsers, older implementations of this interface return undefined.
>
> If the destination is an AudioParam, connect() returns undefined.

I'm not familiar enough with the Web Audio API to fix this, or I'd have sent a PR.  Would it be enough to do something like:

```js
var left = this.node_treble_left
left.connect(this.node_bass_left)
left.connect(this.node_gain_left)
left.connect(this.node_merger, 0, 0);
```

Or is the argument to the previous call supposed to get connected?

At any rate, this bug causes v86 to be unusable on Safari.

ErnWong:
Oops, this is my bad. Thanks for the report!

> ...Or is the argument to the previous call supposed to get connected?

Yup, the previous argument would've been returned to create a chain in series like this:

```
node_treble_left -> node_bass_left -> node_gain_left -> node_merger
```

I've made a quick PR (#218), but I don't have a Safari to test it on and I fear there could be other pieces from the Web Audio API that have similar problems with different browsers (I'll double check the docs).

@humphd would you like to checkout PR #218 and see if it works on Safari?

humphd:
Done.  Works great!  @ErnWong++

PR #218, opened 2018-06-23
Unchain AudioNode#connect methods to support Safari

ErnWong:
A proposed fix for #216 

Haven't had this tested on Safari ironically.
(Sadly I don't have access to a Safari on a desktop)


humphd:
I've tested this PR on macOS 10.13.5 in Safari (Safari 11.1.1), Chrome (67.0.3396.87), and Firefox (60.0.2) and am happy to report that it works in all cases.  Thanks for doing this fix so quickly.

copy:
Cheers!

PR #235, opened 2018-09-15
SSE2 instruction

davidbullado:
- Implement pmuludq
- Automated nasm tests : Add eight 128-bit SSE registers from xmm0 to xmm8


BenNottelling:
This build has failed because of an issue on google's end, so the failing test should be ignored atm

BenNottelling:
lgtm
I don't know how I could run the other tests , but this one worked

ElCresh:
Maybe coun be tired a "recent" version of Windows (gonna be slow but if nothing else is missing could start). Am I right?

davidbullado:
> Maybe coun be tired a "recent" version of Windows (gonna be slow but if nothing else is missing could start). Am I right?

That would be nice, but many SSE2 & SSE1 instructions still need to be implemented :) This PR is just a first step, if its accepted I could definitively work on this :)

davidbullado:
As it is my first PR in the project, I would like a feedback from the maintainers before implementing more :)

davidbullado:
I just found some issues. I will close the PR and correct them.

davidbullado:
Should be better :)

PR #241, opened 2018-09-18
Implementation of cvtpi2ps

davidbullado:
+ Implement instruction cvtpi2ps (sse1)
+ Improve output of nasm tests
+ Minor adjustments


src/arith.js
```diff
@@ -1627,3 +1627,32 @@ CPU.prototype.saturate_uw = function(v)
...
+        // rc=3 is truncate -> floor for positive numbers
+    else if(rc === 1 || (rc === 3 && f > 0))
+    {
+        return Math.floor(f);
+    }
+    else
+    {
+        return Math.ceil(f);
+    }
+}
```
copy: Add a trailing newline here.

src/arith.js
```diff
@@ -1627,3 +1627,32 @@ CPU.prototype.saturate_uw = function(v)
     dbg_assert(v >= 0);
     return v > 0xFFFF ? 0xFFFF : v;
 };
+
+CPU.prototype.integer_round = function(f)
```
copy: This is from fpu.js, isn't it? Could you refactor the code in fpu.js to use this bit?

src/instructions.js
```diff
@@ -2089,11 +2089,140 @@ t[0x29] = cpu => {
...
-t[0x2D] = cpu => { cpu.unimplemented_sse(); };
-t[0x2E] = cpu => { cpu.unimplemented_sse(); };
+t[0x2A] = cpu => { 
+    // cvtpi2ps xmm, mm/m64
+    dbg_assert((cpu.prefixes & (PREFIX_MASK_REP | PREFIX_MASK_OPSIZE)) === 0);
+    cpu.task_switch_test_mmx();
+    cpu.read_modrm_byte();
+    let data = cpu.read_mmx_mem64s();
+    let float32 = new Float32Array(2) ;
+    let res32 = new Uint32Array(float32.buffer) ;
```
copy: Unnecessary space before `;`.

tests/nasm/cvtpi2ps.asm
```diff
@@ -0,0 +1,22 @@
...
+quad1high:
+	dd	0
+myaddress:
+	dd	0xdeadbeef
+%include "header.inc"
+
+	cvtpi2ps	xmm0, [quad0low]
+	; fill xmm1 in order to ensure that the high quadword remain inchanged
+	pshufd      xmm1, xmm0, 0    
+	cvtpi2ps	xmm1, [quad1low]
```
copy: Please don't use tabs.

tests/nasm/run.js
```diff
@@ -172,26 +172,28 @@ else {
...
                         actual: evaluated_mmxs[i],
                         expected: test.fixture_array[i]
                     });
                 }
                 json_index++;
             }
-
-            for (let i = 0; i < evaluated_xmm32.length; i++) {
-                if (evaluated_xmm32[i] !== test.fixture_array[json_index]) {
+
```
copy: Remove trailing spaces.

src/arith.js
davidbullado: Yes it does. I have refactored the code in *fpu.js*, with an encapsulation of *cpu.integer_round* to minimize the number of impacted lines. (But at the cost of an extra function call, which is not necessarily a good thing in terms of optimization, in my opinion.)

tests/nasm/cvtpi2ps.asm
davidbullado: It seems that the other nasm tests are using tabs instead of spaces. Perhaps I should replace the spaces by tabs, to be consistent with the other tests ?

PR #285, opened 2019-05-05
Use ES intrinsics for mul / imul, bsr / bsf

iamahuman:
For integer multiplication, direct product of two 32-bit integers in
double precision FP arithmetic yields its higher 53 bits which are
sufficient for higher half of the product, while the lower half may be
computed with `Math.imul()`. A polyfill is provided for < IE 10.

For `floor(log2(n))` where `n` is a 32-bit integer, `31 - Math.clz32(n)`
is an accurate substitute for `v86util.int_log2` and preserves
`int_log2(0) = -1` (should it ever occur).

### DIFF
--- a/src/instructions.js
+++ b/src/instructions.js
@@ -4841,24 +4841,24 @@ t[0xF4] = cpu => {
 
         let i = (cpu.modrm_byte >> 3 & 7) << 2;
 
-        let result = cpu.do_mul32(destination[0] , source[0]);
-        cpu.reg_xmm32s[i] = result[0];
-        cpu.reg_xmm32s[i + 1] = result[1];
+        let d = destination[0];
+        let s = source[0];
+        cpu.reg_xmm32s[i] = v86util.imul(d, s);
+        cpu.reg_xmm32s[i + 1] = v86util.imul_high(d >>> 0, s >>> 0);
 
-        result = cpu.do_mul32(destination[2] , source[2]);
-        cpu.reg_xmm32s[i + 2] = result[0];
-        cpu.reg_xmm32s[i + 3] = result[1];
+        d = destination[2];
+        s = source[2];
+        cpu.reg_xmm32s[i + 2] = v86util.imul(d, s);
+        cpu.reg_xmm32s[i + 3] = v86util.imul_high(d >>> 0, s >>> 0);
     }
     else
     {
         // pmuludq mm1, mm2/m64
         dbg_assert((cpu.prefixes & (PREFIX_MASK_REP | PREFIX_MASK_OPSIZE)) == 0);
-        let source64s = cpu.read_mmx_mem64s();
-        let destination_low = cpu.reg_mmxs[2 * (cpu.modrm_byte >> 3 & 7)];
+        let s = cpu.read_mmx_mem64s()[0];
+        let d = cpu.reg_mmxs[2 * (cpu.modrm_byte >> 3 & 7)];
 
-        let result = cpu.do_mul32(destination_low,source64s[0])
-
-        cpu.write_mmx64s(result[0], result[1]);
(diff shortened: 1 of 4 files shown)


BenNottelling:
Did you test your changes? `do_mul32` is also called in these locations of `instructions.js`, causing the build to fail (and those instructions as well)
https://github.com/copy/v86/blob/master/src/instructions.js#L4844
https://github.com/copy/v86/blob/master/src/instructions.js#L4848
https://github.com/copy/v86/blob/master/src/instructions.js#L4859

iamahuman:
after push i immediately noticed it and am fixing it, just a moment

>

iamahuman:
@BenNottelling it's just ~5 lines and maybe better than nothing :p

I was coding in my phone and I didn't have time to port i386 toolchains to the platform (although I do have node.js). More often than not when I contribute to projects like this without a laptop in my reach I find myself offloading all the test work to the CI. Sorry for that ;(

BenNottelling:
@iamahuman Haha it's alright, tests passed, now it's up to copy to check and see if he has any changes requested!
(I'm just here to pretend I know things and spam everyone :P)

ErnWong:
@iamahuman Coding up a pull request from a phone? That's pretty awesome :smile:

iamahuman:
@ErnWong nah just lazy me who don't even want to pull my laptop off my bag

src/instructions.js
```diff
@@ -4841,24 +4841,26 @@ t[0xF4] = cpu => {
...
         dbg_assert((cpu.prefixes & (PREFIX_MASK_REP | PREFIX_MASK_OPSIZE)) == 0);
         let source64s = cpu.read_mmx_mem64s();
         let destination_low = cpu.reg_mmxs[2 * (cpu.modrm_byte >> 3 & 7)];
+        let source = source64s[0];
 
-        let result = cpu.do_mul32(destination_low,source64s[0])
-
-        cpu.write_mmx64s(result[0], result[1]);
+        cpu.write_mmx64s(v86util.imul(destination_low, source),
+                         (destination_low >>> 0) * (source >>> 0) / 0x100000000);
```
copy: Could you move this expression into a function? `v86util.imul_high` or so? (also above)

src/instructions.js
```diff
@@ -4841,24 +4841,26 @@ t[0xF4] = cpu => {
...
+            let d = destination[j];
+            let s = source[j];
+            let x = i + j;
 
-        result = cpu.do_mul32(destination[2] , source[2]);
-        cpu.reg_xmm32s[i + 2] = result[0];
-        cpu.reg_xmm32s[i + 3] = result[1];
+            cpu.reg_xmm32s[x] = v86util.imul(d, s);
+            cpu.reg_xmm32s[x + 1] = (d >>> 0) * (s >>> 0) / 0x100000000;
+        }
```
copy: Not a fan of this for loop, could you undo the change?

src/lib.js
```diff
@@ -131,6 +131,38 @@ SyncBuffer.prototype.get_buffer = function(fn)
 
 (function()
 {
+    var clz32 = Math.clz32;
+    if(typeof clz32 === 'function' && clz32(0) === 32 &&
```
copy: Use double quotes.

src/instructions.js
iamahuman: @copy I was not sure if I should write separate versions for signed and unsigned or delegate that matter to the caller. Any feedback is appreciated.

src/instructions.js
iamahuman: Unrolled.

src/lib.js
iamahuman: ...or maybe we can just get rid of this polyfill at all?

copy:
Cheers!

Issue #283, opened 2019-04-30
Destroying emulator


filips123:
Is it possible to destroy emulator programmatically? With destroying I mean removing all adapters (mouse and keyboard should be accessible again) and cleaning up memory.

This is useful if the site needs to load multiple systems, but not at the same time. Currently, the only possible solution is to reload the site.

iamahuman:
`v86.reset();`

filips123:
@iamahuman What does this do? It is not in docs.

I want to remove all adapters and cleanup memory.

copy:
There's a partial implementation here: https://github.com/copy/v86/blob/eb736ab406c87e55e156beb7f889372a4dbfaf92/src/browser/starter.js#L459, but it needs to be extended to also destroy the other adapters. Pull requests are welcome.

filips123:
I can submit PR for other adapters.
What adapters are used? Do all of them have `destroy` method?

Main problem with destroying is probably to clean up memory. How can this be done?

copy:
> I can submit PR for other adapters.
> What adapters are used? Do all of them have destroy method?

Have a look at the code. IIRC most of them have a `destroy` but I'm not sure if they're complete. You'll need to check manually.

> Main problem with destroying is probably to clean up memory. How can this be done?

It should be garbage collected automatically as soon as all emulator objects are unreachable. You can check this in Chromiums developer tools by taking a heap snapshot in the memory tab and looking for an arraybuffer object that is larger than 128 MB.

iamahuman:
@filips123 Are there any other references to ArrayBuffer objects? Even they are normal objects and any competent Javascript engine would garbage-collect it once it decides it can never be reached to by user code.

filips123:
@iamahuman I think no. But maybe this is because adapters are still used.

iamahuman:
@filips123 wait, I think I missed DOM event listeners whose closures do need to still refer to them. IMHO they can still be cut out by removing either the listeners with `removeEventListener` or the attached nodes altogether.

iamahuman:
@filips123 I think the best way to figure it out would be to use a profiler that can enumerate all references to a specified object or value on runtime.

filips123:
@iamahuman I currently don't have much time to do this. I will probably wait until #290 is merged to see if references are deleted.

---

If you also want to test this, I created simple emulator on [GitHib Gist](https://gist.github.com/filips123/1d0807f77b35182550c968ac40e66105).

[...]

PR #290, opened 2019-05-29
Add other adapters to destroy method

filips123:
I've added other adapters to the `destroy` method. I haven't added `SpeakerAdapter` because it doesn't have the `destroy` method. I also didn't add `ModemAdapter` because it is not used.

This is related to #283. However, this issue shouldn't be closed because of https://github.com/copy/v86/issues/283#issuecomment-496960821.


copy:
See https://github.com/copy/v86/commit/9211c4fc876c436a1976fdd64ce69da83d29cbdd

Issue #217, opened 2018-06-23
Tabulator issue in lua example


niansa:
I get a list of all linux commands if i use multiple tabulators in the examples/lua.html

mprey:
Could you explain this more? I would like to help.

niansa:
https://copy.sh/v86/examples/lua.html <-- simply put 2 tabulators into the code textarea

niansa:
![image](https://user-images.githubusercontent.com/20720267/68681232-5be72e80-0563-11ea-98df-39d2b178d033.png)

copy:
Fixed in https://github.com/copy/v86/commit/131f4e128b957771713d567e3128d4bdf409130b.

PR #303, opened 2019-11-12
Fix tabulators from escaping CLI (#217)

mprey:
Previously, tabulators (\t) would escape the CLI input into `lua -e` and cause unknown output. See #217 

To remedy, the input must be escaped from tabs before escaping special characters into quotes. 

Tested inputs with correct output:
```
	k = 1
	x = 0

while k < 1000 do
    x = x + 1 / (k * k)
    k = k + 2
end

print(math.sqrt(x*8))

	function factorial(n)
   	if n == 0 then
        return 1
    else
    		return n * factorial(n - 1)
    	end
end

print("factorial(10):", factorial(10))
```
`	` (\t)
`		` (\t\t)
```
	print("1")
		
print("2")
```

### DIFF
--- a/examples/lua.html
+++ b/examples/lua.html
@@ -86,6 +86,7 @@ window.onload = function()
 // Implement bash string escaping.
 function bashEscape(arg)
 {
+    arg = arg.replace(/\t+/g, "");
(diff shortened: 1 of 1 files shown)


copy:
Cheers!

PR #307, opened 2020-01-08
Build libv86 with GitHub Actions

### LINKED ISSUE #297: Automatically build libv86 on CI
Can you automatically build `libv86.js` on Travis CI at every commit and publish it to some server or GitHub Pages? This would be very good because users then won't need to install and configure Closure Compiler and other build dependencies.

Can you also configure `Makefile` to also build sourcemap file for `libv86.js`?

Also, the version on [your website](https://copy.sh/v86/) is very old (from 2018). Can you also update it?

filips123:
This closes #297. It automatically builds libv86 with Travis CI and publishes it to GitHub Pages.

I only added ACPI build, but normal build can also be added if needed. File, as well as sourcemap, will then be published to `gh-pages` branch.

However, compaling fails because of some error (at least on Travis CI on my repository):

```
java.lang.NullPointerException: No warning class for name: ambiguousFunctionDecl

	at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:980)

	at com.google.javascript.jscomp.DiagnosticGroups.setWarningLevel(DiagnosticGroups.java:703)

	at com.google.javascript.jscomp.AbstractCommandLineRunner.setWarningGuardOptions(AbstractCommandLineRunner.java:327)

	at com.google.javascript.jscomp.AbstractCommandLineRunner.setRunOptions(AbstractCommandLineRunner.java:368)

	at com.google.javascript.jscomp.AbstractCommandLineRunner.doRun(AbstractCommandLineRunner.java:1107)

	at com.google.javascript.jscomp.AbstractCommandLineRunner.run(AbstractCommandLineRunner.java:546)

	at com.google.javascript.jscomp.CommandLineRunner.main(CommandLineRunner.java:2131)
```

It is interesting that even without any modifications to `Makefile` (in final version I did some modifications to include sourcemap), it fails. So it may be problem with newer version of Closure Compiler.


copy:
I don't think a branch is a good place to store compiled artefacts. Doesn't Travis have some sort of storage where artefacts can be stored?

filips123:
Travis has [artefacts storage](https://docs.travis-ci.com/user/uploading-artifacts/) but it requires Amazon S3 account. Some other solution (maybe even better) may be to provide libv86 as npm package.

copy:
How about a Github release? Something like https://github.com/marketplace/actions/automatic-releases

filips123:
This would probably work. So there would always be `latest` release which would contain libv86.

However, I think it would be also good to publish it to npm so it would be easier for other packages to include and update it, as well as some proper versioning, so packages will still be able to use old and download old versions, maybe in case of backwards-incompatible changes.

filips123:
@copy Sorry for delay. I switched to GitHub Actions for building libv86 and then releasing it to GitHub Releases.

copy:
@filips123 Manually rebased and merged into master, thanks! Latest release is here: https://github.com/copy/v86/releases

Github Actions looks pretty solid, I think eventually we should also run tests there.

filips123:
@copy Nice. Can you modify GitHub workflow to also release `libv86.js.map` file? I forgot to add it while I was creating that changes, but I think it would be quite useful to have it, especially if you are developing and debugging site that uses v86.

Workflow file probably just needs to be updated to include:
```yaml
files: |
  build/*.js
  build/*.js.map
```

copy:
@filips123 Done.

PR #310, opened 2020-02-10
Don't assert on unknown MSR, raise #GP0 instead.

basicer:
Some builds of the kernel probe for MSRs and cactch the GP0 during boot.
This allows some kernels that to boot that previously would not.

### DIFF
--- a/src/instructions.js
+++ b/src/instructions.js
@@ -2324,7 +2324,8 @@ t[0x30] = cpu => {
             break;
 
         default:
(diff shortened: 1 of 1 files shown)


copy:
Note that this only affects debug mode. I'm a fan of failing quickly. v86 has a lot of debug output, making it easy to miss log messages. Can you instead stub only those MSRs that the kernel actually probes?

copy:
I'll close this for now. Feel free to reopen with specific MSRs.

PR #329, opened 2020-03-22
use specific version of closure

Hackerjef:
Using an older version of closure to allow builds to succeed instead of using lastest version of closure


copy:
Thanks!

PR #367, opened 2020-09-05
Don't release ACPI builds, also include v86_all.js 

BenNottelling:
It doesn't make sense to release ACPI builds when they aren't the default, also, release `v86_all.js` so people can use the included `index.html` file.

---
This change is [Reviewable](https://reviewable.io/reviews/copy/v86/367)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Build v86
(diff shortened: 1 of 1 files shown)


copy:
The acpi change is fine.

I'm not sure about releasing v86_all.js. It will appear as a stand-alone JavaScript on the releases page without context. I think people who want to use the html files in the repo should be able to compile it themselves or use debug.html.

BenNottelling:
Done.

Issue #332, opened 2020-03-26
Cannot boot 9front due to APIC issues


ddevault:
ISO from here:

http://9front.org/iso/

It almost boots, but runs into this:

https://code.9front.org/hg/plan9front/file/d416b601a9df/sys/src/9/pc/mp.c#l167

copy:
Done: https://copy.sh/v86/?profile=9front

Feel free to send me bootable disk image if you'd like to add anything.

ddevault:
Nice, thanks!

PR #368, opened 2020-09-06
Enable APIC without ACPI

BenNottelling:
Related to https://github.com/copy/v86/issues/332
ACPI and APIC are not the same and don't need to be bundled together. 
Now 9front reaches this point (requiring input I don't understand, but seemilying working) 
image
V.S. the behavior described in the issue
image

### DIFF
--- a/src/cpu.js
+++ b/src/cpu.js
@@ -715,10 +715,11 @@ CPU.prototype.init = function(settings, device_bus)
         this.devices.pic = new PIC(this);
         this.devices.pci = new PCI(this);
 
-        if(ENABLE_ACPI)
+        this.devices.ioapic = new IOAPIC(this);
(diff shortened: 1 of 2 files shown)


copy:
I don't think we can unconditionally enable apic, its implementation is not complete and it might break other OSes.

BenNottelling:
> I don't think we can unconditionally enable apic, its implementation is not complete and it might break other OSes.

Very well, I'll look into potentially completing the implementation if it's not too difficult

lc67846:
Yeah I Searched About It And It Dis Not Break Anything

copy:
ACPI+APIC is now configured at runtime by giving `acpi: true` to V86Starter. It's used in the 9front profile only for now.

PR #378, opened 2020-11-29
Ne2k Driver Fixes to Support Running in Windows!

ikreymer:
I believe this fix should enable the use of the network card under Windows! I've tested it in Win98 so far.

The fix is includes two components, both based simply on comparing to the logic in the working qemu implementation.

1. The initial buffer setup appears to be incorrect. The 0x57 byte should be set on bytes 28 through 31.
This is simply by comparing to what qemu does in: https://github.com/qemu/qemu/blob/master/hw/net/ne2000.c#L130
I think this is the crux of the fix!

2. The write logic is also slightly off. In qemu's implementation, the command register is always set, and then bitwise op is applied if condition is met: https://github.com/qemu/qemu/blob/master/hw/net/ne2000.c#L281
Also changed an | to and & to match.

I am no expert on this card, basically just compared to what QEMU was doing here to see if I could find any difference, and this is what I found.

I am not sure if 2. is necessary, maybe just 1. is enough to get this to work, but I have been testing with both.

Note: The card is detected as a `Realtek RTL8029(AS) PIC Ethernet NIC` not as NE2000 Compatible, the same as in QEMU. Initially installed the drivers in QEMU and was able to reinstall in v86 with this fix.!

Here's a screenshot of running IE6 in Windows98 with the provided network. (Yes, you can still access msn.com with IE6!)
Screen Shot 2020-11-29 at 2 00 27 PM

### DIFF
--- a/src/ne2k.js
+++ b/src/ne2k.js
@@ -125,9 +125,10 @@ function Ne2k(cpu, bus)
         this.memory[i << 1] = this.memory[i << 1 | 1] = mac[i];
     }
 
-    for(var i = 28; i < 32; i++) {
-      this.memory[i] = 0x57;
-    }
+    // the PROM signature of 0x57, 0x57 is also doubled
+    // resulting in setting the 4 bytes at the end, 28, 29, 30 and 31 to 0x57
+    this.memory[14 << 1] = this.memory[14 << 1 | 1] = 0x57;
+    this.memory[15 << 1] = this.memory[15 << 1 | 1] = 0x57;
 
     dbg_log("Mac: " + h(mac[0], 2) + ":" +
                       h(mac[1], 2) + ":" +
--- a/src/ne2k.js
+++ b/src/ne2k.js
@@ -125,7 +125,9 @@ function Ne2k(cpu, bus)
         this.memory[i << 1] = this.memory[i << 1 | 1] = mac[i];
     }
 
-    this.memory[14] = this.memory[15] = 0x57;
+    for(var i = 28; i < 32; i++) {
+      this.memory[i] = 0x57;
+    }
 
     dbg_log("Mac: " + h(mac[0], 2) + ":" +
                       h(mac[1], 2) + ":" +
@@ -152,7 +154,7 @@ function Ne2k(cpu, bus)
 
     io.register_write(this.port | E8390_CMD, this, function(data_byte)
     {
-        this.cr = data_byte & ~4;
+        this.cr = data_byte;
         dbg_log("Write command: " + h(data_byte, 2) + " newpg=" + (this.cr >> 6) + " txcr=" + h(this.txcr, 2), LOG_NET);
 
         if(this.cr & 1)
@@ -160,7 +162,7 @@ function Ne2k(cpu, bus)
             return;
         }
 
-        if((data_byte | 0x18) && this.rcnt === 0)
+        if((data_byte & 0x18) && this.rcnt === 0)
         {
             this.do_interrupt(ENISR_RDC);
         }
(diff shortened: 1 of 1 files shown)


ikreymer:
Can send the hd image for the above if you'd like to test/confirm.

copy:
Nice work.

> Can send the hd image for the above if you'd like to test/confirm.

That would be useful. More importantly, what steps did you take to configure networking on Windows 98? I tried some time ago and failed to get it working even in qemu.

ikreymer:
> That would be useful. More importantly, what steps did you take to configure networking on Windows 98? I tried some time ago and failed to get it working even in qemu.

Ah yes, that took some time to find also! The key is to have Windows recognize the PCI cards is to change the `Plug and Play BIOS` to `PCI Bus`. (It has an exclamation mark in your Win98 image). Then, Windows will start detecting all of the PCI devices. I did this first in QEMU and installed the card there.

A longer explanation of it is here: https://cyberpillar.com/dirsver/1/mainsite/software/opsys/fullos/msopsys/runwncod/mswindws/mswnuser/mswin98/msw98ins/msw98ins.htm#w98pcidr

BenNottelling:
> The card is detected as a `Realtek RTL8029(AS) PIC Ethernet NIC` not as NE2000 Compatible, the same as in QEMU. Initially installed the drivers in QEMU and was able to reinstall in v86 with this fix.!

From this page: https://rtitwww.realtek.com/en/products/communications-network-ics/item/rtl8029as

> The RTL8029AS controller is an NE2000 compatible Ethernet Controller for PCI interface
So, a non-issue or concern.

Also, I've been working on a ReactOS for v86 (you can find it [here](https://ros.bennottelling.com/?profile=reactos)), but noticed that the network would often work for a little bit and then stop. I'm yet to test it, but I'm wondering if the PR has potentially fixed that issue?

src/ne2k.js
```diff
@@ -125,7 +125,9 @@ function Ne2k(cpu, bus)
         this.memory[i << 1] = this.memory[i << 1 | 1] = mac[i];
     }
 
-    this.memory[14] = this.memory[15] = 0x57;
+    for(var i = 28; i < 32; i++) {
```
copy, line 128: Insert newline before `{`

src/ne2k.js
```diff
@@ -125,7 +125,9 @@ function Ne2k(cpu, bus)
...
+      this.memory[i] = 0x57;
+    }
```
ErnWong, line 130: I'm not an expert in this either, but I think QEMU seems to be setting `0x57` into bytes 14 and 15 as well as bytes 28-31, but here we seem to be setting `0x57` only into bytes 28-31?

Since you were trying to copy over QEMU's behaviour, I thought it might be good to double check whether this deviation from QEMU was intentional.

<sub>(I didn't spend much time reading QEMU's source code, so I could be wrong)</sub>

ikreymer: My understanding is that the PROM is 'doubled' for 16-bit cards, which apparently means the same byte is duplicated for a 2-byte word?

So:
```
byte[2 * 14] = byte[2 * 14 + 1] = 0x57
byte[2 * 15] = byte[2 * 15 + 1] = 0x57
```
(Maybe it should be written out like that instead of the loop)

In QEMU, this doubling is a applied for all bytes in reverse so: 
```
byte[14] == byte[7 * 2] 
byte[15] == byte[7 * 2 + 1]
```
The value of byte[14] and byte[15] ends up being byte[7] *after* the doubling.

A quick search on this confirms seems to confirm the doubling (includes an example 32-byte PROM):
https://tldp.org/HOWTO/Ethernet-HOWTO-2.html

Comment in linux kernel: https://github.com/torvalds/linux/blob/master/drivers/net/ethernet/8390/ne2k-pci.c#L199

ErnWong: Gotcha. I see, I must have misread QEMU's loop (I didn't realise it overwrites bytes 14 and 15 during the doubling - very sneaky). Thanks, I learnt something new today.

src/ne2k.js
ikreymer: See updated, removed the loop and added comment for clarity, think that may be better to explain what's going on.

copy:
Thanks!

PR #379, opened 2020-11-30
Replaced the scan code for Home with the correct scan code.

mat-sz:
Before the change, the scan code for Home was the same as for End. I've replaced the lower byte with a correct one (0x47) as per [this page](https://www.fountainware.com/EXPL/bios_key_codes.htm).


copy:
Thanks!

PR #382, opened 2020-12-07
Network Customization, Improved Save/Restore, misc fixes

ikreymer:
Includes a few fixes for networking as well as a couple additional fixes.

The networking exposes support for a custom `network_adapter` which can be used in place of the `network_relay_url` as well as an optional fixed mac address with `network_mac`.

networking:
- support for custom 'network_adapter' instead of 'network_relay_url' for using custom network adapter
- network state save/restore: save additional registers, full memory buffer (32k)
- support fixed mac address with 'network_mac' init option

sound init: to avoid undefined var, set 'DEBUG = false' in sound worker if DEBUG is false
starter: ensure 'initial_state' is added to settings from V86Starter constructor

### DIFF
--- a/src/browser/speaker.js
+++ b/src/browser/speaker.js
@@ -742,10 +742,8 @@ function SpeakerWorkletDAC(bus, audio_context, mixer)
     var worklet_code_end = worklet_string.lastIndexOf("}");
     var worklet_code = worklet_string.substring(worklet_code_start, worklet_code_end);
 
-    if(DEBUG)
-    {
-        worklet_code = "var DEBUG = true;\n" + worklet_code;
-    }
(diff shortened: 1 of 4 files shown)


copy:
Thanks! I tested this and got networking working in my Windows 98 image, even though multiple VMs will be sharing a mac address using this method (for W98 this is the only option, since it can't reload the ne2k driver without a reboot, unlike the other OSes). I'll include clean up and commit this change in a separate PR.

copy:
These changes are now included in the master branch. The `initial_state` change and the `worklet_code` change have been omitted since similar fixes were already present in the recent merge, but let me know if anything broke for you. The `network_mac` option has been replaced by a `preserve_mac_from_state_image` option which should have a similar effect.

PR #419, opened 2021-03-22
Assorted Docker and makefile fixes

aidanhs:


Makefile
```diff
@@ -18,9 +20,9 @@ STRIP_DEBUG_FLAG=--v86-strip-debug
 endif
 
 default: build/v86-debug.wasm
-all: build/v86_all.js build/libv86.js build/v86.wasm
+all: build/v86_all.js build/libv86.js build/v86.wasm build/xterm.js
 all-debug: build/libv86-debug.js build/v86-debug.wasm
-browser: build/v86_all.js
+browser: build/v86_all.js build/xterm.js
```
copy, line 25: I'd prefer if xterm.js stayed optional — the code should handle a 404 fine (but let me know if it doesn't).

gen/util.js
```diff
@@ -18,7 +18,12 @@ function hex(n, pad)
 
 function mkdirpSync(dir)
 {
-    fs.mkdirSync(dir, { recursive: true });
+    let stat = fs.statSync(dir, { throwIfNoEntry: false });
+    if (stat === undefined) {
+        fs.mkdirSync(dir, { recursive: true });
+    } else if (!stat.isDirectory()) {
+        throw 'path exists but is not dir';
+    }
```
copy, line 26: Should be written in [EAFP](https://docs.python.org/3/glossary.html#term-eafp) style. Please also use the same style as the surrounding code (double quotes, newline before `{`, `const` over `let`, etc.). Personally, I think the existing error is good enough, so feel free to remove this commit.

tools/docker/test-image/Dockerfile
```diff
@@ -1,21 +1,24 @@
...
 
 RUN \
+        export DEBIAN_FRONTEND=noninteractive && \
         dpkg --add-architecture i386 && \
         apt-get update -qq && \
-        apt-get install -y nasm gdb unzip p7zip-full openjdk-8-jre wget python python3 qemu-system-x86 git-core build-essential libc6-dev-i386-cross libc6-dev-i386 && \
-        wget https://nodejs.org/dist/v8.9.4/node-v8.9.4-linux-x64.tar.xz && \
-        tar xfv node-v8.9.4-linux-x64.tar.xz && \
-        rm node-v8.9.4-linux-x64.tar.xz && \
+        apt-get dist-upgrade -y
```
copy, line 7: Is the dist-upgrade necessary?

tools/docker/test-image/Dockerfile
```diff
@@ -1,21 +1,24 @@
...
         export PATH="$HOME/.cargo/bin:$PATH" && \
         rustup toolchain install nightly && \
         rustup default nightly && \
         rustup target add wasm32-unknown-unknown --toolchain nightly && \
-        rustup component add rustfmt-preview --toolchain nightly && \
-        apt-get clean && \
-        apt-get autoclean && \
-        apt-get autoremove && \
-        rm -rf /var/lib/apt/lists/*
+        rustup component add rustfmt-preview --toolchain nightly
```
copy, line 24: I'm no docker expert, but the following were intentional to keep the image small:

- Run all commands in a single `RUN` block
- Run the `apt clean` at the end.

Please let me know if there are good reasons for this change, or revert them otherwise.

tools/docker/test-image/Dockerfile
```diff
@@ -1,21 +1,24 @@
...
+        apt-get dist-upgrade -y
+RUN \
+        export DEBIAN_FRONTEND=noninteractive && \
+        apt-get install -y nasm gdb unzip p7zip-full openjdk-8-jre wget python python3 qemu-system-x86 git-core build-essential libc6-dev-i386-cross libc6-dev-i386 clang curl
+RUN \
+        NODEVSN=v14.16.0 && \
+        wget https://nodejs.org/dist/$NODEVSN/node-$NODEVSN-linux-x64.tar.xz && \
+        tar xfv node-$NODEVSN-linux-x64.tar.xz && \
+        rm node-$NODEVSN-linux-x64.tar.xz && \
+        echo 'export PATH="$PATH:/node-$NODEVSN-linux-x64/bin"' >> ~/.bashrc && \
```
copy, line 16: I think it might be better to use the version of Node that comes with Ubuntu 20.04 these days. Could you try that? I'll fix any scripts that are broken (if any) afterwards.

copy (changes requested):

tools/docker/test-image/Dockerfile
aidanhs: Not necessary, it was a compromise given that replaced `ubuntu:latest` with `ubuntu:20.04`

tools/docker/test-image/Dockerfile
aidanhs: My personal practice is that what you describe is appropriate when creating docker images that then get pushed around/used in 'production' - but this seems to be an image used for development purposes, and so the ease of e.g. 
 - tweaking lines without rerunning all of the (potentially long) previous steps
 - hacking around with new packages at the end without `apt-get update`
(both of which I was personally doing) outweighed the desire for an optimally compact image.

tools/docker/test-image/Dockerfile
aidanhs: Willdo

copy:
Cherry-picked https://github.com/copy/v86/commit/34339fe8d8bb1b502b0ba3218fc75db01e68dfb3 and added some extra fixes in https://github.com/copy/v86/commit/4f6acde82186873873c95f33270236b36a6dff2e. I believe the switch to Ubuntu 21.04 fixed the issue in util.js. Cheers!

Issue #452, opened 2021-04-21
Provide API for memory access


skadisch:
When using the V86Starter interface, there seems to be no way to access memory of the machine. Internally one could use `this.v86.cpu.read_blob()` and `this.v86.cpu.write_blob()` to access the memory. Would it make sense to forward and provide these functions to the outside world? Or is there already another way to do this?

copy:
I think it would be reasonable to expose those functions in the public interface.

PR #454, opened 2021-04-21
Provide access to CPU memory via V86Starter

skadisch:
This implements #452 by publicly exposing existing internal API functionality.


copy:
Cheers!

PR #464, opened 2021-05-04
avoid ending jit_run_interpreted in the middle of a basic block

viorelcanja:
Right now jit_run_interpreted may exit on any instruction causing a new entry point to be created at the exit location. This patch avoids that by waiting for a backward jump to exit the function. This in turn avoids unnecessary entry points and unnecessary BB splitting.


copy:
Cheers!

PR #469, opened 2021-05-09
Mem fast path optimization

viorelcanja:
Removes one addition from the memory access fast path by including the offset to the memory::mem8 array directly in the page entry data stored in tlb_data. This improves performance.

### DIFF
--- a/src/rust/codegen.rs
+++ b/src/rust/codegen.rs
@@ -695,8 +695,6 @@ pub fn gen_get_phys_eip_plus_mem(ctx: &mut JitContext, address_local: &WasmLocal
     // XXX: Currently does not use ctx.start_of_current_instruction, but rather assumes that eip is
     //      already correct (pointing at the current instruction)
 
-    // test, force pipeline rerun
-
     let cont = ctx.builder.block_void();
     ctx.builder.get_local(&address_local);
 
--- a/src/rust/codegen.rs
+++ b/src/rust/codegen.rs
@@ -620,7 +620,8 @@ fn gen_safe_read(
         },
     }
     ctx.builder.tee_local(&entry_local);
-    ctx.builder.eqz_i32();
+    ctx.builder.const_i32(1);
+    ctx.builder.and_i32();
 
     if cfg!(feature = "profiler") {
         ctx.builder.if_void();
@@ -628,7 +629,8 @@ fn gen_safe_read(
         ctx.builder.block_end();
 
         ctx.builder.get_local(&entry_local);
-        ctx.builder.eqz_i32();
+        ctx.builder.const_i32(1);
+        ctx.builder.and_i32();
     }
 
     ctx.builder.br_if(ctx.exit_with_fault_label);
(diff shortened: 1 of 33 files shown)


viorelcanja:
hi! @copy , how can I download the image for this test ? --> Test failed: Tiny Core 11 CD. 
Also, on my machine some tests are segfaulting (for instance: linux reboot). Any suggestions for a proper setup for debugging the tests ?

viorelcanja:
@copy, the tiny core test failed because the test agent took a bit longer to execute it (in previous runs it was at 288s and the limit is at 300s). Other then that I fixed the expect tests. Any feedback from your side ?

proxy-m:
What is the percent of speed optimisation?

viorelcanja:
Very small. I estimate from benchmarks between 0.5 to 1% on general code. I need to estimate because at this level the noise in the benchmarks is relevant. Anyway, I did a couple of runs with the patch and selected the best one. After that I did run the old code about 10 times and it always had a bit lower score but with a small margin. 

The patch does remove an addition but that is probably just one register operation on the host and it is only for the memory fast path which is probably more than 10 - 20 instructions  in native code. 

Overall ... I am not sure if the more convoluted code (baking in an offset in some values) is worth the gain ...

copy:
> the tiny core test failed because the test agent took a bit longer to execute it (in previous runs it was at 288s and the limit is at 300s).

No worries. CI speed isn't very stable so I wouldn't put that on this PR. I'll bump the limit on that test.

> Overall ... I am not sure if the more convoluted code (baking in an offset in some values) is worth the gain ...

I think this is a worthwhile optimisation. Memory reads or writes often make up 50% of the executed instructions (and even more of the code size), and this is an obvious win.

src/rust/codegen.rs
```diff
@@ -723,6 +729,14 @@ pub fn gen_get_phys_eip(ctx: &mut JitContext, address_local: &WasmLocal) {
...
     ctx.builder.get_local(&address_local);
     ctx.builder.call_fn1_ret("get_phys_eip_slow_jit");
+
+    ctx.builder.get_local(&address_local);
+    ctx.builder.xor_i32();
+    ctx.builder.const_i32(unsafe { memory::mem8 } as i32);
+    ctx.builder.add_i32();
+    ctx.builder.get_local(&address_local);
+    ctx.builder.xor_i32();
+
```
copy, line 739: I think this could be moved into `get_phys_eip_slow_jit` (keeps the generated code size small).

copy (commented):
Code looks good, one minor comment, plus two things that would be nice-to-have:

- `get_phys_eip_slow_jit`, `safe_read_slow_jit` and `safe_write_slow_jit` could be changed to return 0 on page fault. 0 is a more obvious error code and slightly reduces code size by allowing to use `eqz`
- Should run rustfmt (`cargo fmt --all`)

src/rust/codegen.rs
viorelcanja: moved

copy:
@viorelcanja Thanks for the contribution.

FYI, I had to revert the eqz optimisation I suggested earlier — my bad. I thought we were returning `mem8 + phys`, which can't be zero, but we're actually returning `(mem8 + phys) ^ virt`. There's another optimisation I'm planning to implement that will further reduce the number of operations of memory accesses (at the cost of increasing the tlb size), which will bring back the eqz part. But this PR is good progress, squashed and merged.

Issue #460, opened 2021-04-29
DOS text-mode menu characters rendered as "undefined"


parkertomatoes:
Sometimes in DOS menu characters are rendered as "undefined":
<img width="1112" alt="Screen Shot 2021-04-29 at 7 05 00 AM" src="https://user-images.githubusercontent.com/6298931/116542641-df612080-a8ba-11eb-85ca-385a63890c8b.png">

I was able to reproduce this using Safari, Firefox, and Chrome on MacOS (M1), as well as Firefox and Chrome on Windows. I'm not sure what triggers it, but it wasn't always there - I remember it rendering correctly when I used v86 a year ago.

Steps to reproduce:
- Start the DOS 6.22 demo: https://copy.sh/v86/?profile=msdos
- Run the command "edit" to start the MS-DOS editor
- Press enter to close the welcome dialog
- Menu is now "undefined" repeated.

parkertomatoes:
I took a stab at debugging this, thinking it was a simple Javascript issue due to all the "undefined"'s.

What's happening, is: 
 1. `mmap_write8` gets called with an address in the VGA text buffer range (B8000h+) and a very large value (260,000+). 
 2. That invokes [`vga_memory_write`](https://github.com/copy/v86/blob/3791d63ffd09881782690902c6ee57f0edbd56a3/src/vga.js#L654), 
 3. ...which invokes [`vga_memory_write_text_mode`](https://github.com/copy/v86/blob/3791d63ffd09881782690902c6ee57f0edbd56a3/src/vga.js#L885). 
 4. ...which triggers an event, ["screen-put-char"](https://github.com/copy/v86/blob/3791d63ffd09881782690902c6ee57f0edbd56a3/src/browser/screen.js#L198) on the bus
 5. ...which places the value in a UInt32Array called `text_mode_data`
 6. The value [gets passed to a CP437 lookup table](https://github.com/copy/v86/blob/3791d63ffd09881782690902c6ee57f0edbd56a3/src/browser/screen.js#L478), which returns `undefined` because the table is only 256 entries and the value is very large

Masking the value at any point in that data flow is _seems like_ enough to superficially fix the issue, but why does mmap_write8 get called with a large value in the first place? I don't really know how to debug into the rust code to tell.

copy:
[...]

PR #472, opened 2021-05-13
Mask values written to mmap_write{8,16}

parkertomatoes:
Fix for #460

### DIFF
--- a/src/memory.js
+++ b/src/memory.js
@@ -8,7 +8,7 @@ CPU.prototype.mmap_read8 = function(addr)
 
 CPU.prototype.mmap_write8 = function(addr, value)
 {
-    dbg_assert(value <= 0xFF);
+    dbg_assert(value >= 0 && value <= 0xFF);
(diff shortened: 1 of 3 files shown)


src/memory.js
```diff
@@ -8,6 +8,7 @@ CPU.prototype.mmap_read8 = function(addr)
 
 CPU.prototype.mmap_write8 = function(addr, value)
 {
+    dbg_assert(value <= 0xFF);
```
copy, line 11: Should also check `value >= 0`

src/memory.js
```diff
@@ -22,6 +23,7 @@ CPU.prototype.mmap_write16 = function(addr, value)
 {
     var fn = this.memory_map_write8[addr >>> MMAP_BLOCK_BITS];
 
+    dbg_assert(value <= 0xFFFF);
```
copy, line 26: Same as above

copy:
I think tests are failing because `write8` and `write16` in memory.rs also need a mask.

copy:
Cheers!

PR #477, opened 2021-05-26
Use `queueMicrotask` for `fast_next_tick` in Web Worker environment.

losfair:
If v86 is running in a Web Worker environment, use `queueMicrotask` for `fast_next_tick`.

With a worker thread dedicated to v86 we can run ticks without worrying about blocking the UI, but `setTimeout()` is still called every 256 ticks for a chance to run the `onmessage` callback.

Previously, `fast_next_tick` falls back to `setTimeout` and is slow.

CoreMark results under different configurations:

Main thread:

```
~ # coremark
2K performance run parameters for coremark.
CoreMark Size    : 666
Total ticks      : 11288
Total time (secs): 11.288000
Iterations/Sec   : 1771.793055
Iterations       : 20000
Compiler version : GCC8.3.0
Compiler flags   : -O2   -lrt
Memory location  : Please put data memory location here
                        (e.g. code in flash, data on heap etc)
seedcrc          : 0xe9f5
[0]crclist       : 0xe714
[0]crcmatrix     : 0x1fd7
[0]crcstate      : 0x8e3a
[0]crcfinal      : 0x382f
Correct operation validated. See README.md for run and reporting rules.
CoreMark 1.0 : 1771.793055 / GCC8.3.0 -O2   -lrt / Heap
```

Worker (`setTimeout`):

```
~ # coremark
2K performance run parameters for coremark.
CoreMark Size    : 666
Total ticks      : 13898
Total time (secs): 13.898000
Iterations/Sec   : 79.148079
Iterations       : 1100
Compiler version : GCC8.3.0
Compiler flags   : -O2   -lrt
Memory location  : Please put data memory location here
                        (e.g. code in flash, data on heap etc)
seedcrc          : 0xe9f5
[0]crclist       : 0xe714
[0]crcmatrix     : 0x1fd7
[0]crcstate      : 0x8e3a
[0]crcfinal      : 0x33ff
Correct operation validated. See README.md for run and reporting rules.
CoreMark 1.0 : 79.148079 / GCC8.3.0 -O2   -lrt / Heap
```

Worker (`queueMicrotask`):

```
~ # coremark
2K performance run parameters for coremark.
CoreMark Size    : 666
Total ticks      : 16578
Total time (secs): 16.578000
Iterations/Sec   : 1809.627217
Iterations       : 30000
Compiler version : GCC8.3.0
Compiler flags   : -O2   -lrt
Memory location  : Please put data memory location here
                        (e.g. code in flash, data on heap etc)
seedcrc          : 0xe9f5
[0]crclist       : 0xe714
[0]crcmatrix     : 0x1fd7
[0]crcstate      : 0x8e3a
[0]crcfinal      : 0x5275
Correct operation validated. See README.md for run and reporting rules.
CoreMark 1.0 : 1809.627217 / GCC8.3.0 -O2   -lrt / Heap
```

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -82,12 +82,20 @@ v86.prototype.init = function(settings)
     this.bus.send("emulator-ready");
 };
 
-if(typeof setImmediate !== "undefined")
-{
+
+if(typeof importScripts === 'function' && typeof queueMicrotask === 'function') {
+    var tickCounter = 0;
+
     /** @this {v86} */
     var fast_next_tick = function()
     {
-        setImmediate(() => { this.do_tick(); });
+        if(tickCounter == 256) {
+            tickCounter = 0;
+            setTimeout(() => { this.do_tick(); }, 0);
+        } else {
+            tickCounter += 1;
+            queueMicrotask(() => { this.do_tick(); });
+        }
     };
 
     /** @this {v86} */
@@ -95,6 +103,20 @@ if(typeof setImmediate !== "undefined")
 
     /** @this {v86} */
     var unregister_tick = function() {};
+}
+else if(typeof setImmediate !== "undefined")
+{
+    /** @this {v86} */
+    fast_next_tick = function()
+    {
+        setImmediate(() => { this.do_tick(); });
+    };
+
+    /** @this {v86} */
+    register_tick = function() {};
+
+    /** @this {v86} */
+    unregister_tick = function() {};
 }
 else if(typeof window !== "undefined" && typeof postMessage !== "undefined")
 {
--- a/src/main.js
+++ b/src/main.js
@@ -83,17 +83,21 @@ v86.prototype.init = function(settings)
 };
 
 
-if(typeof importScripts === 'function' && typeof queueMicrotask === 'function') {
-    var tickCounter = 0;
+if(typeof importScripts === "function" && typeof queueMicrotask === "function")
+{
+    let tick_counter = 0;
 
     /** @this {v86} */
     var fast_next_tick = function()
     {
-        if(tickCounter == 256) {
-            tickCounter = 0;
+        if(tick_counter === 256)
+        {
+            tick_counter = 0;
             setTimeout(() => { this.do_tick(); }, 0);
-        } else {
-            tickCounter += 1;
+        }
+        else
+        {
+            tick_counter++;
             queueMicrotask(() => { this.do_tick(); });
         }
     };


copy:
Interesting, it's the first time I'm hearing about `queueMicrotask`. The benchmark results look very nice.

> With a worker thread dedicated to v86 we can run ticks without worrying about blocking the UI, but setTimeout() is still called every 256 ticks for a chance to run the onmessage callback.

I'm curious about this bit. If the `onmessage` callback isn't received, then *what* happens between the call to `queueMicrotask` and its callback being called? Or to be more direct, what is the advantage of using `queueMicrotask` over calling `do_tick` 256 times?

losfair:
@copy The browser maintains both a *task* queue and a *microtask* queue ([WHATWG spec](https://html.spec.whatwg.org/multipage/webappapis.html#task-queue)). The task queue is responsible for all "macrotasks" (e.g. setTimeout callbacks, and `onmessage`), while the microtask queue is responsible for microtasks (including `Promise` callbacks and `queueMicrotask`s.). All microtasks in the queue are dequeued and executed after the current macrotask has finished, but before the next macrotask runs, until there are no more microtasks remaining in the queue.

So:

> what happens between the call to queueMicrotask and its callback being called?

The callback is deferred to *immediately* after the current macrotask finishes execution. Like a `setImmediate` (which is not available in browsers), but always runs earlier, even before `onmessage`.

> what is the advantage of using queueMicrotask over calling do_tick 256 times?

I'm just trying to preserve the original "deferred execution" semantics :)

urish:
Thanks for this pull request! I just pulled the changes into [web-gdb](https://blog.wokwi.com/running-gdb-in-the-browser/), and you can see the huge difference in the performance profiler's flame chart —

Before:

image

Now:

image

As you can see, the CPU utilization is much better now, approaching 100%. Also, I can tell the responsiveness of web-gdb has improved: it's now ~3 times faster. So thank you!

copy:
@losfair  Interesting, thanks for the explanation! Somehow I was under the impression that web workers were exempt from the 4ms timeout rule, but as @urish has shown, that doesn't seem to be the case. 

There's probably some alternatives that we should consider:

- Bumping [`TIME_PER_FRAME`](https://github.com/copy/v86/blob/4e5f62bbc14941efa76f2218d71714903c12c290/src/config.js#L55) when running in workers. It's currently set to 1ms, which is where the 4x slowdown is coming from (1ms execution time, 4ms sleeping). I'm afraid the current implementation may takes as long as 256ms before reading messages under load, so it may cause a significant input latency, but I'll test that.
- Finding a different implementation that properly yields to the browser without delay, like the current `postMessage` version. Something like [this](https://github.com/YuzuJS/setImmediate/blob/f1ccbfd/setImmediate.js#L124-L134) might be worth a try.

In any case, I'm inclined to merge this as-is, since it is a significant improvement for web workers, and we can still improve it later.

urish:
> Finding a different implementation that properly yields to the browser without delay, like the current postMessage version. Something like this might be worth a try.

That's [also what I use](https://github.com/wokwi/avr8js/blob/5f711e1a016aae0f59144834eb191e1aff4abce2/demo/src/task-scheduler.ts#L4) in AVR8js and for [wokwi.com](wokwi.com) - the advantage is it works well both inside a worker and on the main thread, and while still allowing good input latency.

> In any case, I'm inclined to merge this as-is, since it is a significant improvement for web workers, and we can still improve it later.

👍

copy:
@losfair Thanks!

PR #483, opened 2021-06-04
Using and splitting images without server-side

Pixelsuft:
Slower than default part file buffer, but can run without server-side, for example - on github pages, where file size limit is 100mb

## How to use
Example on Windows 3.1 image:
```sh
python3 tools/imgsplit.py -f win31.img -s 4096
```
src/browser/main.js
```js
{
	id: "windows31",
	memory_size: 64 * 1024 * 1024,
	hda: {
		"url": "/win31_out/win31.img",
		"async": true,
		"size": 34463744,
		"step": 4096, // just add this to use without server-side
		use_parts: true
	},
	name: "Windows 3.1",
}
```
When {step} is smaller, it will be faster, but have more part files.
I don't recommending to use {step} smaller than 512

### DIFF
--- a/src/browser/lib.js
+++ b/src/browser/lib.js
@@ -460,7 +460,7 @@ var ASYNC_SAFE = false;
      * @param {string} filename Name of the file to download
      * @param {number|undefined} size
      */
-    function AsyncXHRPartfileBuffer(filename, size)
+    function AsyncXHRPartfileBuffer(filename, size, step)
     {
         const parts = filename.match(/(.*)(\..*)/);
 
@@ -478,6 +478,8 @@ var ASYNC_SAFE = false;
         /** @const */
         this.block_size = 256;
         this.byteLength = size;
+        this.use_step_ = typeof step == 'number'
+        this.step_ = step
 
         this.loaded_blocks = Object.create(null);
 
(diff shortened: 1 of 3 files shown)


Pixelsuft:
For example: https://pixelsuft.github.io/windows95/?profile=windows95
(Original windows95.img is 1GB size)

tools/imgsplit.py
```diff
@@ -0,0 +1,74 @@
...
+    temp_file.write(readf[i:i + step])
+    temp_file.close()
+    i += step
+
+
+temp_file = open(os.path.join(out_dir, f'{no_ext}-{i}.{ext}'), 'wb')
+temp_file.write(readf[i:file_length])
+temp_file.close()
+
+opened.close()
```
copy, line 74: Perhaps this could be replaced with a shell script that uses [gnu split](https://www.gnu.org/software/coreutils/manual/html_node/split-invocation.html).

src/browser/lib.js
```diff
@@ -522,18 +524,44 @@ var ASYNC_SAFE = false;
...
+			v86util.load_file(part_filename, {
+				done: function done(buffer)
+				{
+					dbg_assert(buffer.byteLength === len);
+					var block = new Uint8Array(buffer);
+					this.handle_read(offset, len, block);
+					fn(block);
+				}.bind(this),
+			});
+		}
```
copy, line 564: Please follow the style of the rest of the code:

- No space after `if`
- `{` on its own line
- Indent with 4 spaces
- double quotes
- semicolon at end of line
- `===` over `==`
- No `_` at the end of properties

hello-smile6:
What about doing the ISO split in the browser using Arch?

Pixelsuft:
@hello-smile6 
How I know, in v86 you can export only file (, not dir).
But as idea all files can be packed into zip archive

hello-smile6:
@Pixelsuft 
> @hello-smile6
> How I know, in v86 you can export only file (, not dir).
> But as idea all files can be packed into zip archive

You can use https://github.com/humphd/browser-shell 's code to do that.

Pixelsuft:
I've fixed problems with `total_count` and `slice`

PR #501, opened 2021-07-04
(feat) Docker: Add Support for Running inside Docker Container

Progyan1997:
Uses Alpine image as base to have minimal size. 2-step build process to get rid of source and library files after compilation and during runtime. Uses Python3 HTTP Server to serve static assets.

Signed-off-by: Progyan Bhattacharya <bprogyan@gmail.com>

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -5,7 +5,7 @@ WebAssembly modules at runtime in order to achieve decent performance. Here's a
 list of emulated hardware:
 
 - An x86-compatible CPU. The instruction set is around Pentium III level,
- including full SSE2 support. Some features are missing, in particular:
+  including full SSE2 support. Some features are missing, in particular:
   - Task gates, far calls in protected mode
   - Some 16 bit protected mode features
   - Single stepping (trap flag, debug registers)
@@ -30,8 +30,7 @@ list of emulated hardware:
 - A virtio filesystem.
 - A SoundBlaster 16 sound card.
 
-Demos
--
+## Demos
 
 [Arch Linux](https://copy.sh/v86/?profile=archlinux) —
 [Damn Small Linux](https://copy.sh/v86/?profile=dsl) —
@@ -51,9 +50,7 @@ Demos
 [KolibriOS](https://copy.sh/v86/?profile=kolibrios) —
 [QNX](https://copy.sh/v86/?profile=qnx)
 
-
-Compatibility
--
+## Compatibility
 
 Here's an overview of the operating systems supported in v86:
 
@@ -87,8 +84,7 @@ Here's an overview of the operating systems supported in v86:
 
 You can get some infos on the disk images here: https://github.com/copy/images.
 
-How to build, run and embed?
--
+## How to build, run and embed?
 
 You need:
 
@@ -102,7 +98,6 @@ You need:
 
 See `tools/docker/test-image/Dockerfile` for a full setup on Debian.
 
-
 - Run `make` to build the debug build (at `debug.html`).
 - Run `make all` to build the optimized build (at `index.html`).
 - ROM and disk images are loaded via XHR, so if you want to try out `index.html`
@@ -111,9 +106,15 @@ See `tools/docker/test-image/Dockerfile` for a full setup on Debian.
 - If you only want to embed v86 in a webpage you can use libv86.js. For
   usage, check out the [examples](examples/).
 
+### Alternatively, to build using docker
 
-Testing
--
+- If you have docker installed, you can run the whole system inside a container.
+- See `tools/docker/exec` to find Dockerfile required for this.
+- You can run `docker build -f tools/docker/exec/Dockerfile -t v86:alpine-3.14 .` from the root directory to generate docker image.
+- Then you can simply run `docker run -it v86:alpine-3.14 -p 8000:800` to start the server.
+- Check `localhost:8000` for hosted server.
+
+## Testing
 
 The disk images for testing are not included in this repository. You can
 download them directly from the website using:
@@ -124,9 +125,7 @@ Run all tests: `make jshint rustfmt kvm-unit-test nasmtests nasmtests-force-jit 
 
 See [tests/Readme.md](tests/Readme.md) for more infos.
 
-
-API examples
--
+## API examples
 
 - [Basic](examples/basic.html)
 - [Programatically using the serial terminal](examples/serial.html)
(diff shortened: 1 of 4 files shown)


.dockerignore
```diff
@@ -0,0 +1,11 @@
...
+.github
+docker
+docs
+examples
+tests
+.editorconfig
+.gitattributes
+.gitignore
+.jshint.json
+Readme.md
```
copy, line 11: Instead of having this file, can you specify a list of files to copy in the `COPY` statement below instead? I have some big directories in my v86 directory that wouldn't be properly excluded by this anyway.

docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+WORKDIR /v86
+
+# Install runtime dependencies for building artifacts
+RUN apk add --update curl clang make openjdk8 npm python3
+
+# Install Rustup from Official Release using Install Shell Script
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+
+# Update `PATH` variable to point to Rust Binaries (Does not work automatically) and install required Crates for the target platform
+RUN PATH="${HOME}/.cargo/bin:${PATH}" rustup target add wasm32-unknown-unknown
```
copy, line 12: See the style of the [other Dockerfile](https://github.com/copy/v86/blob/master/tools/docker/test-image/Dockerfile). Adjacent `RUN` lines should be combined to avoid unnecessary intermediate images. Comments are only necessary for non-obvious lines (I'd say no comments are necessary for this Dockerfile).

docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# Copy built artifacts from intermediate container
+COPY --from=v86-builder v86 .
+
+# Start Pythom HTTP Server to serve static files
+ARG PORT=8000
+CMD python3 -m http.server ${PORT}
+
+# Allow external modules and clients to connect
+EXPOSE ${PORT}
```
copy, line 32: Should be moved into `tools/docker/run` or so. Minor suggestion, would you mind adding a readme.md file to `tools/docker` that describes each subdirectory, i.e. `debian` is used to create an image to run inside of v86, `test-image` can be used to run v86's tests and this one can be used to run `v86` on port 8000.

copy (changes requested):
Thanks! This looks very useful and I'd like to merge it. Some suggestions below.

.dockerignore
Progyan1997: Can you help me out in this? @copy What should we copy for build. I believe I have clear set of files for runtime already.

Progyan1997: Btw the image size so formed is of 49MB

copy: Tor build: .cargo gen lib src tools cargo.toml Makefile
For runtime: index.html, bios and v86.css. Optionally debug.html, loader.js and src for debug mode. Optionally examples.

Progyan1997: Updated. Image so formed is now 48.9 Mb.

copy: Sorry, perhaps I wasn't clear. My idea is to remove the `.dockerignore` file and change the `COPY . . ` line to an explicit list of files (the ones I listed above should work, but let me know if it doesn't). The reason for that is that I have several directories in my v86 directory containing image files, which would significantly increase the size of the docker image.

copy (changes requested):

.dockerignore
Progyan1997: I get that you want to get rid of `.dockerignore` file, but I am just not able to understand why. If any directory of your's have fat static files, why can't we use docker's own process to get rid of that. As a side note, `COPY` does not work well with directories, so I would probably have to write multiple `COPY` to get them as is to the workspace tree.
Also the final image does not have any of those build time dependencies. Was that also your concern? cc:/ @copy

ndabas: Just a quick point of clarification for @copy: when you run `docker build`, the entire contents of the current directory (the _build context_) are sent to the Docker daemon, unless you have a `.dockerignore` file to tell the command what _not_ to send to the daemon. If you  have lots of large files in the current directory, those will slow down the build and eat up lots of memory as well during the build.

- Those files are not automatically added to the built Docker image. You have to explicitly copy files from the build context into the image. That's what `COPY` does in a Dockerfile.
- If you want to be explicit about which files are sent, an option is to create a tarball of only those files and specify that as the build context for `docker build`.

copy: Thanks for the clarification @ndabas. I had a misunderstanding about how docker works. I'd say a tarball is the way to go. Does that sound reasonable to you @mprey?

Progyan1997: @ndabas Thanks for adding this clarification. Just a quick question, have you run `docker build` with this file and if so, what was the experience?

ndabas: @Progyan1997 happy to have added to the discussion. I haven't actually tried out `docker build` with the new Dockerfile though.

Progyan1997: @ndabas It would be really appreciated if you can give some time to try it out and share pain points if any here.
@copy Sorry for the trouble again. I don't really see the point of creating a tarball as additional step before running the docker build and most of the files you're concerned about would get dropped from the context (Docker daemon's own generated tarball) just because of `.dockerignore`. And if you run this `Dockerfile` using `docker build`, I believe you will get a much more clear insight over build time and UFS size.

Progyan1997: On a different note, after running `docker build` multiple times, I experienced a certain delay during compilation process and I believe there is certain room for improvement in `Makefile` to make it more efficient.

Readme.md
```diff
@@ -115,6 +106,13 @@ See `tools/docker/test-image/Dockerfile` for a full setup on Debian.
 - If you only want to embed v86 in a webpage you can use libv86.js. For
   usage, check out the [examples](examples/).
 
+### Alternatively, to build using docker
+
+- If you have docker installed, you can run the whole system inside a container.
+- See `tools/docker/exec` to find Dockerfile required for this.
+- You can run `docker build -f tools/docker/exec/Dockerfile -t v86:alpine-3.14 .` from the root directory to generate docker image.
+- Then you can simply run `docker run -it v86:alpine-3.14` to start the server.
```
copy, line 114: Does this need a `-p 8000:8000`?

.dockerignore
copy: `.dockerignore` is not sufficient, for example if there is a 64GB img file in `test-images/arch.img`. I've added a build script to this directory that uses tar to send only the required files. Let me know if this looks good to you and it should be good to squash and merge.

copy (approved):

.dockerignore
Progyan1997: Looks good. @copy

copy:
@Progyan1997 Thanks for the contribution!

Progyan1997:
My pleasure, @copy and thanks for building this.

PR #526, opened 2021-09-20
Fix debian build

nwtgck:
Hi,

I fixed `tools/docker/debian` to build debian image. `FROM i386/debian` causes errors, "... bullseye-security InRelease' is not signed.", so I specified `:buster`. I added `../` to `build-container.sh` and `build-state.js`.

One thing I'm not sure is the following statement in tools/docker/debian/Readme.md.

> Go to debug.html?profile=debian to start the generated container.

I don't know how to start the built debian. I confirmed I could run the built debian, modifying examples/arch.html instead.

image

But anyway, this PR fixed building of debian.

### DIFF
--- a/tools/docker/debian/Dockerfile
+++ b/tools/docker/debian/Dockerfile
@@ -1,4 +1,4 @@
-FROM i386/debian
+FROM i386/debian:buster
 
 ENV DEBIAN_FRONTEND noninteractive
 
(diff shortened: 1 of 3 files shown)


copy:
Thanks!

> One thing I'm not sure is the following statement in tools/docker/debian/Readme.md.

At one point in the past it was planned to release a debian profile, but it was scrapped and instead the Arch profiled was updated to Archlinux 32.

nwtgck:
Thank you for the information!

PR #540, opened 2021-10-17
Supports disabling A20 through PS/2

cshung:
This change added the logic to support disabling A20 through PS/2.

This is to fix Xv6 keyboard input. Without this fix, the OS cannot accept any keyboard inputs. With this fix, Xv6 can accept keyboard inputs.

The problem was that the emulated hardware didn't raise the keyboard irq when a keyboard event happens.

PS/2 is not raising the keyboard irq because `next_byte_is_ready` is `true`, and that is because a keyboard irq was raised but not handled. The keyboard irq was raised because when the Xv6 bootloader tried to disable a20, the PS/2 controller fall-through the switch cases for keyboard command and raised a keyboard irq.

The associated Xv6 code is here.
https://github.com/mit-pdos/Xv6-public/blob/eeb7b415dbcb12cc362d0783e41c3d1f44066b17/bootasm.S#L37

For more info about disabling A20, see here:
https://wiki.osdev.org/A20_Line

For more info about the PS/2 control output port, see
https://wiki.osdev.org/%228042%22_PS/2_Controller

This change is probably not correct, `this.cpu.disable_a20` is no longer available, and we might want to implement the control output port logic more closely to what the hardware actually do than to just tie it to the disable a20 case (althrough that seems sufficient)

Sharing a blog entry discussing this issue in more depth: https://cshung.github.io/posts/ps2-and-a20/

FYI for @onriv who is also interested in getting Xv6 running.

### DIFF
--- a/src/ps2.js
+++ b/src/ps2.js
@@ -111,9 +111,11 @@ function PS2(cpu, bus)
     }, this);
 
     this.command_register = 1 | 4;
+    // TODO: What should be the initial value?
+    this.controller_output_port = 0;
     this.read_output_register = false;
     this.read_command_register = false;
-    this.disable_a20 = 0;
+    this.read_controller_output_port = false;
 
     cpu.io.register_read(0x60, this, this.port60_read);
     cpu.io.register_read(0x64, this, this.port64_read);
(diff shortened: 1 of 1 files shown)


copy:
> This change is probably not correct, this.cpu.disable_a20 is no longer available, and we might want to implement the control output port logic more closely to what the hardware actually do than to just tie it to the disable a20 case (althrough that seems sufficient)

It would nice to write and read back the register and add some logging, but I don't think precise emulation is necessary. It hasn't come up in any other OS so far.

Issue #562, opened 2021-11-28
Linux-only ls option breaks build on MacOS


parkertomatoes:
This commit makes a minor tweak to the Makefile to show files sizes in KB using the `--block-size` option in ls
https://github.com/copy/v86/commit/ffae108c4297cd904c1ae2bd3d8b34d69bf49837

Unfortunately, the `--block-size` option causes this error to occur on MacOS:
```
ls -l --block-size=K build/v86-debug.wasm
ls: illegal option -- -
usage: ls [-@ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1%] [file ...]
make: *** [build/v86-debug.wasm] Error 1
```

A work-around might be to use the environment variable version of this option:
```
BLOCK_SIZE=K ls -l build/v86-debug.wasm
```

The output would be the same on Linux, and while it wouldn't show the file sizes in kilobytes on MacOS, the build would pass.

copy:
Cheers!

PR #563, opened 2021-11-28
fix build on MacOS

parkertomatoes:
This attempts to work around #562 by moving the block-size option in ls to an environment variable


copy:
Thanks!

Issue #448, opened 2021-04-15
Config file for v86


raspiduino:
It would be nice to add a config file feature to v86, so people can load config files to the emulator instead of re-configure them every time. Thanks!

copy:
I think extending and documenting the url interface would be a good step in this direction, i.e. making something like this work:

https://copy.sh/v86/?cdrom=https://k.copy.sh/linux4.iso&m=512M&networking=0&acpi=1

Or perhaps a json file inspired by the internal representation of profiles, e.g.: https://github.com/copy/v86/blob/690354e69ef8473535c308de23868df9479f003f/src/browser/main.js#L193-L204

broten15:
Hi, so I see that the URL config feature is merged, but are you still looking to add a feature for uploading a config file that automatically fills out the set up fields? If so, I could work on that.

copy:
@broten15 Sure!

copy:
This is now done via url parameters: Input fields propagate to url parameters, which in turn pre-fill the inputs if you open a page without a profile.

Feel free to create a new issue if anything is missing.

PR #566, opened 2021-12-08
Added url config

Shizuka0000:
Added url config option for networking, audio, acpi.
Custom profile can now also be configured using the url.
#448

### DIFF
--- a/src/browser/main.js
+++ b/src/browser/main.js
@@ -898,17 +898,17 @@
                     settings.vga_memory_size = vram * 1024 * 1024;
                 }
 
-                const networking = parseInt(query_args["networking"]);
+                const networking = parseInt(query_args["networking"], 10);
                 //leave blank to disable
                 if (networking == 0)
                 {
                     settings.networking = ""
                 }
 
-                const audio = !!parseInt(query_args["audio"]);
+                const audio = !!parseInt(query_args["audio"], 10);
                 settings.audio = audio;
 
-                const acpi = !!parseInt(query_args["acpi"]);
+                const acpi = !!parseInt(query_args["acpi"], 10);
                 settings.acpi = acpi;
             }
 
--- a/src/browser/main.js
+++ b/src/browser/main.js
@@ -842,40 +842,34 @@
             {
                 $("boot_options").style.display = "none";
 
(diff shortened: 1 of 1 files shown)


src/browser/main.js
```diff
@@ -842,10 +842,36 @@
             {
                 $("boot_options").style.display = "none";
 
+                parse_param()
                 start_emulation(settings, done);
             }
         }
 
+        function parse_param()
```
copy, line 850: Could you pass settings to this function and move it to the outer scope (above `onload`)?

src/browser/main.js
```diff
@@ -842,10 +842,36 @@
             {
                 $("boot_options").style.display = "none";
 
+                parse_param()
```
copy, line 845: Missing a trailing semicolon.

src/browser/main.js
```diff
@@ -842,10 +842,36 @@
...
+            const vram = parseInt(query_args["vram"], 10);
+            if(vram > 0)
+            {
+                settings.vga_memory_size = vram * 1024 * 1024;
+            }
+
+            const networking = parseInt(query_args["networking"], 10);
+            settings.networking = networking == 0 ? undefined : networking;
+
+            const audio = !!parseInt(query_args["audio"], 10);
```
copy, line 867: I believe this should be called `disable_audio`. Alternatively, if you want to keep the name `audio` (which is a bit nicer), it should default to true.

src/browser/main.js
```diff
@@ -842,10 +842,36 @@
...
+                settings.memory_size = Math.max(16, m) * 1024 * 1024;
+            }
+
+            const vram = parseInt(query_args["vram"], 10);
+            if(vram > 0)
+            {
+                settings.vga_memory_size = vram * 1024 * 1024;
+            }
+
+            const networking = parseInt(query_args["networking"], 10);
```
copy, line 864: Should probably be called `networking_proxy` to make it a bit clearer.

Shizuka0000:
Thanks for your reviews! I addressed the issues that you mentioned, please comment if there are any additional issues.

copy:
Thanks!

Issue #299, opened 2019-10-13
Boot SerenityOS


copy:
Pretty cool hobby OS: https://github.com/SerenityOS/serenity

Requires hardware task switching support: https://github.com/SerenityOS/serenity/blob/7fc903b/Kernel/Scheduler.cpp#L574-L578

asiawin:
Yeah! I dont think you can boot SerenityOS because it's way too large!

asiawin:
Of course! You can boot Cadex OS! Go to https://github.com/opencreeck/Cadex-OS-Official, go to the releases section and download any version of Cadex OS. It's just 4 MB and it can support up to 3 GB of RAM because it's a 32-Bit os and it has some fun things! I would like it if I could run SerenityOS because it is a very good OS!

copy:
Serenity seems to have removed (most uses of) hardware tasking in https://github.com/SerenityOS/serenity/commit/fb41d89384cb5bbdf602ae41ae7e038fb48a34ce. Next blocker: PAE (https://github.com/copy/v86/issues/97)

asiawin:
Oh, I didn't see that! I am now trying to build SerenityOS and it says CMake error. Asked Andreas but he said search google for solution

copy:
This is now working (except for newer versions, which are likely affected by #597): https://copy.sh/v86/?profile=serenity

copy:
(cc @pwmarcz, thanks for the info in https://github.com/copy/v86/pull/599#issuecomment-1013635444)

I fixed an ACPI bug in v86 (it was not being applied). Now running into https://github.com/SerenityOS/serenity/issues/11980 (even with ACPI on in v86). Will wait on a fix on their side and then look into the IDE and graphics issues.

supercomputer7:
https://github.com/SerenityOS/serenity/issues/11980 is closed as it was resolved.
I'm working on support for the QEMU ISA-PC machine type for SerenityOS:
https://github.com/SerenityOS/serenity/pull/12046
I don't know in what state this project is in right now, just thought it might help you somehow :)

copy:
After the ACPI bugfixes we're running into "ASSERTION FAILED: !m_storage_devices.is_empty()". It looks like a bug in v86's pci code, which I'm investigating. A Serenity build with legacy ISA port access would probably work too.

copy:
Seems to be three separate issues (with `enable_ioapic=off acpi=off`):

[...]

Issue #97, opened 2016-05-23
Unimplemented: PAE


Darker:
This is a message I get when booting Debian 8. I know that's a new system but I thought it would be really cool to run actual modern working OS in the emulator. What would be needed to fix that error? I'm a javascript programmer and I'd really like to help!

warpcoil:
PAE is the Physical Address Extensions of the Paging System on x86 platforms.  You can find out more here:

http://wiki.osdev.org/PAE

It shouldn't be too difficult to extend the existing paging mechanism in v86 . . .  I wouldn't be surprised if that was the only issue when booting Debian 8 though.

Darker:
Well, I don't know anything about operating systems and low level operation. I guess I could follow some documentation and verbatim implement what the documentation says.

copy:
PAE is described in [this manual](https://www-ssl.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-system-programming-manual-325384.pdf).

Normal 32-bit paging is implemented [here](https://www-ssl.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-manual-325462.pdf). PAE paging follows the same structure, but there is an extra indirection, the tables have different sizes and each entry is 64 bit.

Note that PAE is actually quite useless since JavaScript can't address that much memory anyway. Actually, it's bad for performance (you can see that PAE translation is more complex that normal 32-bit translation). It would be a better idea to install Debian with a non-pae kernel. I'll give it a try.

>  I wouldn't be surprised if that was the only issue when booting Debian 8 though.

The only thing that comes to mind is MMX/SSE.

copy:
Debian 8.4 boots perfectly fine after installing `linux-image-586`. Maybe I'll release this on the front page.

I'll leave this issue open for PAE anyway.

warpcoil:
Yeah, I done some experiments too . . .  Gentoo latest works as well

ayunami2000:
bump. even if it seems useless, many OSes (namely newer versions of windows) require this and refuse to boot without it.

iAmInActions:
Any updates regarding this?

copy:
v86 is small project with a small number of non-fulltime contributors. As far as I know, nobody is currently working on this. If you really want it, feel free to contribute it yourself.

iAmInActions:
im not good at wasm so i probably cant help

BenNottelling:
[...]

Issue #597, opened 2022-01-09
FreeNOS v.1.0.3 stops before logging in system


SuperMaxusa:
Before logging into the system, the speed in the emulator drops to 0 mIPS.

Settings:
RAM: 128 MB
VRAM: 8 MB
ACPI: enabled

Version (v86): 6b9d1f74d
Debug log: https://pastebin.com/339c1e8a
ISO: http://www.freenos.org/pub/FreeNOS/releases/X86/FreeNOS-1.0.3.iso
Client: Chrome 96 and Edge 97, Windows 10

![image](https://user-images.githubusercontent.com/41739128/148679986-d0f59367-d70f-4dfb-ad23-2da316b40b61.png)

copy:
I can reproduce this.

SuperMaxusa:
UPD: I was able to run FreeNOS v.0.0.1 (http://www.freenos.org/pub/FreeNOS/releases/X86/FreeNOS-0.0.1.iso), all versions above v.1.0.0 do not start.
![image](https://user-images.githubusercontent.com/41739128/148821544-77d0f6fc-dd36-44c7-804b-320548ee701e.png)

copy:
https://copy.sh/v86/?profile=freenos

PR #599, opened 2022-01-13
Add support for PAE (#97)

pwmarcz:
Physical memory is still limited to 32-bit addresses, but systems that enable PAE should work now.

## Ubuntu 12

I managed to boot the Ubuntu 12 ISO. However, it takes a lot of time to start and is too slow to do anything...

ubuntu12c

(During the boot, it complains about `soft lockup - CPU #0 stuck for 23s!`. Apparently it happens during the inintrd decompression `unlzma`, because it's pretty slow. However, if you just wait, it should go through).

## Serenity OS (#299)

PAE also should unlock Serenity OS. I tried it, but there are still some issues. After applying some workarounds, I managed to boot it and it looks like it initializes network, mouse, and starts various processes. However, graphics fail to display. I'll post an update later.

### DIFF
--- a/src/rust/cpu/cpu.rs
+++ b/src/rust/cpu/cpu.rs
@@ -20,8 +20,8 @@ use cpu::global_pointers::*;
 use cpu::memory;
 use cpu::memory::mem8;
 use cpu::memory::{
-    in_mapped_range, read8, read16, read32s, read64s, read128, read_aligned32, write8,
-    write_aligned32,
+    in_mapped_range, read8, read16, read32s, read64s, read128, read_aligned32,
+    read_aligned64, write8, write_aligned32,
 };
 use cpu::misc_instr::{
     adjust_stack_reg, get_stack_pointer, getaf, getcf, getof, getpf, getsf, getzf, pop16, pop32s,
@@ -1797,6 +1797,20 @@ pub unsafe fn do_page_translation(addr: i32, for_writing: bool, user: bool) -> O
     }
 }
 
+/*
+ * 32-bit paging:
+ * - 10 bits PD | 10 bits PT | 12 bits offset
+ * - 10 bits PD | 22 bits offset (4MB huge page)
+ *
+ * PAE paging:
+ * - 2 bits PDPT | 9 bits PD | 9 bits PT | 12 bits offset
+ * - 2 bits PDPT | 9 bits PD | 21 bits offset (2MB huge page)
+ *
+ * Note that PAE entries are 64-bit, and can describe physical addresses over 32
+ * bits. However, since we support only 32-bit physical addresses, we require
+ * the high half of the entry to be 0 (except for the execute-disable bit in
+ * PDE and PTE).
+ */
 pub unsafe fn do_page_walk(
     addr: i32,
     for_writing: bool,
@@ -1816,16 +1830,25 @@ pub unsafe fn do_page_walk(
     else {
         profiler::stat_increment(TLB_MISS);
 
-        let page_dir_addr = (*cr.offset(3) as u32 >> 2).wrapping_add((page >> 10) as u32) as i32;
-        let page_dir_entry = read_aligned32(page_dir_addr as u32);
-        // XXX
-        let kernel_write_override = !user && 0 == *cr & CR0_WP;
-        if 0 == page_dir_entry & PAGE_TABLE_PRESENT_MASK {
-            // to do at this place:
-            //
-            // - set cr2 = addr (which caused the page fault)
-            // - call_interrupt_vector  with id 14, error code 0-7 (requires information if read or write)
-            // - prevent execution of the function that triggered this call
+        let pae = *cr.offset(4) & CR4_PAE != 0;
+
+        let (page_dir_addr, page_dir_entry) =
+            match walk_page_directory(pae, addr) {
+                Some((a, e)) => (a, e),
+                // to do at this place:
+                //
+                // - set cr2 = addr (which caused the page fault)
+                // - call_interrupt_vector  with id 14, error code 0-7 (requires information if read or write)
+                // - prevent execution of the function that triggered this call
+                None => return Err(PageFault {
+                    addr,
+                    for_writing,
+                    user,
+                    present: false,
+                }),
+            };
+
+        if page_dir_entry & PAGE_TABLE_PRESENT_MASK == 0 {
             return Err(PageFault {
                 addr,
                 for_writing,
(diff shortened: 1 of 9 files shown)


src/rust/cpu/cpu.rs
```diff
@@ -1967,6 +1997,47 @@ pub unsafe fn do_page_walk(
...
+    pae: bool,
+    addr: i32,
+    page_dir_entry: i32
+) -> (i32, i32) {
+    let page_table = (page_dir_entry as u32 & 0xFFFFF000) >> 2;
+    if pae {
+        let page_table_idx = (addr as u32 >> 12) & 0x1FF;
+        let page_table_addr = page_table.wrapping_add(page_table_idx << 1);
+        let page_table_entry = read_aligned32(page_table_addr);
+        return (page_table_addr as i32, page_table_entry);
```
copy, line 2032: This is where the 64-bit address is read from, is that correct?

Could you change it to something like:

```
        let page_table_entry = read_aligned64(page_table_addr);
        dbg_assert!(page_table_entry & 0xFFFF_FFFF_0000_0000 == 0, "Unsupported: Physical 64-bit address");
        return (page_table_addr as i32, page_table_entry as i32);
```

src/rust/cpu/cpu.rs
```diff
@@ -1816,23 +1829,35 @@ pub unsafe fn do_page_walk(
...
+                // - prevent execution of the function that triggered this call
+                None => return Err(PageFault {
+                    addr,
+                    for_writing,
+                    user,
+                    present: false,
+                }),
+            };
+
+        if page_dir_entry & PAGE_TABLE_PRESENT_MASK == 0 {
```
copy, line 1850: Just in case you're wondering why this function contains so many casts: It used to be written in C where the casts were implicit. It was convert to Rust using c2rust, which made them explicit (by inserting `… as {i,u}32`). The general convention for the codebase is that virtual addresses are i32 and physical addresses u32. I will clean it up once this is merged.

copy:
Brilliant work, thanks. Only change I'd suggest is to assert when a 64-bit address is encountered, so we don't silently cut them off in debug mode.

By the way, I got an older version of SerenityOS working (September 2020). Will add to the website once this is merged.

Screenshot_2022-01-13_12-21-47

Newer versions seem to fail with an issue similar to #597.

src/rust/cpu/cpu.rs
pwmarcz: Interesting, I didn't know the project was converted from C!

src/rust/cpu/cpu.rs
pwmarcz: Done.

There's no `read_aligned64` and I wasn't sure what the convention for the argument should be. I went with "byte address >> 2", for consistency with `read_aligned32` (using "byte address >> 3" sounded like a more confusing option).

The highest bit of PDE and PTE (but not the highest entry, PDPTE) is execute-disable and can be set. I added comments about it.

copy:
Thanks!

copy:
@pwmarcz Out of curiosity, what workaround were you referring to?

> PAE also should unlock Serenity OS. I tried it, but there are still some issues. After applying some workarounds, I managed to boot it and it looks like it initializes network, mouse, and starts various processes.

pwmarcz:
@copy Here's what happened:
- I couldn't figure out how to boot from the QEMU image, but the extlinux one worked
- Serenity crashed (kernel panic on serial console) when initializing a pseudo-filesystem and adding information about ACPI; it's a null pointer exception since the ACPI subsystem failed to initialize. I commented out that part:
```diff
--- a/Kernel/Firmware/ACPI/Parser.cpp
+++ b/Kernel/Firmware/ACPI/Parser.cpp
@@ -69,6 +69,9 @@ UNMAP_AFTER_INIT ACPISysFSComponent::ACPISysFSComponent(NonnullOwnPtr<KString> t
 
 UNMAP_AFTER_INIT void ACPISysFSDirectory::find_tables_and_register_them_as_components()
 {
+    if (!ACPI::Parser::the())
+        return;
+
     NonnullRefPtrVector<SysFSComponent> components;
     size_t ssdt_count = 0;
     ACPI::Parser::the()->enumerate_static_tables([&](StringView signature, PhysicalAddress p_table, size_t length) {
```
* Then, apparently it did not like the fact that the emulated machine only has primary IDE channel, and got stuck trying to initialize the secondary IDE channel (`IDEChannel: reset failed, busy flag on master stuck`). When the secondary IDE device is absent, the port that Serenity looks at (IDE control) is treated by v86 as unmapped and hardcoded to return 0xFF.

  Unfortunately enabling the secondary IDE channel (i.e. adding a cdrom device, or an "empty" device) didn't work, I think extlinux failed to boot.

  The following workaround (in v86) worked for me:
```diff
--- a/src/cpu.js
+++ b/src/cpu.js
@@ -847,6 +847,8 @@ CPU.prototype.init = function(settings, device_bus)
         if(settings.cdrom)
         {
             this.devices.cdrom = new IDEDevice(this, settings.cdrom, undefined, true, ide_device_count++, device_bus);
+        } else {
+            this.io.register_read(0x0376, null, () => 0);
         }
 
         this.devices.pit = new PIT(this, device_bus);
```
* After that, Serenity still did not like v86's hard disk. I noticed that it tries to access it through addresses reported by PCI, and when you disable that and use the "standard" addresses, it works.
```diff
--- a/Kernel/Storage/ATA/IDEController.cpp
+++ b/Kernel/Storage/ATA/IDEController.cpp
@@ -133,9 +133,9 @@ UNMAP_AFTER_INIT void IDEController::initialize(bool force_pio)
     if (!is_bus_master_capable())
         force_pio = true;
 
-    auto bar0 = PCI::get_BAR0(pci_address());
+    auto bar0 = 0; //PCI::get_BAR0(pci_address());
     auto primary_base_io = (bar0 == 0x1 || bar0 == 0) ? IOAddress(0x1F0) : IOAddress(bar0 & (~1));
-    auto bar1 = PCI::get_BAR1(pci_address());
+    auto bar1 = 0; //PCI::get_BAR1(pci_address());
     auto primary_control_io = (bar1 == 0x1 || bar1 == 0) ? IOAddress(0x3F6) : IOAddress(bar1 & (~1));
     auto bar2 = PCI::get_BAR2(pci_address());
     auto secondary_base_io = (bar2 == 0x1 || bar2 == 0) ? IOAddress(0x170) : IOAddress(bar2 & (~1));
```
* At this point, I have an apparently working installation (according to serial logs), except for graphics - nothing appears on screen and the mode doesn't change.

I don't know anything about IDE or PCI, so it's hard for me to say if the above is the fault of v86 or Serenity.

PR #601, opened 2022-01-15
Add NodeOS URL to index.html

Unzor:
Changes:
- Add NodeOS URL to index.html to navigate to it easier


copy:
The Arch profile already includes node, and I'd like to keep the front page clean. Perhaps we need another way to discover profiles, but gonna close this for now.

PR #615, opened 2022-02-04
Remove duplicate link to create_file

Nikhil1920:


Nikhil1920:
So, I have just noticed that api.md does not exist and the links are broken is it  a work in progress?

copy:
> So, I have just noticed that api.md does not exist and the links are broken is it a work in progress?

It used to exist, now I'm mostly pointing directly at the code: https://github.com/copy/v86/blob/master/src/browser/starter.js#L1169

Feel free to add another commit to this PR that removes the link.

copy:
Thanks!

Issue #622, opened 2022-02-13
low resolution 2x scale overflows on many devices


pengan1987:
Commit 514c2ad3031fdc61db00fe91bf1cd7800b9e9717 add 2x scale to all resolution lower or equal to 640 width, when running a VGA(640x480) guest system on 1366x768 browser (still the 2nd popular resolution on the web) and 1440x900 or 1600x900 display will overflow since scaled picture has 960 height.

It's better to add a switch parameter to enable/disable this function

![autoscale_overflow](https://user-images.githubusercontent.com/2333797/153740477-3992bb60-86f4-45db-8394-ae7f9194bdf1.png)

copy:
I'd suggest a only applying the base scaling if it fits into `window.innerWidth` and `window.innerHeight`. Mind sending a PR?

PR #623, opened 2022-02-13
Only scale display when window big enough (#622)

pengan1987:
2x Scale graphic display only when 
- Guest system running in low resolution (width <= 640) graphic mode
- Host browser has enough window size to show 2x scale.


copy:
Thanks!

PR #624, opened 2022-02-16
fix v86_bin_fallback full path: getDirname (2)

proxy-m:
Make same dirname for `v86_bin_fallback` and `v86_bin` paths. It is needed when defined options["wasm_path"].

### DIFF
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -153,6 +153,17 @@ function V86Starter(options)
 
         "__indirect_function_table": wasm_table,
     };
+       
+    let getDirname = function (fullPath) {
+        let a = fullPath.split(/[/]/);
+        let back = false;
+        if (a.length <= 1) {
+            a = fullPath.split(/[\\]/);
+            back = true;
+        }
+        a.pop(); /* basename */
+        return (!back) ? a.join("/") : a.join("\\"); /* dirname */
+    };
 
     let v86_bin = DEBUG ? "v86-debug.wasm" : "v86.wasm";
     let v86_bin_fallback = "v86-fallback.wasm";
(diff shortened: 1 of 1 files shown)


copy:
v86-fallback.wasm is for old versions of Safari that don't support wasm bulk memory operations. It should never be loaded on Node, so I don't think this code is necessary.

proxy-m:
image

proxy-m:
> v86-fallback.wasm is for old versions of Safari that don't support wasm bulk memory operations. It should never be loaded on Node, so I don't think this code is necessary.

It is always loaded for me at Chromium browser on i386/i686 host system.

copy:
> It is always loaded for me at Chromium browser on i386/i686 host system.

Also in Node?

proxy-m:
https://github.com/proxy-m/jx86/commit/1dcd06efc10509e4aac852110d3597e960100f5a

```
$ arch
i686
$ node examples/nodejs.js 
Now booting, please stand by ...
v86_bin_fallback /home/proxym/Solutions/jx86/build/v86-fallback.wasm
~% whoami
root

```

proxy-m:
> > It is always loaded for me at Chromium browser on i386/i686 host system.
> 
> Also in Node?

Yes for Node 8.12 (and Node 8.10).

copy:
> Yes for Node 8.12 (and Node 8.10).

That's a very old version. Do you actually have a use-case for running a version that old?

proxy-m:
I often use it in production, for stability and speed, e.g. on embedded systems. Newer versions have more functions and just eat more size and more computation resources.

copy:
Makes sense. I won't guarantee supporting Node 8 indefinitely, but this is a good fix.

PR #625, opened 2022-02-21
Floppy image switched with bzimage (sugar for user)

proxy-m:
User can select bzImage/vmlinuz instead floopy disk image, difference detected by name (you can use other detection).
These commits adapted from my fork https://github.com/proxy-m/jx86

### DIFF
--- a/index.html
+++ b/index.html
@@ -75,7 +75,7 @@
             </tr>
 
             <tr>
-                <td><label for="floppy_image">Floppy disk image</label></td>
+                <td><label for="floppy_image">Floppy image / bzImage / vmlinuz.bin</label></td>
                 <td> <input type="file" id="floppy_image"><br></td>
             </tr>
 
--- a/src/floppy.js
+++ b/src/floppy.js
@@ -93,8 +93,18 @@ function FloppyController(cpu, fda_image, fdb_image)
         }
         else
         {
-            throw "Unknown floppy size: " + h(fda_image.byteLength);
+            console.error("Unknown floppy size: " + h(this.floppy_size));
+            floppy_type = ((this.floppy_size >> 10) > 1440) ?  floppy_types[2880] : floppy_types[1440];
+            console.log("Floppy type rounded: " + (Object.keys(floppy_types)[Object.values(floppy_types).indexOf(floppy_type)]));
+
+            cpu.devices.rtc.cmos_write(CMOS_FLOPPY_DRIVE_TYPE, floppy_type.type << 4);
+
+            sectors_per_track = floppy_type.sectors;
+            number_of_heads = floppy_type.heads;
+            number_of_cylinders = floppy_type.tracks;
+
         }
+        console.log("Floppy details: ", floppy_type); 
 
         this.sectors_per_track = sectors_per_track;
         this.number_of_heads = number_of_heads;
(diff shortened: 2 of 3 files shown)


src/floppy.js
```diff
@@ -93,8 +93,18 @@ function FloppyController(cpu, fda_image, fdb_image)
...
+            floppy_type = ((this.floppy_size >> 10) > 1440) ?  floppy_types[2880] : floppy_types[1440];
+            console.log("Floppy type rounded: " + (Object.keys(floppy_types)[Object.values(floppy_types).indexOf(floppy_type)]));
+
+            cpu.devices.rtc.cmos_write(CMOS_FLOPPY_DRIVE_TYPE, floppy_type.type << 4);
+
+            sectors_per_track = floppy_type.sectors;
+            number_of_heads = floppy_type.heads;
+            number_of_cylinders = floppy_type.tracks;
+
         }
```
copy, line 106: Out of curiosity, what guest OS is this for?

index.html
```diff
@@ -75,7 +75,7 @@ <h4>Setup</h4>
             </tr>
 
             <tr>
-                <td><label for="floppy_image">Floppy disk image</label></td>
+                <td><label for="floppy_image">Floppy image / bzImage / vmlinuz.bin</label></td>
```
copy, line 78: I think it would be better to add a separate input for this.

src/floppy.js
proxy-m: > Out of curiosity, what guest OS is this for?

I actually use bzImage or vmlinuz from /boot directory for fast tests. Buildroot+Busybox-like linux. Drafts: https://gitlab.com/mikb007/buildroot-dl-mini (console) , https://github.com/proxy-m/buildroot-nano-gcc4-linux26 (x11). Linux 4+ allows simple usage of built-in to kernel rootfs, if I remember (initramfs). So without second image (initrd).

Idea was to use minimal os as basic core.

_But you can also use floppy-storaged OS like KolibriOS, MenuetOS, [Plop boot plpbt.img](https://www.plop.at/en/bootmanager/plpbt.bin.html#runflp), [GAG Boot Manager](http://gag.sourceforge.net/), [REFInd BM](https://en.wikipedia.org/wiki/REFInd), Fdlinux, [OpenBSD floppy70.img](https://www.openbsd.org/faq/faq4.html#DosFlop), [Clover EFI (iso)](https://wiki.archlinux.org/title/Clover), [Grub4dos on floppy](https://github.com/chenall/grub4dos), [LILO BM](https://www.linuxquestions.org/questions/linux-newbie-8/can-you-put-lilo-onto-a-floppy-218/), [DSL floppy](http://www.damnsmalllinux.org/wiki/boot_floppies.html), [DR-DOS](https://en.wikipedia.org/wiki/DR-DOS), [SCP MS-DOS 1.25 (MIT)](https://en.wikipedia.org/wiki/MS-DOS), [FreeDOS 1.2 floppy](https://www.freedos.org/download/) or even [Pascal 7 (joke)](https://theylied.info/core/tp70/)_

index.html
proxy-m: Yes. I just wanted less buttons on screen for user.

proxy-m:
Also additional examples of floppy boot mages, but more common, dos: http://bootcd.narod.ru/images_e.htm

PR #629, opened 2022-03-06
Allow receiving wasm as a function via wasm_fn

giulioz:
Some bundlers allow directly importing a .wasm file ([https://vitejs.dev/guide/features.html#webassembly](https://vitejs.dev/guide/features.html#webassembly)).

This modification to the starter would allow receiving the wasm file via an import:
```js
import v86Wasm from "../build/v86.wasm";
const emu = new V86Starter({
    wasm_fn: v86Wasm,
});
```

### DIFF
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -200,13 +200,13 @@ function V86Starter(options)
                 }
             });
         });
-    }
+    };
     
     const loadWASM = async () =>
     {
-        const exports = options["wasm_fn"]
-            ? await options["wasm_fn"]({ "env": wasm_shared_funcs })
-            : await loadWASMFromUrl(v86_bin);
+        const exports = options["wasm_fn"] ?
+            await options["wasm_fn"]({ "env": wasm_shared_funcs }) :
+            await loadWASMFromUrl(v86_bin);
 
         wasm_memory = exports.memory;
         exports["rust_init"]();
@@ -215,7 +215,7 @@ function V86Starter(options)
         cpu = emulator.cpu;
 
         this.continue_init(emulator, options);
-    }
+    };
 
     loadWASM();
 }
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -175,54 +175,68 @@ function V86Starter(options)
         v86_bin_fallback = "build/" + v86_bin_fallback;
     }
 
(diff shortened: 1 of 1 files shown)


cekkr:
How could be used wasm functions?

giulioz:
@cekkr Vite, Webpack, Rollup and other bundlers provide a way to import wasm files, returning an async function that when executed returns the wasm exports.

Using this way of loading the v86 wasm file more flexibility is possible, since you don't have to specify an explicit url and you could potentially use even an array buffer (if you call WebAssembly.instantiate yourself).

More info is in this proposal: [https://github.com/WebAssembly/esm-integration/tree/main/proposals/esm-integration](https://github.com/WebAssembly/esm-integration/tree/main/proposals/esm-integration)

src/browser/starter.js
```diff
@@ -175,54 +175,68 @@ function V86Starter(options)
...
+        options["wasm_fn"]({ "env": wasm_shared_funcs })
+            .then((exports) => {
+                const imports = wasm_shared_funcs;
+                wasm_memory = exports.memory;
+                exports["rust_init"]();
+
+                const emulator = this.v86 = new v86(this.emulator_bus, { exports, wasm_table });
+                cpu = emulator.cpu;
+
+                this.continue_init(emulator, options);
```
copy, line 188: Could you refactor this into a function? (or, even better, refactor it to use async/await)

giulioz: Unfortunately since it's a constructor I can't make it async, is it ok if I make V86Starter call async stuff without returning a promise? Users would need to listen to emulator-ready

giulioz: Updated it btw

copy:
@giulioz I pushed a small refactor. Could you let me know if it looks good to you (including the interface of wasm_fn), and I'll push the squash-and-merge button

giulioz:
@copy I fixed the signature of wasm_fn and now it should be ok :) thanks for the refactor!

On a side note, I'm trying to maintain a bundler-friendly version of v86 on npm: https://www.npmjs.com/package/v86

copy:
> On a side note, I'm trying to maintain a bundler-friendly version of v86 on npm: https://www.npmjs.com/package/v86

Brilliant, thanks for that.

Issue #632, opened 2022-03-10
Use async/await


copy:
E.g. for load_files, see https://github.com/copy/v86/search?l=JavaScript&q=.then

copy:
@nitintejuja `make qemutests` or `make nasmtests` or `make kvm-unit-test`

PR #633, opened 2022-03-15
Fixed issue #632 async/await

nitintejuja:
@copy I have modified files and changed usages of .then to async/await and it fixes issue #632 . please review it .

### DIFF
--- a/src/browser/filestorage.js
+++ b/src/browser/filestorage.js
@@ -95,16 +95,14 @@ function ServerFileStorageWrapper(file_storage, baseurl)
  * @param {string} sha256sum
  * @return {!Promise<Uint8Array>}
  */
-ServerFileStorageWrapper.prototype.load_from_server = function(sha256sum)
+ServerFileStorageWrapper.prototype.load_from_server = async function(sha256sum)
 {
-    return new Promise((resolve, reject) =>
+    let data ;
+    v86util.load_file(this.baseurl + sha256sum, { done: buffer =>
     {
-        v86util.load_file(this.baseurl + sha256sum, { done: buffer =>
-        {
-            const data = new Uint8Array(buffer);
-            this.cache(sha256sum, data).then(() => resolve(data));
-        }});
-    });
+        data = new Uint8Array(buffer);
+    }});
+    return await this.cache(sha256sum, data);
 };
 
 /**
--- a/src/browser/filestorage.js
+++ b/src/browser/filestorage.js
@@ -95,14 +95,17 @@ function ServerFileStorageWrapper(file_storage, baseurl)
  * @param {string} sha256sum
  * @return {!Promise<Uint8Array>}
  */
-ServerFileStorageWrapper.prototype.load_from_server = async function(sha256sum)
+ServerFileStorageWrapper.prototype.load_from_server = function(sha256sum)
 {
-    let data ;
-    v86util.load_file(this.baseurl + sha256sum, { done: buffer =>
+    return new Promise((resolve, reject) =>
     {
-        data = new Uint8Array(buffer);
-    }});
-    return await this.cache(sha256sum, data);
+        v86util.load_file(this.baseurl + sha256sum, { done: async buffer =>
+        {
+            const data = new Uint8Array(buffer);
+            await this.cache(sha256sum, data);
+            resolve(data);
+        }});
+    });
 };
 
 /**
--- a/src/browser/filestorage.js
+++ b/src/browser/filestorage.js
@@ -103,8 +103,7 @@ ServerFileStorageWrapper.prototype.load_from_server = function(sha256sum)
         {
             const data = new Uint8Array(buffer);
             await this.cache(sha256sum, data);
-
-            return Promise.resolve(data);
+            return data;
         }});
     });
 };
--- a/src/browser/filestorage.js
+++ b/src/browser/filestorage.js
@@ -103,7 +103,7 @@ ServerFileStorageWrapper.prototype.load_from_server = function(sha256sum)
         {
             const data = new Uint8Array(buffer);
             await this.cache(sha256sum, data);
-            return data;
+            resolve(data);
         }});
     });
 };
--- a/src/browser/filestorage.js
+++ b/src/browser/filestorage.js
@@ -103,7 +103,8 @@ ServerFileStorageWrapper.prototype.load_from_server = function(sha256sum)
         {
             const data = new Uint8Array(buffer);
             await this.cache(sha256sum, data);
-            resolve(data);
+
+            return Promise.resolve(data);
         }});
     });
 };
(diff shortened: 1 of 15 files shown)


nitintejuja:
@copy can you help me to understand this undefined error ?

src/browser/speaker.js
```diff
@@ -439,7 +439,7 @@ PCSpeaker.prototype.start = function()
  * @param {!AudioContext} audio_context
  * @param {!SpeakerMixer} mixer
  */
-function SpeakerWorkletDAC(bus, audio_context, mixer)
+async function SpeakerWorkletDAC(bus, audio_context, mixer)
```
copy, line 442: This is a constructor, I don't think it can be async.

nitintejuja:
@copy i think all testcases passed . you can review it

copy (commented):
Sorry about the delay, I've been quite busy.

Could you also convert `v86util.load_file` to async? It would help simplify some of the other code using this function.

nitintejuja:
@copy are you saying about function implementation of load_file present in ```src/browser/lib.js``` ?

copy:
> @copy are you saying about function implementation of load_file present in `src/browser/lib.js` ?

Yes (it has a done callback that should be returned as a promise).

nitintejuja:
@copy i have modified files . you can review it .

src/browser/lib.js
```diff
@@ -1,5 +1,7 @@
 "use strict";
 
+const { resolve } = require("path");
```
copy, line 3: What is this for?

nitintejuja: unnecessary import . deleting it

src/browser/filestorage.js
```diff
@@ -99,10 +99,11 @@ ServerFileStorageWrapper.prototype.load_from_server = function(sha256sum)
 {
     return new Promise((resolve, reject) =>
     {
-        v86util.load_file(this.baseurl + sha256sum, { done: buffer =>
+        v86util.load_file(this.baseurl + sha256sum, { done: async buffer =>
```
copy, line 102: Could you change v86util.load_file so that we can get rid of the Promise constructor here and make this function async too?

src/browser/starter.js
```diff
@@ -1214,7 +1209,7 @@ V86Starter.prototype.create_file = function(file, data, callback)
  * @param {function(Object, Uint8Array)} callback
  * @export
  */
-V86Starter.prototype.read_file = function(file, callback)
+V86Starter.prototype.read_file = async function(file, callback)
```
copy, line 1212: (same as above). We should probably remove the callback and return a promise instead.

src/browser/starter.js
```diff
@@ -1174,7 +1169,7 @@ V86Starter.prototype.mount_fs = async function(path, baseurl, basefs, callback)
  * @param {function(Object)=} callback
  * @export
  */
-V86Starter.prototype.create_file = function(file, data, callback)
+V86Starter.prototype.create_file = async function(file, data, callback)
```
copy, line 1172: We should probably remove the callback and return a promise instead.

src/browser/starter.js
```diff
@@ -607,18 +603,17 @@ V86Starter.prototype.continue_init = async function(emulator, options)
 
             if(options["bzimage_initrd_from_filesystem"])
             {
-                const { bzimage, initrd } = this.get_bzimage_initrd_from_filesystem(settings.fs9p);
+                let { bzimage, initrd } = this.get_bzimage_initrd_from_filesystem(settings.fs9p);
```
copy, line 606: Could you rename these variables (e.g. to bzimage_path, initrd_path) and change the let back to a const?

src/cpu.js
```diff
@@ -1293,7 +1293,7 @@ CPU.prototype.codegen_finalize = function(wasm_table_index, start, state_flags,
         const result = new WebAssembly.Instance(module, { "e": this.jit_imports });
         const f = result.exports["f"];
 
-        this.codegen_finalize_finished(wasm_table_index, start, state_flags);
+        await this.codegen_finalize_finished(wasm_table_index, start, state_flags);
```
copy, line 1296: codegen_finalize_finished isn't async, so the await doesn't do anything.

copy (commented):
Apologies about the delay in reviewing. A few more things that should be improved before this can be merged.

src/browser/starter.js
nitintejuja: @copy  i have modified code

src/browser/filestorage.js
```diff
@@ -99,10 +99,14 @@ ServerFileStorageWrapper.prototype.load_from_server = function(sha256sum)
...
         {
             const data = new Uint8Array(buffer);
-            this.cache(sha256sum, data).then(() => resolve(data));
+            await this.cache(sha256sum, data);
+
+            return new Promise((resolve,reject)=>{
+                resolve(data);
+            });
```
undefined-moe, line 109: I think this can be simplified to `return Promise.resolve(data);` （?

src/browser/starter.js
```diff
@@ -1226,14 +1225,15 @@ V86Starter.prototype.read_file = function(file, callback)
...
+    const result = await fs.read_file(file);
+
+    return new Promise((resolve,reject) => {
+        if (result)
         {
-            callback(new FileNotFoundError(), null);
+            resolve(result);
+        } 
+        else{
+            reject(new FileNotFoundError());
```
undefined-moe, line 1236: the same, try this:

```ts
if (result) return Promise.resolve(result);
return Promise.reject(new FileNotFoundError());
```

src/browser/filestorage.js
nitintejuja: @undefined-moe i have modified this piece of code.

src/browser/starter.js
```diff
@@ -780,34 +778,22 @@ V86Starter.prototype.remove_listener = function(event, listener)
...
-        try
-        {
-            callback(null, this.v86.save_state());
-        }
-        catch(e)
-        {
-            callback(e, null);
-        }
-    }.bind(this), 0);
+    console.assert(arguments.length === 0);
```
undefined-moe, line 795: what about providing throw a errmsg like `Callback is deprecated, please use async/await` to notify user about the api change?

src/browser/starter.js
```diff
@@ -1197,15 +1180,11 @@ V86Starter.prototype.create_file = function(file, data, callback)
...
-            .then(() => callback(null));
+        await fs.CreateBinaryFile(filename, parent_id, data);
     }
     else
     {
-        setTimeout(function()
-        {
-            callback(new FileNotFoundError());
-        }, 0);
+        return Promise.reject(new FileNotFoundError());
```
undefined-moe, line 1187: I just forgot this simply equals to `throw new FileNotFoundError();` 🤣

undefined-moe (approved):

copy:
Thanks!

FYI, I removed some uses of async/await that weren't really async, and made some other public api functions async.

nitintejuja:
Thanks @copy and @undefined-moe

PR #655, opened 2022-05-17
Add dac_mask to fix text mode color

Pixelsuft:
Before:
before1
before2
before3
After:
after1
after2
after3


src/vga.js
```diff
@@ -254,6 +254,8 @@ function VGAScreen(cpu, bus, vga_memory_size)
     this.dac_color_index_read = 0;
     this.dac_state = 0;
 
+    this.dac_mask = 0xFF;
```
copy, line 257: FYI, I changed the initial value to 0xFF

copy:
Thanks!

PR #657, opened 2022-05-19
Mouse Wheel Support

Pixelsuft:
Mouse Wheel Support [Microsoft Intellimouse].
Sorry for my code, but it is working.
Tested under KolibriOS.

### DIFF
--- a/src/ps2.js
+++ b/src/ps2.js
@@ -597,7 +597,7 @@ PS2.prototype.port60_write = function(write_byte)
             this.sample_rate = 100;
             this.scaling2 = false;
             this.resolution = 4;
-            this.wheel_activate_counter = 0
+            this.wheel_activate_counter = 0;
             this.wheel_enabled = false;
             this.wheel_movement = 0;
             break;
@@ -615,7 +615,7 @@ PS2.prototype.port60_write = function(write_byte)
             this.scaling2 = false;
             this.resolution = 4;
 			
-            this.wheel_activate_counter = 0
+            this.wheel_activate_counter = 0;
             this.wheel_enabled = false;
             this.wheel_movement = 0;
 
--- a/src/ps2.js
+++ b/src/ps2.js
@@ -67,6 +67,13 @@ function PS2(cpu, bus)
     this.sample_rate = 100;
 
     /** @type {number} */
+    this.wheel_movement = 0;
+    /** @type {number} */
+    this.wheel_activate_counter = 0;
+    /** @type {boolean} */
+    this.wheel_enabled = false;
+
+    /** @type {number} */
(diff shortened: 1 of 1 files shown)


copy:
Nice work, thanks!

I tested this change with a few profiles. ReactOS, Haiku, SerenityOS and Windows 2000 also support the mouse wheel with this patch. 

In Damn Small Linux the mouse is completely broken by this patch. Could you have a look? (you can get it from here: http://distro.ibiblio.org/damnsmall/release_candidate/dsl-4.11.rc2.iso)

Pixelsuft:
DSL is really strange.
It looped requests to change rate 200->100->80 (for 0x03) when mouse move.
When I [randomly fixed it](https://github.com/Pixelsuft/v86/blob/master/src/ps2.js#L599), DSL is working 0x03 instead of 0x04 (I added it). I didn't test other OSes, but KolibriOS as working as well with 0x04.
0x04 allows to use 4th, 5th buttons ([not finished](https://github.com/Pixelsuft/v86/blob/master/src/ps2.js#L352)) and X scroll (finished, not tested).

copy:
Thanks. All modern OSes seem to work now. 4th/5th mouse button and horizontal scroll are not terribly useful in my opinion. Good to merge from my side.

src/ps2.js
```diff
@@ -331,6 +345,20 @@ PS2.prototype.send_mouse_packet = function(dx, dy)
     this.mouse_buffer.push(info_byte);
     this.mouse_buffer.push(delta_x);
     this.mouse_buffer.push(delta_y);
+    if (this.mouse_id == 0x04) {
+        this.mouse_buffer.push(
+			0 << 7 |
+			0 << 6 |
+			0 << 5 | // TODO: 5th button
+			0 << 4 | // TODO: 4th button
+			this.wheel_movement
```
copy, line 354: Should this be masked to the lower four bits? Otherwise the upper bits are set if wheel_movement is negative.

src/ps2.js
```diff
@@ -106,8 +115,13 @@ function PS2(cpu, bus)
 
     this.bus.register("mouse-wheel", function(data)
     {
-        // TODO: Mouse Wheel
-        // http://www.computer-engineering.org/ps2mouse/
+        this.wheel_movement -= data[0];
+        this.wheel_movement -= data[1] * 2; // X Wheel Movement
```
copy, line 119: Why `data[1] * 2`? Shouldn't `data[1]` be ignored if we don't implement x wheel support?

copy: (answering my own question: Yes, this is how it's specified to work)

PR #665, opened 2022-05-27
Add keyboard lock support

hello-smile6:
Allows system keybindings to be intercepted by the guest. I've been using this for a while now myself.

### DIFF
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -1029,7 +1029,10 @@ V86Starter.prototype.screen_go_fullscreen = function()
         var focus_element = document.getElementsByClassName("phone_keyboard")[0];
         focus_element && focus_element.focus();
     }
-
+    // Use keyboard lock when possible
+    if(navigator?.keyboard?.lock instanceof Function) {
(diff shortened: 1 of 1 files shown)


src/browser/starter.js
```diff
@@ -1029,7 +1029,10 @@ V86Starter.prototype.screen_go_fullscreen = function()
         var focus_element = document.getElementsByClassName("phone_keyboard")[0];
         focus_element && focus_element.focus();
     }
-
+    // Use keyboard lock when possible
+    if(navigator?.keyboard?.lock instanceof Function) {
+        try {navigator.keyboard.lock();}catch(e) {};
+    }
```
copy, line 1035: `?` isn't supported by our build tools, how about:

```
try {
   navigator.keyboard.lock();
}
catch(e) {}
```

hello-smile6: @copy Okay, I thought I should avoid that, because it can fail for other reasons than a lack of support.

hello-smile6: Resolved in c66ad25fb9b825d30b54ce7b1c66e40f7528d67b

PR #666, opened 2022-05-28
docs: add instructions for Windows XP

stasoid:


hello-smile6:
Your computer's fast!

docs/windows-xp.md
```diff
@@ -49,7 +49,7 @@ onload = function()
 </div>
 ```
 To open this html file local http server is needed. Chrome has `--allow-file-access-from-files` command line option which allows using XHR to read local files, but it doesn't work in this case, probably because of `Range` header (`async: true`). Standard Python server `python -m SimpleHTTPServer` and Ruby server `ruby -run -e httpd` can't handle partial downloads either.  
-You can use [nodejs server](https://www.npmjs.com/package/http-server) or [this python server](https://github.com/smgoller/rangehttpserver).  
+You can use [nodejs server](https://www.npmjs.com/package/http-server) or [this python server](https://github.com/smgoller/rangehttpserver) (much slower).
```
hello-smile6, line 52:
```suggestion
You can use [the Node.JS `http-server` package](https://www.npmjs.com/package/http-server). Or, if Python is more convenient, you could use [RangeHTTPServer](https://github.com/smgoller/rangehttpserver) (much slower).
## What if I can't run an HTTP server?
If you can't use an HTTP server, just use [the website](https://copy.sh/v86). For the hard drive, select your Windows XP disk image. Set the memory to 512 MB, and Windows XP will boot.
```

hello-smile6 (changes requested):
Don't make it seem like people can't use the website.

docs/windows-xp.md
```diff
@@ -62,4 +65,4 @@ Windows XP load time (until start button becomes responsive) in Chrome on my com
 * 12 min second time if Network tab in Developer Tools is open
 * 17 min (first time or if cache is disabled) and Network tab in Developer Tools is open
 
-Sometimes Windows XP hangs during boot in v86, displaying only desktop wallpaper without taskbar or desktop icons.
\ No newline at end of file
+Sometimes Windows XP hangs during boot in v86, displaying only desktop wallpaper without taskbar or desktop icons.
```
hello-smile6, line 68: Technically, this would be after boot.
```suggestion
Sometimes Windows XP hangs after boot (before it is interactive) in v86, displaying only desktop wallpaper without taskbar or desktop icons.
```

stasoid:
Are there any problems with this PR? Why it is not getting merged?

copy:
Thanks!

PR #679, opened 2022-06-08
PS2 wheel code improvements + winnt fix

Pixelsuft:
I've looked in the [ps2 qemu code](https://github.com/qemu/qemu/blob/master/hw/input/ps2.c) and changed something.
About Windows NT 4.0:
It sends 200->100->80 to use 0x03 mouse mode, then it resets mouse and mouse_id is now 0x00. But windows is still working like with 0x03 mode.
But then it sets rate to 60, so, let's detect 200->100->80->60 and turn on workaround mouse_id reset.
In the qemu it's as working as well without any fixes.
But windows with SP6 uses default 0x00 mouse type (it's changing rate 20->40->60->60), so, that can be partically a windows bug.

### DIFF
--- a/src/ps2.js
+++ b/src/ps2.js
@@ -67,10 +67,13 @@ function PS2(cpu, bus)
     this.sample_rate = 100;
 
     /** @type {number} */
-    this.intellimouse_activation_counter = 0;
+    this.mouse_detect_state = 0;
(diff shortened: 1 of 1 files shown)


Pixelsuft:
oops

PR #687, opened 2022-06-15
char size for text mode

Pixelsuft:
char_width can be 9 or 8.
char_wide is scaling text screen 2 times by X. (for example, tetris game in msdos profile)


Pixelsuft:
In other emulators, when font size is changing, x pos still same:
halfix
But I don't know how to do it there, so I'm just scaling them by x:
v86

Pixelsuft:
I think it has bad realisation

PR #689, opened 2022-06-16
Record that Ubuntu 16 works

hello-smile6:
I have tested Ubuntu 16, and it works with v86.


catplanet007:
Can you give an example of how you did that? I tried https://github.com/copy/v86/pull/771 but doesn't work, it's stucked when running ./build-state.js

PR #693, opened 2022-06-19
DAC Fix for Win9x bootlogo progressbar

Pixelsuft:
I don't know how, but I've randomly fixed it.
Before:
before
After:
after

### DIFF
--- a/src/vga.js
+++ b/src/vga.js
@@ -1544,11 +1544,11 @@ VGAScreen.prototype.port3C9_write = function(color_byte)
     var index = this.dac_color_index_write / 3 | 0,
         offset = this.dac_color_index_write % 3,
         color = this.vga256_palette[index];
-	
-	var b = color_byte & 1;
+
 	color_byte &= 0x3F;
     if (!(this.dispi_enable_value & 0x20))
 	{
+		var b = color_byte & 1;
 		color_byte = color_byte << 2 | b << 1 | b;
 	}
 
--- a/src/vga.js
+++ b/src/vga.js
@@ -1578,7 +1578,7 @@ VGAScreen.prototype.port3C9_read = function()
     var color = this.vga256_palette[index];
 
     this.dac_color_index_read++;
-    return (color >> (2 - offset) * 8 & 0xFF) / 255 * 63 | 0;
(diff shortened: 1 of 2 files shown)


src/vga.js
```diff
@@ -1578,7 +1578,7 @@ VGAScreen.prototype.port3C9_read = function()
     var color = this.vga256_palette[index];
 
     this.dac_color_index_read++;
-    return (color >> (2 - offset) * 8 & 0xFF) / 255 * 63 | 0;
+    return (color >> (2 - offset) * 8 & 0xFF) / 256 * 64 | 0;
```
copy, line 1581: Is this correct though? It should probably be the inverse of the formula on line 1548. I can't find any reference on how the hardware converts 18-bit to 24-bit colours.

Pixelsuft:
QEMU:
```c
static inline int c6_to_8(int v)
{
    int b;
    v &= 0x3f;
    b = v & 1;
    return (v << 2) | (b << 1) | b;
}
```
Halfix:
```c
static uint8_t c6to8(uint8_t a)
{
    if (vga.vbe_enable & VBE_DISPI_8BIT_DAC)
        return a;
    uint8_t b = a & 1;
    return a << 2 | b << 1 | b;
}
```

copy:
> QEMU:
> Halfix:

Watch out, we can't use GPL code in v86 without tainting the license (I'll accept this one since the code is trivial).

src/vga.js
```diff
@@ -1545,7 +1545,12 @@ VGAScreen.prototype.port3C9_write = function(color_byte)
         offset = this.dac_color_index_write % 3,
         color = this.vga256_palette[index];
 
-    color_byte = (color_byte & 0x3F) * 255 / 63 | 0;
+	color_byte &= 0x3F;
+    if (!(this.dispi_enable_value & 0x20))
+	{
+		var b = color_byte & 1;
+		color_byte = color_byte << 2 | b << 1 | b;
+	}
```
copy, line 1553: Should probably update the read code too.

hello-smile6:
> > QEMU:
> > Halfix:
> 
> Watch out, we can't use GPL code in v86 without tainting the license (I'll accept this one since the code is trivial).

Why don't you want GPL code?

copy:
Cheers!

PR #696, opened 2022-06-23
Update Debian build info

shanginn:
This MR addresses 2 issues:
## In the README about building Debian image 1 step is missing.
You need to add a new profile into main.js file in order for http://localhost/debug?profile=debian to load something, otherwise it just loads default debug page instead of actual image.

## There is an error while building Debian image
Error looks like this: 
> WARNING: The requested image’s platform (linux/386) does not match the detected host platform (linux/amd64) and no specific platform was requested

Adding platform in the build script explicitly fixes the issue.

ps: I guess I should have opened separate merge requests for each issue...?

### DIFF
--- a/src/browser/main.js
+++ b/src/browser/main.js
@@ -179,17 +179,6 @@
         // Abandonware OS images are from https://winworldpc.com/library/operating-systems
         var oses = [
             {
-                id: "debian",
-                name: "Debian",
-                memory_size: 512 * 1024 * 1024,
-                vga_memory_size: 8 * 1024 * 1024,
-                state: { url: host + "debian-state-base.bin" },
-                filesystem: {
-                    "baseurl": host + "debian-9p-rootfs-flat/",
-                    "basefs": { url: host + "debian-base-fs.json" }
-                }
-            },
-            {
                 id: "archlinux",
                 name: "Arch Linux",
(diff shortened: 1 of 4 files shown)


src/browser/main.js
```diff
@@ -178,6 +178,17 @@
...
+                id: "debian",
+                name: "Debian",
+                memory_size: 512 * 1024 * 1024,
+                vga_memory_size: 8 * 1024 * 1024,
+                state: { url: host + "debian-state-base.bin" },
+                filesystem: {
+                    "baseurl": host + "debian-9p-rootfs-flat/",
+                    "basefs": { url: host + "debian-base-fs.json" }
+                }
+            },
```
copy, line 191: The problem with this is that it will create a broken debian profile on the deployed version of v86. One of the following would be better:

1. Add a debian example in the examples directory
2. Add this change to the readme

I'd prefer 1. but would accept 2. too.

copy:
> ps: I guess I should have opened separate merge requests for each issue...?

No, it's better in a single PR.

src/browser/main.js
shanginn: did both, how does that look?

copy: Looks good, thanks!

copy:
Cheers!

PR #702, opened 2022-07-12
Allow sharing files with guest using UI, even with custom profile

hello-smile6:
The 9p filesystem is always accessible by the guest, even if the UI is hidden. This will make the file picker UI be shown when needed; all the user has to do is run `ls` on the 9p mount.

### DIFF
--- a/lib/9p.js
+++ b/lib/9p.js
@@ -733,6 +733,7 @@ Virtio9p.prototype.ReceiveRequest = async function (bufchain) {
             marshall.Marshall(["Q"], [inode.qid], this.replybuffer, 7);
             this.BuildReply(id, tag, 13);
             this.SendReply(bufchain);
+            this.bus.send("9p-attach");
             break;
 
         case 108: // tflush
--- a/src/browser/main.js
+++ b/src/browser/main.js
@@ -1263,12 +1263,15 @@
         $("runtime_options").style.display = "block";
         $("runtime_infos").style.display = "block";
         $("screen_container").style.display = "block";
-
-        if(settings.filesystem)
-        {
-            init_filesystem_panel(emulator);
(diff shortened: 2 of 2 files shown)


src/browser/main.js
```diff
@@ -1263,12 +1263,15 @@
...
-
-        if(settings.filesystem)
-        {
-            init_filesystem_panel(emulator);
+        init_filesystem_panel(emulator);
+        if(!(settings.filesystem)) {
+            // Hide the filesystem panel for now
+            $("filesystem_panel").style.display="none";
+            // Show the filesystem panel if it's ever used
+            emulator.add_listener("9p-read-start", function(args) {
```
copy, line 1271: I'd suggest to add `this.bus.send("9p-attach");` here: https://github.com/copy/v86/blob/master/lib/9p.js#L736
and use it instead of `9p-read-start`.

hello-smile6: Okay

lib/9p.js
```diff
@@ -733,6 +733,7 @@ Virtio9p.prototype.ReceiveRequest = async function (bufchain) {
             marshall.Marshall(["Q"], [inode.qid], this.replybuffer, 7);
             this.BuildReply(id, tag, 13);
             this.SendReply(bufchain);
+            this.bus.send("9p-attach");
```
hello-smile6, line 736: Will this have a performance impact?

copy: I don't think so.

copy:
Cheers!

hello-smile6:
@copy Works perfectly! 🎉
image
That's Alpine Linux (the -virt disk image) on https://v86-9pfs.vercel.app/ (a v86 deployment with this)

PR #704, opened 2022-07-12
Connect guest network adapters

hello-smile6:
You'll still need to assign a static IP to each VM, but they can communicate!


copy:
Cheers!

PR #706, opened 2022-07-12
Add type declarations for PS2 code

hello-smile6:


copy:
Doesn't look very useful to me at this stage.

PR #720, opened 2022-08-13
Fix V86.destroy() with xterm.js

jmezzacappa:
`V86.destroy()` was throwing an error due to `SerialAdapterXtermJS`
having no `destroy()` method.

### DIFF
--- a/src/browser/serial.js
+++ b/src/browser/serial.js
@@ -240,8 +240,8 @@ function SerialAdapterXtermJS(element, bus)
     }, this);
 
     this.destroy = function() {
-        on_data_disposable["dispose"]()
-        term["dispose"]()
+        on_data_disposable["dispose"]();
+        term["dispose"]();
     };
 }
 
--- a/src/browser/serial.js
+++ b/src/browser/serial.js
@@ -234,21 +234,15 @@ function SerialAdapterXtermJS(element, bus)
         }
     });
 
-    bus.register("serial0-output-char", serial_char_handler, this);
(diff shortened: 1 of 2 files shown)


src/browser/serial.js
```diff
@@ -227,17 +227,29 @@ function SerialAdapterXtermJS(element, bus)
...
             bus.send("serial0-input", data.charCodeAt(i));
         }
     });
 
-    bus.register("serial0-output-char", function(chr)
+    bus.register("serial0-output-char", serial_char_handler, this);
+
+    this.destroy = function()
+    {
+        bus.unregister("serial0-output-char", serial_char_handler);
```
copy, line 241: This shouldn't be necessary. I think it should be removed, unless there is another reason to keep it.

jmezzacappa: I think you're right. I added this while attempting to fix a memory leak in my project. My thought was that there would be a useless mutual reference between xterm.js and v86, so keeping one around could prevent the other from being garbage collected.

In hindsight, it probably would not make sense to keep _either_ of those instances around if they've both been "destroyed"/"disposed". I'll try removing it to see if my memory leak returns.

jmezzacappa: I have confirmed that this line is not necessary and removed it.

jmezzacappa:
Oops, sorry for my formatting habits :) I'll try to remember to run jshint locally next time.

copy:
Cheers!

PR #721, opened 2022-08-13
Wait for emulation to stop when destroying

jmezzacappa:
In some environments, the "emulator-stopped" event did not always fire when calling the `destroy()` method. This waits for emulation to finish stopping before continuing with the rest of the destructor.

My use case, for more context:
I'm running v86 in a React app, and the page does not reload between destroying one emulator instance and starting another. Setting event listeners for the "emulator-stopped" event seemed to prevent garbage collection, so it quickly consumed lots of memory and would crash the browser tab.

### DIFF
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -698,16 +698,31 @@ V86Starter.prototype.run = async function()
  */
 V86Starter.prototype.stop = async function()
 {
-    this.bus.send("cpu-stop");
+    // Emulation is not guaranteed to stop immediately in the same event loop
+    // cycle.
+    // If the emulator is running, this promise will resolve when the emulator
+    // has finished stopping.
+    // Else, this promise will resolve immediately.
+    const promise = new Promise(resolve => {
+        if (!this.cpu_is_running) return resolve();
+        const listener = () => {
+            this.remove_listener("emulator-stopped", listener);
+            resolve();
+        };
+        this.add_listener("emulator-stopped", listener);
+        this.bus.send("cpu-stop");
+    })
+
+    await promise;
 };
 
 /**
  * @ignore
  * @export
  */
-V86Starter.prototype.destroy = function()
+V86Starter.prototype.destroy = async function()
 {
-    this.stop();
+    await this.stop();
 
     this.v86.destroy();
     this.keyboard_adapter && this.keyboard_adapter.destroy();
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -708,9 +708,8 @@ V86Starter.prototype.stop = async function()
             resolve();
         };
         this.add_listener("emulator-stopped", listener);
+        this.bus.send("cpu-stop");
     });
-
-    this.bus.send("cpu-stop");
 
     await stopped;
 };
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -698,22 +698,21 @@ V86Starter.prototype.run = async function()
  */
 V86Starter.prototype.stop = async function()
 {
-    // Emulation is not guaranteed to stop immediately in the same event loop
-    // cycle.
-    // If the emulator is running, this promise will resolve when the emulator
-    // has finished stopping.
-    // Else, this promise will resolve immediately.
-    const promise = new Promise(resolve => {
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -10,7 +10,7 @@ function v86(bus, wasm)
     this.running = false;
 
     /** @type {boolean} */
-    this.stopped = false;
+    this.stopping = false;
```
jmezzacappa, line 13: Renaming this was not necessary, so I put it into a separate commit (with explanation) so it would be easier to revert, if desired.

copy: `stopping` is a better name, thanks.

src/browser/starter.js
```diff
@@ -698,16 +698,31 @@ V86Starter.prototype.run = async function()
...
+        if (!this.cpu_is_running) return resolve();
+        const listener = () => {
+            this.remove_listener("emulator-stopped", listener);
+            resolve();
+        };
+        this.add_listener("emulator-stopped", listener);
+        this.bus.send("cpu-stop");
+    })
+
+    await promise;
```
copy, line 716:
```suggestion
    if(!this.cpu_is_running)
    {
        return;
    }

    const stopped = new Promise(resolve => {
        const listener = () => {
            this.remove_listener("emulator-stopped", listener);
            resolve();
        };
        this.add_listener("emulator-stopped", listener);
    });

    this.bus.send("cpu-stop");

    await stopped;
```

(I suggest removing the comment, since it expresses the same idea as the code itself. For reference as to why the code is written this way, one can always refer to `git blame`).

copy (commented):
This is a good change, thanks. Just one small stylistic suggestion and should be good to merge.

src/browser/starter.js
jmezzacappa: Ah, right, that is much nicer :)

Would you mind if I move `this.bus.send("cpu-stop");` to line 712, so it's clearly guaranteed to be called immediately after `this.add_listener("emulator-stopped", listener);` within the same anonymous function?

In other words, if I'm debugging some complex race condition, there would be no reason to question whether side effects triggered by `"cpu-stop"` have taken effect before the anonymous function passed into `new Promise()` has been called.

(There's also a good chance that I'm overthinking this, as I have been experimenting with alternative JS runtimes lately 😅)

jmezzacappa: To save time, I went ahead and pushed a commit that applies your suggested changes, and a separate commit that applies my aforementioned modification (moving `this.bus.send("cpu-stop");`). The last one can be reverted if you prefer to stick with the suggested changes :)

src/browser/starter.js
```diff
@@ -698,16 +698,29 @@ V86Starter.prototype.run = async function()
...
+    const stopped = new Promise((resolve) => {
+        const listener = () => {
+            this.remove_listener("emulator-stopped", listener);
+            resolve();
+        };
+        this.add_listener("emulator-stopped", listener);
+        this.bus.send("cpu-stop");
+    });
+
+    await stopped;
```
copy, line 714: Change to `await new Promise`.

src/browser/starter.js
```diff
@@ -698,16 +698,29 @@ V86Starter.prototype.run = async function()
  */
 V86Starter.prototype.stop = async function()
 {
-    this.bus.send("cpu-stop");
+    if (!this.cpu_is_running) {
+        return;
+    }
+
+    const stopped = new Promise((resolve) => {
```
copy, line 705: Remove the parenthesis around `(resolve)`

src/browser/starter.js
```diff
@@ -698,16 +698,29 @@ V86Starter.prototype.run = async function()
  */
 V86Starter.prototype.stop = async function()
 {
-    this.bus.send("cpu-stop");
+    if (!this.cpu_is_running) {
+        return;
+    }
```
copy, line 703: Preserve the style of the other code (remove space after `if`, put `{` on its own line).

copy (commented):
Few more stylistic fixes.

src/browser/starter.js
copy: It's probably easier to reason about, indeed.

copy:
Squashed and merged, cheers!

PR #723, opened 2022-08-15
Faster `BusConnector.prototype.unregister()`

jmezzacappa:
By storing the event listeners in a set, it eliminates the need to
iterate over an array to remove each listener. `.unregister()` is now an
O(1) operation (vs. the previous O(n)).

### DIFF
--- a/src/browser/worker_bus.js
+++ b/src/browser/worker_bus.js
@@ -13,9 +13,8 @@ WorkerBus.Connector = function(pair)
         var data = e.data;
         var listeners = this.listeners[data[0]];
 
-        for(var i = 0; i < listeners.length; i++)
+        for(const listener of listeners)
         {
-            var listener = listeners[i];
             listener.fn.call(listener.this_value, data[1]);
         }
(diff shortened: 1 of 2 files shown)


jmezzacappa:
By the way, I'm embarrassed that this is my third pull request, and I have yet to say "thank you".
This has quickly become one of my favorite repos in all of GitHub, and has enabled some very cool projects. Thank you @copy and other contributors for v86!

copy:
This is a reasonable change, but I'm afraid it will slow down `send` (as iterating a map is slower than an array), which is called much more often than unregister.

> By the way, I'm embarrassed that this is my third pull request, and I have yet to say "thank you".
> This has quickly become one of my favorite repos in all of GitHub, and has enabled some very cool projects. Thank you @copy and other contributors for v86!

Glad you're enjoying it and thanks for your contributions!

jmezzacappa:
Ohh, I did not think of that! Slowing down `send` would be a terrible trade-off. I'll close this one :)

PR #755, opened 2022-10-30
Add support for fixed_chunk_size to async Range downloads.

asiekierka:
This allows reducing the number of separate HTTP requests, which considerably improves load times at the cost of increased bandwidth use. Tested locally, to a limited extent.

### DIFF
--- a/src/buffer.js
+++ b/src/buffer.js
@@ -203,7 +203,7 @@
         if(this.fixed_chunk_size)
         {
             requested_start = offset - (offset % this.fixed_chunk_size);
-            requested_length = (Math.floor((offset + len - 1) / this.fixed_chunk_size) + 1 - Math.floor(requested_start / this.fixed_chunk_size)) * this.fixed_chunk_size;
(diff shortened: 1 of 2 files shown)


src/buffer.js
```diff
@@ -194,14 +198,29 @@
             return;
         }
 
+        var requested_start = offset;
+        var requested_length = len;
+        if(this.fixed_chunk_size)
+        {
+            requested_start = offset - (offset % this.fixed_chunk_size);
+            requested_length = (Math.floor((offset + len - 1) / this.fixed_chunk_size) + 1 - Math.floor(requested_start / this.fixed_chunk_size)) * this.fixed_chunk_size;
```
copy, line 206: I think this would be more readable (and in line with `AsyncXHRPartfileBuffer.prototype.get`):

```js
requested_length = Math.ceil((offset - requested_start + len) / this.fixed_chunk_size) * this.fixed_chunk_size;
```

copy (commented):
Thanks! Just a small request, then it should be good to merge.

src/buffer.js
asiekierka: Done!

copy:
Cheers!

PR #767, opened 2022-11-20
Correct docker image for no limiter

BelleNottelling:
I've deleted that docker hub account and that image is now published under `bellenottelling/websockproxy`


copy:
Cheers!

PR #804, opened 2023-02-20
VGA text mode should display ⌂ for 0x7F, not DEL.

LCeionia:
To match the rest of the character set currently used (appearing to be Code page 437), VGA text modes should output ⌂ for 0x7F rather than a 0x7F (DEL) character. This prevents the VGA text output from losing a character in the output when an on screen character is 0x7F, and brings the text display in line with behavior on most real hardware.


copy:
Nice find, cheers!

PR #820, opened 2023-03-14
Make screenshot function return an Image

JoeOsborn:
Returning an image rather than always opening a new window slightly expands the utility of embedding v86---we may want to take a screenshot and use just its data without asking the user to download it.


copy:
Cheers!

PR #846, opened 2023-04-26
Replaced force_disable_jit constant by propagating boolean argument up the call chain 

kmahapatra530:


copy:
Thanks! Would be nice to have this in a way that doesn't affect performance, but I'll merge and address that later.

PR #869, opened 2023-05-29
fix queue selected

Joinhack:


copy:
Thanks! Our of curiosity, how did you find the bug?

Joinhack:
> 

I'm try to implement the v86 in WASI, so I read and study the code.

Issue #882, opened 2023-06-19
Debian 12 still wont work


LeoKids:
![fail2](https://github.com/copy/v86/assets/68135774/2298d0bd-a76e-4b6f-a915-89cd55539b9b)
![fail](https://github.com/copy/v86/assets/68135774/dfd755f6-f7a9-4211-a178-ec8cc9e25710)

LeoKids:
Installed version,netinstall

SuperMaxusa:
Lastest Debian 12 with 1024 MB RAM and disabled ACPI:
![debian](https://github.com/copy/v86/assets/41739128/4bce0be4-c8aa-4f8a-ae26-ae634e7402f5)

As far as I understand, those errors with `mce` and `ACPI BIOS Error` should not seem to affect the system startup.

Can you try press key `e` in GRUB menu, change `quiet` on `acpi=off console=ttyS0 console=tty0`, press _Ctrl+X_ and check startup log?

copy:
Correct, these errors can be ignored. Everything working as expected, as far as I can tell.

LeoKids:
The next time I try to install,I will ignore it and it will get to the setup.

PR #885, opened 2023-06-26
Update Readme.md

LeoKids:
According to #882,these errors I had can be ignored so here.


copy:
Doesn't sounds too useful to have this in the readme.

Issue #54, opened 2015-04-10
Change removable media (floppy and cdrom) while running


Creeper20428:
When using this emulator to run Windows XP, I encounter the program "setup". However, it tells me to insert the next image into drive A. Where's Drive A? Jamming a 1.44MB floppy into my laptop would probably destroy my computer and not help at all. So is there a way?

A possible solution is to upload the images into / (maybe A:) but I haven't had any luck using the filesystem yet...

Thank you in advance.

sinni800:
Agreed, a little button to switch out images would be awesome.

Creeper20428:
You would need this feature to install operating systems from the 1980s where they had like 10 install disks for each version of DOS.

abundant-ray:
5 years later and im guessing this will never be added in?

JoeOsborn:
I’m not clear on what the OS (BIOS?) is looking for to notice fdd changes.  According to https://wiki.osdev.org/Floppy_Disk_Controller it sounds like changing  port3F7_read to send a 1 in bit 7 when the disk door is opened (disk removed?) and again when closed (disk inserted?) should let us change the underlying buffer used by the floppy controller smoothly.  Does that sound about right?

JoeOsborn:
As for CD-ROM, I think the IDEInterface would need to change in a few ways...  Here's my understanding of the problem based on the osdev wiki ( https://wiki.osdev.org/PCI_IDE_Controller#Ejecting_an_ATAPI_Drive ) and https://inst.eecs.berkeley.edu/~cs150/fa01/labs/project/ATAPI_Spec.pdf .

First off, when a disk is inserted, the ide controller should generate a unit attention condition that persists until it's cleared by the guest when it sends a "request sense" about it; my understanding is that this should cause all reads and stuff to error out with a "unit attention" error.  At that point the controller can do what it was doing before, but with a new buffer.  I think this is the first thing to implement: a button to load a CDROM into an empty cdrom controller.

For the emulator user to trigger an eject, we can do it by telling the interface to respond to queries with an error status with bit 3 (Media Change Requested) set.  Then the controller should wait for the guest to send an Acknowledge Media Change command (code DB), after which the guest promises to send the eject sequence.  (For a less graceful way, we could just swap the buffers and set that unit attention condition from before, but then the guest OS doesn't get a chance to e.g. prevent the eject or things like that).

[...]

PR #897, opened 2023-08-16
Floppy eject/insert

JoeOsborn:
This patch implements two new methods on FloppyController, `eject_fda` and `insert_fda`.  These are not exposed to the UI yet, but from the browser console in debug mode they work for removing and inserting floppies in MS-DOS at least.

Caveats:
1. FloppyController does not implement the format track command, so you need to give it a pre-formatted disk image.
2. FloppyController does not seem to emulate all the relevant internal controller state, so I made a best-effort patch here (references: [this website](https://www.isdaman.com/alsos/hardware/fdc/floppy.htm) , [osdev](https://wiki.osdev.org/Floppy_Disk_Controller) ).  I did what I could around reset behavior, extending existing commands, and reporting errors on seeks and reads, but this is pretty far outside my wheelhouse.

So, please see this PR as a request for testing and review towards part of #54 .

### DIFF
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -994,7 +994,7 @@ V86Starter.prototype.set_fda = async function(file)
         {
             this.v86.cpu.devices.fdc.set_fda(image);
         };
-        image.load();
+        await image.load();
     }
 };
 
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -395,77 +395,20 @@ V86Starter.prototype.continue_init = async function(emulator, options)
             file.async = false;
         }
 
-        if(file.buffer instanceof ArrayBuffer)
+        if(file.url && !file.async)
         {
-            var buffer = new v86util.SyncBuffer(file.buffer);
             files_to_load.push({
                 name: name,
-                loadable: buffer,
+                url: file.url,
+                size: file.size,
             });
-        }
-        else if(typeof File !== "undefined" && file.buffer instanceof File)
-        {
-            // SyncFileBuffer:
-            // - loads the whole disk image into memory, impossible for large files (more than 1GB)
-            // - can later serve get/set operations fast and synchronously
-            // - takes some time for first load, neglectable for small files (up to 100Mb)
-            //
-            // AsyncFileBuffer:
-            // - loads slices of the file asynchronously as requested
-            // - slower get/set
-
-            // Heuristics: If file is larger than or equal to 256M, use AsyncFileBuffer
-            if(file.async === undefined)
-            {
-                file.async = file.buffer.size >= 256 * 1024 * 1024;
-            }
-
-            if(file.async)
-            {
-                var buffer = new v86util.AsyncFileBuffer(file.buffer);
-            }
-            else
-            {
-                var buffer = new v86util.SyncFileBuffer(file.buffer);
-            }
-
-            files_to_load.push({
-                name: name,
-                loadable: buffer,
-            });
-        }
-        else if(file.url)
-        {
-            if(file.async)
-            {
-                let buffer;
-
-                if(file.use_parts)
-                {
-                    buffer = new v86util.AsyncXHRPartfileBuffer(file.url, file.size, file.fixed_chunk_size, false, this.zstd_decompress_worker.bind(this));
-                }
-                else
-                {
-                    buffer = new v86util.AsyncXHRBuffer(file.url, file.size, file.fixed_chunk_size);
-                }
-
-                files_to_load.push({
-                    name: name,
-                    loadable: buffer,
-                });
-            }
-            else
-            {
-                files_to_load.push({
-                    name: name,
-                    url: file.url,
-                    size: file.size,
-                });
-            }
         }
         else
         {
-            dbg_log("Ignored file: url=" + file.url + " buffer=" + file.buffer);
+            files_to_load.push({
+                name,
+                loadable: v86util.buffer_from_object(file),
+            });
         }
     };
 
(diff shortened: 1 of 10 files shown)


JoeOsborn:
The first failing test is an openbsd floppy test that tests fdb behavior. Arguably the test should never have passed since fdb was never used in floppycontroller.

copy:
Thanks! This works pretty well so far. I've left a few comments above, but in general there's nothing major in the way of merging.

A few notes:
- A couple of tests fail with "Unimplemented floppy command call" (Windows XP, React OS). I'm going to improve the command parser a bit and turn this into a warning (the commands executed by these OSes don't always make sense)
- Currently the floppy disk is only really tested for OSes that boot from it (plus the one test for MS-DOS hard drive + floppy). It would be nice to survey which existing OSes support the current implementation, and make sure this PR doesn't break any of them. I can take care of that.
- A test would be nice that demonstrates ejecting/inserting (i.e., boot MS-DOS, run `dir A:`, insert, run `dir A:`, eject, run `dir A:`)

Few stylistic things:
1. `{` after function/if/else should go on its own line
2. Use `===` and `!==` over `==` and `!=`

src/floppy.js
```diff
@@ -43,60 +36,23 @@ function FloppyController(cpu, fda_image, fdb_image)
...
-        sectors_per_track = floppy_type.sectors;
-        number_of_heads = floppy_type.heads;
-        number_of_cylinders = floppy_type.tracks;
-
-        this.sectors_per_track = sectors_per_track;
-        this.number_of_heads = number_of_heads;
-        this.number_of_cylinders = number_of_cylinders;
+        this.floppy_type = this.insert_fda(fda_image);
     }
+    this.dir = 0x0;
```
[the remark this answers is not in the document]
JoeOsborn: That’s worth a try—it was previously initialized to 0x00 so I just forced it there.

src/floppy.js
```diff
@@ -317,39 +330,67 @@ FloppyController.prototype.port3F2_write = function(value)
 FloppyController.prototype.check_drive_status = function(args)
 {
     dbg_log("check drive status", LOG_FLOPPY);
+    // do nothing if no fda
+    if (this.fda_image) {
+        this.status_reg1 = 0;
+    } else {
+        // TODO: is this right?
+        this.status_reg1 = (1 << 2) | 1;
```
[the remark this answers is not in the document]
JoeOsborn: Yes, will do!

src/floppy.js
```diff
@@ -409,6 +452,12 @@ FloppyController.prototype.done = function(args, cylinder, head, sector, error)
         }
     }
 
+    // clear eject flag if seek or write has taken us to a new cylinder
+    if(cylinder != this.last_cylinder) {
+        this.dir = 0x80;
```
JoeOsborn, line 457: This is the other location @copy mentioned (just adding a comment here so I can track it more easily).

JoeOsborn:
Thanks for the review. I’ll address the style issues and wrong flag sets, and try to add an automated test using eject and insert.

JoeOsborn:
I ended up adding an “extra_images” config option (and field on V86Starter) so that the starter can download/initialize all the floppies or CDs or whatever that will be necessary. This way it’s easy to code up something to swap disks by name (as I do in the integration tests).

src/browser/starter.js
```diff
@@ -39,6 +39,11 @@
  * - `filesystem Object` (No 9p filesystem) - A 9p filesystem, see
  *   [filesystem.md](filesystem.md).
  *
+ * - `extra_images Object` (No extra images) - An object whose keys
+     are extra image names and whose values are images to load (see
+     below).  Names used by other images described above are
+     disallowed.
```
copy, line 45: I'm not too fond of this new parameter. It adds more things to already overly complicated constructor, and doesn't serve much use besides the test (or am I missing something?).

I'd suggest removing it, and in the test reading the file with `fs.readFileSync` and wrapping it in `v86util.SyncBuffer` (after all it's only 1.4 MB). For larger files (cdroms, eventually), we can use `v86util.AsyncXHRBuffer`.

Eventually, I plan to refactor the buffer management, so that `insert_fda` takes the same object as the `fda` property in the constructor, but let's leave that out of this PR.

JoeOsborn: Ok, that makes sense. I was hoping to use it to e.g. set up v86 with a full suite of install floppies for some software, but I can find a way to load up images in the test suite on the fly and/or implement the idea of loading a buffer description in insert_fda, or at least introducing an async utility function on starter that takes a buffer desc and yields a buffer.

JoeOsborn:
I've added an async function on V86 to load an image file description as an image description, which I use from `run.js`.  In the future, this could be used to refactor `continue_init` (( `settings["hda"] = this.disk_images["hda"] = await this.load_images(this.disk_images["hda"]);` )) and allow for loading all the image files in parallel (using e.g. Promise.all).

JoeOsborn:
I went ahead and slightly refactored continue_init in the way I described, just so we have only one code path for loading image files.  I left the put_on_settings function intact and did not use Promise.all since I thought it would complicate the progress bar-related code even more.

JoeOsborn:
Sorry for the spam--if you don't want the last commit, feel free to just use the preceding one and keep the code duplication.  I thought adding the async function to load an image would be a smoother, nicer change than exposing v86util or the buffer types as exports of the library, since they kind of feel like implementation details.

copy:
I've refactored the tests and buffer loading. Also improved the command parser and implemented port 3F4 reset (for ReactOS and winxp, previously this code was never hit because we didn't run it without a floppy disk).

Will merge when tests pass.

copy:
Also pushed another change to expose it in the UI.

copy:
Thanks!

Issue #900, opened 2023-08-17
CD-ROM insert, eject


JoeOsborn:
I think the IDEInterface would need to change in a few ways (and we should always have a CD ROM drive even with no cdrom buffer)...  Here's my understanding of the problem based on the osdev wiki ( https://wiki.osdev.org/PCI_IDE_Controller#Ejecting_an_ATAPI_Drive ) and https://inst.eecs.berkeley.edu/~cs150/fa01/labs/project/ATAPI_Spec.pdf .

An empty drive should report a media type of 0x70.

When a disk is inserted, the ide controller should generate a unit attention condition that persists until it's cleared by the guest when it sends a "request sense" about it; my understanding is that this should cause all reads and stuff to error out with a "unit attention" error.  At that point the controller can do what it was doing before, but with a new buffer.  I think this is the second thing to implement: a function to load a CDROM into an empty cdrom controller.

For the emulator user to trigger an eject, we can do it by telling the interface to respond to queries with an error status with bit 3 (Media Change Requested) set.  Then the controller should wait for the guest to send an Acknowledge Media Change command (code DB), after which the guest promises to send the eject sequence.  (For a less graceful way, we could just swap the buffers and set that unit attention condition from before, but then the guest OS doesn't get a chance to e.g. prevent the eject or things like that).

For the guest OS to trigger an eject, I think there is an ATAPI command that can be implemented (see the osdev link), but it's also possible it's the start/stop unit command with the "LoEj" flag set.  Here I'd want to hook on ATAPI commands and see if the guest is sending something that we're not handling.

_Originally posted by @JoeOsborn in https://github.com/copy/v86/issues/54#issuecomment-1675288579_

JoeOsborn:
So my first roadblock is that I can't boot from hda with a CD inserted.

I don't know much about PCI or IDE or ATA, but here's what I tried so far:

1. I made the windows31 profile also load the windows30 CD in cdrom, adjusting the boot order to 0x132.
2. I changed the log statement in src/ide.js:629 to include this.device.name
3. I saw a number of ATA commands issued to ide0 and ide1; only 0xA1 was ever issued to ide1.
4. I hit a debug assert on a spurious dma write onto ide1 (the CD)

I'm really suspicious that this is because src/ide.js:49 hardcodes the master port to 0xB400 and registers some writes and reads with the CPU, so whichever controller is initialized second is the one that gets DMA stuff done to it.  What's the correct thing to do?

JoeOsborn:
[...]

PR #901, opened 2023-08-18
Fix booting from cd while hda present, booting from hda while cd present

JoeOsborn:
This is towards #900: We can now boot either from HDA or CDROM, with the other present, and it boots successfully.  For testing I'm booting into Windows 3.0 (from CD) with Windows 3.1 on hda or from Windows 3.1 (on hda) with Windows 3.0 in the cdrom.

One big change here: The cd-rom device is now always present, either at ide0 or ide1 depending on whether hda/hdb are present.  Of course we can still boot (at least into Windows) if you only have hda and an empty cdrom, or don't have hda and just have a cdrom.


JoeOsborn:
The Bochs BIOS does not like the empty cd-rom drive.  I'll investigate how it should properly report being empty.  If I keep the "empty drive" status register readout to 0, seaBIOS ignores the drive but bochs BIOS hangs.

JoeOsborn:
It looks like the ata command sending code in bochs doesn’t like how our buffer-less cdrom drive is reporting its state register, so it waits forever before sending the IDENTIFY PACKET DEVICE ATA command.  I’ll investigate what’s going on Monday or Tuesday but my suspicion is that the current approach of hard coding one value to send for empty drives is not going work.

JoeOsborn:
Sorry for the spam.  I think the path of least resistance here is to add a setting for “wants_cdrom” which defaults to true if cdrom is non-null, and a cdrom device should only be created if cdrom is defined or wants_cdrom is true.  The bochs empty cdrom boot behavior really should be a separate PR aimed at #900 .

JoeOsborn:
Whoops, and I’ll revert my changes to the html files and main.js.

JoeOsborn:
I’ve trimmed down the changes made by this commit (I left the changes to logging because they’re actually pretty useful) to just changing the BAR4 location for each IDE device and not asserting on a spurious dma control write that occurs before a dma start command arrives (this is something some OSes or maybe seabios seem to do but it doesn’t seem to cause any harm to allow it, I guess they’ll try to trigger it again after it doesn’t do anything the first time—I wonder if a better fix would be for the start dma commands to start the dma transfer immediately if the control start bit was set already).  I’ll do some more testing locally (I figured out how to run the integration tests locally and added test machines with both hda and cdrom and booting from one or the other) and I’ll push a commit over the next day or two.

JoeOsborn:
The main changes necessary to make this work were, in the end:
1. Change the debug assert to an error condition in the "spurious DMA write" case
2. Change the BAR4 address for each IDE controller.

The wants_cdrom change is not strictly necessary here but will be important for my next PR which will support empty CD-ROMs, so please consider rolling it in to this.

JoeOsborn:
I've gone ahead and added tests for booting with wants_cd true when Bochs bios is active (and inactive) and with varying permutations of hda, fda, and boot order in FreeDOS. It now boots successfully with an empty CD-ROM in both BIOSes.

The problem is that Bochs was enumerating the slave device of every IDE device, but the slave device was seemingly never fully implemented (?)---so instead I have dummied out writes to slave devices when the slave buffer is null.

JoeOsborn:
With these changes, CDs are also readable by at least FreeDOS 13 with the UDVD2 driver.  They don’t seem to be readable in win95.  I’m going to see if I can implement inserting and ejecting disks later this week.

JoeOsborn:
I've rebased this branch on latest master and removed unnecessary changes.  eject is still in progress, but insert seems to work (I can boot FreeDOS 1.3 with an empty CD-ROM, see `dir D:` fail, insert a CD-ROM, and see `dir D:` succeed).

copy:
@JoeOsborn Thanks! I will try to look into this in the coming days.

JoeOsborn:
I see my api-test is hanging, it was tricky to get the timing of abort/retry/fail and inserting the CD right.  It works on my machine but I’m open to advice on getting it working in CI.

JoeOsborn:
Sorry @copy , I don’t want to use up your CI credits or whatever, but I don’t know how to kill the actions jobs from my end.

JoeOsborn:
After more testing, I see that ejecting a CD-ROM and inserting a new one also seems to work.

src/browser/main.js
```diff
@@ -1369,6 +1372,7 @@
             "hda": settings.hda,
             "hdb": settings.hdb,
             "cdrom": settings.cdrom,
+            "wants_cdrom": settings.wants_cdrom,
```
copy, line 1375: I think `cdrom: { ejected: true }` would be a clearer api for this.

index.html
```diff
@@ -80,6 +80,13 @@ <h4>Setup</h4>
                 </td>
             </tr>
 
+            <tr>
+                <td><label for="wants_cdrom">Allow empty cd-rom drive</label></td>
+                <td>
+                    <input id="wants_cdrom" type="checkbox"><br>
+                </td>
+            </tr>
```
copy, line 88: Could you revert this for now? It's confusingly named, can cause problems with the state saving mechanism, and I'd like to eventually redo the UI anyway.

copy:
This causes some problems on Linux:

There are some 0-bytes drives:
```
[   32.946240] sd 1:0:1:0: [sdc] 0 512-byte logical blocks: (0 B/0 B)           
[   32.946845] sd 1:0:1:0: [sdc] Write Protect is off                           
[   32.951335] sd 1:0:1:0: Attached scsi generic sg3 type 0                     
[   32.956020] sd 1:0:1:0: [sdc] Write cache: disabled, read cache: enabled, doe
sn't support DPO or FUA                                                         
[   32.970920] sd 1:0:1:0: [sdc] Attached SCSI disk                             
```

There are some timeouts, e.g. running `fdisk /dev/sr0`:

```
[  176.750270] ata2: lost interrupt (Status 0x58)
[  176.750920] ata2: drained 32 bytes to clear DRQ
[  176.751025] ata2.00: limiting speed to MWDMA1:PIO2
[  176.751340] ata2.00: exception Emask 0x0 SAct 0x0 SErr 0x0 action 0x6 frozen
[  176.751910] ata2.00: cmd a0/01:00:00:20:00/00:00:00:00:00/a0 tag 0 dma 16416 in
[  176.751910]          Get configuration 46 00 00 00 00 00 00 00 20 00res 40/00:02:00:0c:00/00:00:00:00:00/a0 Emask 0x4 (timeout)
[  176.752900] ata2.00: status: { DRDY }
[  176.753640] ata2: soft resetting link
[  176.914875] ata2.00: configured for MWDMA1
[  176.915840] ata2.01: configured for MWDMA2
[  176.916755] ata2: EH complete
[  184.750285] ata2: lost interrupt (Status 0x58)
[  184.751185] ata2: drained 16 bytes to clear DRQ
[  184.751285] ata2.00: limiting speed to PIO2
[  184.751640] ata2.00: exception Emask 0x0 SAct 0x0 SErr 0x0 action 0x6 frozen
[  184.752215] ata2.00: cmd a0/01:00:00:10:00/00:00:00:00:00/a0 tag 0 dma 16400 in
[  184.752215]          Get configuration 46 00 00 28 00 00 00 00 10 00res 40/00:02:00:0c:00/00:00:00:00:00/a0 Emask 0x4 (timeout)
[  184.753650] ata2.00: status: { DRDY }
[  184.755615] ata2: soft resetting link
[  184.914770] ata2.00: configured for PIO2
[  184.915615] ata2.01: configured for MWDMA2
[  184.916485] ata2: EH complete
```

You should be able to reproduce with [tinycore](http://www.tinycorelinux.net/11.x/x86/archive/11.0/TinyCore-11.0.iso) (it can be used both as cdrom and hard drive disk).

src/browser/main.js
JoeOsborn: I’m happy to defer to you here, I’ll make that change.

index.html
JoeOsborn: Ok!

JoeOsborn:
To clarify on the linux issues, is this the set of test cases you’re using?
1. Boot from tinycore as hda, insert tinycore cd, mount it
2. Boot from tinycore as hda with some cd image, run fdisk on the cd? This one’s a bit confusing to me, is it just that the error condition isn’t being handled properly?  Or should something like CD-R support be implemented?

copy:
@JoeOsborn Sorry, here is a better reproduction:

Boot with [linux4.iso](https://k.copy.sh/linux4.iso) as cdrom and tinycore (or anything else) as hard drive, set boot order to CD first. `fdisk /dev/sr0` hangs, but I believe the same hang also happens during boot on tinycore. It looks like an interrupt missing somewhere.

src/browser/main.js
JoeOsborn: I think `cdrom: { ejected: true }` makes sense for the starter options, but it's a bit tricky inside of cpu and so on if the `settings` object is either a buffer or null or `{ejected: true}`.  Should there be a new `EmptyBuffer` buffer type?  Or can `wants_cdrom` continue to exist within the emulator settings, even if it's not part of public API?

copy: Yes, internally `wants_cdrom` is fine (it can be refactored at any point anyway)

JoeOsborn:
Something I’m finding confusing here is that linux seems to be configuring the drive to use DMA even though the capabilities reported are lba-only.  

I think maybe the OS is waiting for some status flag to be set or something, or else it’s ignoring irqs that come in too soon after the command is sent, since liberally adding irqs didn’t seem to help.  Any tips for triggering an IRQ after some delay?

copy:
>  Any tips for triggering an IRQ after some delay?

You can try `setTimeout(() => this.push_irq(), 4)` or similar, but in my experience that isn't usually needed and is somewhat of an anti-pattern.

JoeOsborn:
Just posting to say I'm still investigating this, but the beginning of the semester has slowed down my experiments here.  Once I've settled into teaching a little more I'll have another chance to get back to it.

ericmackrodt:
Was this work abandoned? I think this is essential for the emulator and I'm working on a project that requires this functionality.
I tried to build the version from this branch but it seems like it's broken. I'm getting the following errors:
Screenshot 2023-12-01 at 7 07 09 pm

Sadly I don't think I have the skills to fix this.

JoeOsborn:
It’s not abandoned, but it is delayed.  I likely won’t be able to pick it up until January—it is on my critical path for some preservation work, but other aspects of the project (and my other obligations) have taken the front seat for a while.

ericmackrodt:
That's fair enough, thank you for getting back to me.

ericmackrodt:
Hello,

I feel this PR is now on way outdated code.
I wonder how we could get this issue fixed, I think it's a very important fix.

FunnyCorgi:
Hey, does anyone know if this is still being actively worked on?

copy:
> Does anyone know if this is still being actively worked on?

Probably nobody, but feel free to pick it up.

toncho11:
I see that there are often two drives : C and D that are identical. 
Second should be the CD but it the HDD image as well. I have seen this several times already.
In both DOS and Windows instead of CD you get the a second HDD image.

chschnell:
I've manually merged all essential changes from 2023 to a fresh branch from the current master, see branch [`chschnell/v86:fix-cdrom-hda-boot`](https://github.com/chschnell/v86/tree/fix-cdrom-hda-boot).

Changes and modifications in respect to the original changeset:

* For now I left out almost all of the `dbg_log()` changes, they're like 90% of the original PR and make the essential changes in the diffs a bit hard to read, these can be added back in at a later point in time.
* I've also left out the `wants_cdrom` option and implemented its desired effect with `cdrom: { ejected: true }` as was suggested. I didn't change v86's web UI, yet.
* And I've left out the test `tests/api/cdrom-insert-eject.js`, also to be added later.
* While reviewing the code I left three TODO comments in `IDEInterface.prototype.set_cdrom()`.

This branch should behave like the last changeset from @JoeOsborn in 2023, it is installed and can be tested at [my V86 demo](https://www.rekoba.de/~cs/v86/demo/) site (example direct link: [FreeDOS 1.3](https://www.rekoba.de/~cs/v86/demo/?p=hd_freedos_20m_en)).

I need to do a lot more testing, debugging, and learning, but so far it seems to work like expected (as described in the comments above). I still have quite a learning curve ahead of me in respect to PCI and IDE, ATA and ATAPI etc. (any help or pointers would be highly appreciated).

I'm using SeaBIOS 1.16.2 for my tests (the BIOS does make a difference). I've only tested FreeDOS 1.3 and Debian 12 guests so far. Issues:

**Problem 1:**

When booting under current Debian Linux with an empty CD-ROM device I also get these boot log messages (I can't even interpret that, yet):

```
SeaBIOS (version rel-1.16.2-0-gea1b7a0)
[    9.000923] ata_piix 0000:00:1f.0: can't enable device: BAR 0 [io  size 0x0008] not assigned
[   48.433000] ata2.00: exception Emask 0x0 SAct 0x0 SErr 0x0 action 0x6 frozen
[   48.433923] ata2.00: cmd a0/01:00:00:fe:00/00:00:00:00:00/a0 tag 0 dma 16640 in
[   48.433923]          Inquiry 12 00 00 00 fe 00res 40/00:03:00:08:00/00:00:00:00:00/a0 Emask 0x4 (timeout)
[   48.434923] ata2.00: status: { DRDY }
[   55.727923] ata2.00: exception Emask 0x0 SAct 0x0 SErr 0x0 action 0x6 frozen
[   55.727923] ata2.00: cmd a0/01:00:00:fe:00/00:00:00:00:00/a0 tag 0 dma 16640 in
[   55.727923]          Inquiry 12 01 00 00 fe 00res 40/00:03:00:08:00/00:00:00:00:00/a0 Emask 0x4 (timeout)
[   55.728000] ata2.00: status: { DRDY }
```

**Problem 2:**

I can boot FreeDOS 1.3 from HDA with empty or non-empty CD-ROM device, and the device gets properly assigned to drive letter `D:`. I get a proper `Drive not ready` error when accessing `D:` with ejected CD.

When booting with an empty CD-ROM device I can mount a CD disk image at guest runtime, but when I then unmount it again it stays ready in FreeDOS until I access a path on it that I didn't access before (uncached), it then fails with `File not found` errors but continues to print out the disk's volume label as though the disk was still inserted. Seems like FreeDOS somehow doesn't get properly informed about the ejected CD-ROM.

I think we could start by collecting the bugs that remain with this patch in its current state. I would like to get this to fully work, any help - as always - would be much appreciated.

chschnell:
@copy: A very basic question, which [ATA/ATAPI specification](https://en.wikipedia.org/wiki/Parallel_ATA#Features_introduced_with_each_ATA_revision) is targeted in ide.js. Is it ATA/ATAPI-6?

copy:
> @copy: A very basic question, which [ATA/ATAPI specification](https://en.wikipedia.org/wiki/Parallel_ATA#Features_introduced_with_each_ATA_revision) is targeted in ide.js. Is it ATA/ATAPI-6?

Yes, feel free to use that one.

Frankly, the code is a mess. It's partially written against various specs and partially against random driver code of various OSes I was trying to get to boot across the years. The good news is that the OS tests are [extensive](https://github.com/copy/v86/blob/a3840e32b30d7005f2e7533a557b769333127425/tests/full/run.js#L80).

> I need to do a lot more testing, debugging, and learning, but so far it seems to work like expected (as described in the comments above). I still have quite a learning curve ahead of me in respect to PCI and IDE, ATA and ATAPI etc. (any help or pointers would be highly appreciated).

Highly appreciate your work on this.

I will do some testing with your branch. Feel free to create a PR if you think it's ready for us to have a look at the code.

chschnell:
> Frankly, the code is a mess.

Haha, that's fine.

> It's partially written against various specs and partially against random driver code of various OSes I was trying to get to boot across the years. The good news is that the OS tests are [extensive](https://github.com/copy/v86/blob/a3840e32b30d7005f2e7533a557b769333127425/tests/full/run.js#L80).

I hope that we can straigthen some of it out.

copy:
Continued in #1325

Issue #877, opened 2023-06-12
Idle for Windows 95 and lower


copy:
It would be good to document a way to load an idle driver for Windows 95 and MS-DOS, as they don't use `hlt`.

Alternatively, v86 could detect the idle loops of these OSes, and handle them specially.

For FreeDOS, adding `IDLEHALT=1` to config.sys works.

copy:
I've installed amnhlt and dosidle in Windows 95 and MS-DOS respectively.

PR #907, opened 2023-09-01
docs: adding instructions about IDLE driver for DOS and Windows 9x (#877)

SuperMaxusa:


copy:
Thanks!

volkertb:
@SuperMaxusa @copy Under the section `## Windows 9x (using AmnHLT)`, you might also want to add that AmnHLT is redundant when ACPI is enabled in Windows 9x, and that you can force ACPI to be enabled in the Windows 98 installer by starting it with `setup /p j`. This is needed for instance when you're installing Windows 98 in a VirtualBox VM. See also https://forums.virtualbox.org/viewtopic.php?f=6&t=32989

SuperMaxusa:
@volkertb Thanks for the advice, I read somewhere that Windows 98 (FE or SE) can support HLT, but I didn't find any specific information about it.

PR #911, opened 2023-09-07
Use localstorage to save options

BelleNottelling:
This small change just incorporates the usage of the browser's local storage to remember and restore a handful of options for a minor increase to the overall quality of life.
Namely it saves and restores:
1. System memory size 
2. Video memory size
3. The networking proxy that's configured (saving this will be rather nice if testing a local one)
4. The boot order
5. If audio is enabled
6. If ACPI is enabled

Worst case, this doesn't save info in a way that someone finds helpful and they have to change the value as they would have normally. Best case, people re-use the config for a bunch of different testing and this change saves them a bit of clicks / keystrokes.

### DIFF
--- a/src/browser/main.js
+++ b/src/browser/main.js
@@ -101,6 +101,29 @@
 
     function onload()
     {
+        // Load previously entered options via the local storage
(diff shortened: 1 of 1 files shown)


copy:
Good idea, thanks!

BelleNottelling:
Cheers!

PR #913, opened 2023-09-10
update readme.md to specify sse3 capability (see pr#906)

spetterman66:


BelleNottelling:
Why not add the note in your already existing and still unmerged PR? This doesn't need to be an entirely separate PR

Issue #916, opened 2023-09-13
Keyboard event target is too broad


JoeOsborn:
Currently keyboard event listeners are attached to the window instead of the canvas or container div, which to me seem like more natural choices.  Would it be okay to change the target element or at least make it configurable?

My use case is that I’d like to embed multiple v86 instances on one page without repeatedly disabling/enabling event handling on each one, which seems like it’s just duplicating the browser’s machinery.

PR #918, opened 2023-09-14
Target keyboard events to screencontainer instead of window

JoeOsborn:
This is in reference to #916; I think it's reasonable for keyboard events to go to the screen container, in case you don't want v86 consuming all keyboard actions anywhere on the page.  For cases where you have multiple v86 instances or want keyboard navigation outside of the emulator this is helpful.  (Mouse events already filter based on mouse position so those are fine.)

I decided just to change the keyboard event registration instead of making it configurable because the current behavior feels wrong to me personally, but if @copy or others feel strongly one can always add another config option or alter the "disable keyboard" option to be something like a `keyboard_target` HTMLElement (and if null, act as if the keyboard is disabled).

### DIFF
--- a/src/browser/keyboard.js
+++ b/src/browser/keyboard.js
@@ -11,8 +11,13 @@ var SCAN_CODE_RELEASE = 0x80;
  *
  * @param {BusConnector} bus
  */
-function KeyboardAdapter(bus)
+function KeyboardAdapter(bus, screen_container)
 {
+    // Make screen container focusable
+    if(screen_container.tabIndex === undefined)
+    {
+        screen_container.tabIndex = 0;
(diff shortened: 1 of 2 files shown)


copy:
I disagree with this change. It makes it more likely to be in a situation where keyboard inputs are unexpectedly not being sent to the VM.

For the embedded or multi-VM case, there is an example that shows how to route keyboard inputs only when the screen is focused: https://github.com/copy/v86/blob/master/examples/two_instances.html

JoeOsborn:
Alright, I understand.  It still feels duplicative of browser machinery to me, but I can make my use case work with `keyboard_set_status`.

PR #919, opened 2023-09-15
More graceful HTTP content negotiation

JoeOsborn:
When v86 needs to load a chunk of a file, it can silently download garbage if the server is using gzip encoding (it gets a range, but of the wrong data).  This patch does two things:

1. Error out instead of proceeding if the content-encoding of a download after a range request exists and is not "identity".
2. Send the nonstandard "X-Accept-Encoding: identity" header when making range requests.

We can't use `Accept-Encoding` because it's on the forbidden header list, but some web servers respect `X-Accept-Encoding`.

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -578,6 +578,7 @@ function load_file(filename, options, n_tries)
         let start = options.range.start;
         let end = start + options.range.length - 1;
         http.setRequestHeader("Range", "bytes=" + start + "-" + end);
+        http.setRequestHeader("X-Accept-Encoding", "identity");
 
         // Abort if server responds with complete file in response to range
         // request, to prevent downloading large files from broken http servers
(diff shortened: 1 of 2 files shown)


src/lib.js
```diff
@@ -604,6 +605,16 @@ function load_file(filename, options, n_tries)
             }
             else if(http.response)
             {
+                if(options.range)
+                {
+                    let enc = http.getResponseHeader("Content-Encoding");
+                    if(enc && enc !== "identity")
+                    {
+                        console.error("Range requested from compressed resource " + filename, enc);
+                        http.abort();
```
copy, line 614: This breaks if at one point in the future there is a new Content-Encoding that is compatible with range requests. I'd suggest removing the `http.abort()`, and just leave the warning.

It would be nice to improve the warning message, e.g.:

`console.warn("Server sent Content-Encoding in response to ranged request", { filename, enc });`

copy:
Cheers!

PR #926, opened 2023-09-19
Update Readme.md to specify operating system compatibility in Virtual x86

spetterman66:
Read the title to see what this PR is about :)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -26,7 +26,7 @@ list of emulated hardware:
 - A PCI bus. This one is partly incomplete and not used by every device.
 - An IDE disk controller.
 - An NE2000 (RTL8390) PCI network card.
-- A virtio filesystem.
+- A VirtIO filesystem.
 - A SoundBlaster 16 sound card.
 
 ## Demos
@@ -68,12 +68,7 @@ Here's an overview of the operating systems supported in v86:
 
 - Linux works pretty well. 64-bit kernels are not supported.
   - Damn Small Linux (2.4.31 kernel) works.
-  - KNOPPIX 9.1 works.
-  - Elementary OS 0.3.2 works.
-  - Slax works.
-  - AntiX works
   - Fedora 30 works.
-  - Zorin OS 12.6 works.
   - All tested versions of TinyCore work.
   - [Buildroot](https://buildroot.uclibc.org) can be used to build a minimal image.
     [humphd/browser-vm](https://github.com/humphd/browser-vm) and
--- a/Readme.md
+++ b/Readme.md
@@ -67,7 +67,13 @@ list of emulated hardware:
 Here's an overview of the operating systems supported in v86:
 
 - Linux works pretty well. 64-bit kernels are not supported.
-  - Damn Small Linux (2.4 Kernel) works.
+  - Damn Small Linux (2.4.31 kernel) works.
+  - KNOPPIX 9.1 works.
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -67,7 +67,13 @@ list of emulated hardware:
...
 
 - Linux works pretty well. 64-bit kernels are not supported.
-  - Damn Small Linux (2.4 Kernel) works.
+  - Damn Small Linux (2.4.31 kernel) works.
+  - KNOPPIX 9.1 works.
+  - Elementary OS 0.3.2 works.
+  - Slax works.
+  - AntiX works
+  - Fedora 30 works.
+  - Zorin OS 12.6 works.
```
copy, line 76: We don't really need to name every Linux distribution.  How about just Fedora, the is based on Debian or Ubuntu anyway.

spetterman66:
yeah, good reason
________________________________

spetterman66:
better?

copy:
Looks good now. That said, does Fedora 30 *actually* work?

spetterman66:
yes. see the screenshots i posted in the gitter discussion channel

copy:
I can't reproduce it, does only the LXDE version work?

spetterman66:
what specifications do you allocate to the virtual machine? put it on 2048 mb ram, 128 mb vram, and acpi off (not needed). boot arguments loglevel=7 earlyprintk=serial,keep for best verbosity and to get info where it gets stuck at

spetterman66:
As you can see in the screenshot below, Fedora 30 (i386, LXDE) works fine in Virtual x86. I have not tried other spins.
image

copy:
Does it work correctly? I wouldn't expect any distro (not even bloated ones) to take more than a few minutes to boot.

BelleNottelling:
> Does it work correctly? I wouldn't expect any distro (not even bloated ones) to take more than a few minutes to boot.

It looks like Fedora 30 just outright works (although **very** slowly unless you are using one of the optional desktop environments.
- `Fedora-Workstation-Live-i386-30-1.2.iso` Booted to the live image about 8.5 min, but once it got to that point performance was awful and it's not even worth attempting, moving the mouse would be met with 30+ second long delays.
- `Fedora-LXDE-Live-i386-30-1.2.iso` Got me to a complete LiveCD desktop in about 7 min and felt very usable in performance.
- `Fedora-LXQt-Live-i386-30-1.2.iso` Took about 6min to get the LiveCD to the desktop and once again felt very usable.
- `Fedora-Xfce-Live-i386-30-1.2.iso` Took about 10 min for the LiveCD to get to the desktop and again performance felt fine.

These were tested with `2048MB` of RAM and `64MB` of VRAM allocated in the VM. ACPI was disabled and in all cases I used the default live image option that I was presented with.

I'm guessing the *very* long boot time in that other screenshot was either because of the boot arguments that were used or potentially a slow host was being used. I am testing things with an `i7 13700K` so it's somewhat a best-case scenario for CPU performance.

copy:
> Fedora-Workstation-Live-i386-30-1.2.iso Booted to the live image about 8.5 min, but once it got to that point performance was awful and it's not even worth attempting, moving the mouse would be met with 30+ second long delays.

Hmm, indeed. It's similarly slow in qemu-kvm, so let's assume v86 isn't at fault here.

Issue #927, opened 2023-09-20
Multi-byte UTF-8 characters not displaying correctly on xterm.js


robertaboukhalil:
When using xterm.js with v86, UTF-8 characters that are defined by multiple bytes are not displayed correctly.

**Reproduce:**
* Go to https://copy.sh/v86/?profile=archlinux
* Type `echo -e '\x24'` (single-byte) and that [correctly](https://en.wikipedia.org/wiki/UTF-8#Encoding_process) shows the `$`:
![Screenshot 2023-09-19 at 8 19 41 PM](https://github.com/copy/v86/assets/2009870/43ebaf5e-acba-441f-8d55-704d43e4ac19)

* Type `echo -e '\xc2\xa3'` (two bytes) and that shows the correct symbol in the original v86 renderer but not xterm.js:
![Screenshot 2023-09-19 at 8 20 49 PM](https://github.com/copy/v86/assets/2009870/a62e7410-8037-4b53-9c66-0330132ddf19)

I noticed this issue when using the linux `tree` command, where I see `â` characters instead of a horizontal line, but **only with xterm.js**:
![Screenshot 2023-09-19 at 8 24 11 PM](https://github.com/copy/v86/assets/2009870/022c1a5a-51be-466e-b8a9-d9b83217bbae)

I wasn't able to find the root cause but I **think** it has something to do with the UART logic sending [treating each byte separately](https://github.com/copy/v86/blob/23a4f4c772adea7ead30faa900aa39fc729c9a01/src/uart.js#L345), which then gets output one at a time in the [xterm.js serial adapter](https://github.com/copy/v86/blob/master/src/browser/serial.js#L240). Whereas with the regular serial adapter, the bytes are [appended before being rendered](https://github.com/copy/v86/blob/master/src/browser/serial.js#L60)?

SuperMaxusa:
Judging by the debug log, these two bytes seem to be sent correctly.
```08:55:45+890 [SERI] interrupt enable: 0x7
08:55:45+891 [SERI] read interrupt identification: 0x2
08:55:45+892 [SERI] read line status: 0x60
08:55:45+892 [SERI] read modem status: 0x0
08:55:45+893 [SERI] data: 0xC2 
08:55:45+895 [SERI] data: 0xA3 
08:55:45+896 [SERI] interrupt enable: 0x5 
08:55:45+896 [SERI] read interrupt identification: 0x1 
08:55:45+896 [SERI] interrupt enable: 0x7
[...]

PR #928, opened 2023-09-20
Replace serial-output-char with serial-output-byte

robertaboukhalil:
Fixes https://github.com/copy/v86/issues/927


copy:
Cheers!

PR #929, opened 2023-09-21
Performance improvements

spetterman66:
Would this speed up v86 with the release build? If so, merge this :D


Makefile
```diff
@@ -62,7 +62,10 @@ CLOSURE_FLAGS=\
 		--jscomp_error unknownDefines\
 		--jscomp_error visibility\
 		--use_types_for_optimization\
+		--compilation_level ADVANCED_OPTIMIZATIONS\
```
copy, line 65: This will likely break the library version of v86.

Makefile
```diff
@@ -172,9 +175,10 @@ src/rust/gen/analyzer0f.rs: $(ANALYZER_DEPENDENCIES)
 build/v86.wasm: $(RUST_FILES) build/softfloat.o build/zstddeclib.o Cargo.toml
 	mkdir -p build/
 	-BLOCK_SIZE=K ls -l build/v86.wasm
+	RUSTFLAGS="--emit asm"
 	cargo rustc --release $(CARGO_FLAGS)
 	cp build/wasm32-unknown-unknown/release/v86.wasm build/v86.wasm
-	-$(WASM_OPT) && wasm-opt -O2 --strip-debug build/v86.wasm -o build/v86.wasm
+	-$(WASM_OPT) && wasm-opt -O4 --strip-debug build/v86.wasm -o build/v86.wasm
```
copy, line 181: Also check that this doesn't increase the size of v86.wasm too much.

spetterman66: it's for performance, not size, and yes my internet connection can easily handle it

spetterman66:
compare https://github.com/copy/v86/actions/runs/6277121788/job/17048249482 to https://github.com/copy/v86/actions/runs/6256963615/job/16988589524 :D

spetterman66:
If you compare those two GitHub Actions, then there's a large performance improvement with Tiny Core 11 HD, and Tiny Core 11 CD.

.github/workflows/ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Install APT packages
         run: |
           sudo apt-get update -y
-          sudo apt-get install nasm gdb qemu-system-x86 libc6-dev-i386 -y
+          sudo apt-get install nasm gdb qemu-system-x86 libc6-dev-i386 binaryen -y
```
copy, line 34: Does this make any difference? It's not event used. 

Could you please stop trying random things?

copy:
> If you compare those two GitHub Actions, then there's a large performance improvement with Tiny Core 11 HD, and Tiny Core 11 CD.

Tests use v86-debug.wasm, and this PR didn't make any changes to that file (please don't send a PR that "optimises" v86-debug.wasm, it's for tests with extra assertions and debug information).

CI isn't stable enough for benching. Bench locally.

copy:
I've tested a few of these changes:

- `assume_function_wrapper` reduces the size of libv86.js by 50kB, so that's a nice improvement, will cherry-pick
- `rewrite_polyfills=false` reduces the size of libv86.js and v86_all.js by about 1kB by removing some polyfills. `language_in=ECMASCRIPT_2020` `language_out=ECMASCRIPT_2020` has the same effect, so I'll use that instead
- `wasm-opt -all` makes no measurable difference
- `codegen-units = 1` makes no measurable difference

spetterman66:
also consider using a different memory allocator designed for rust wasm such as lol_alloc or wee_alloc.

PR #931, opened 2023-09-25
Both SkiffOS and SkiftOS are mentioned in the readme, but it's kinda mixed up.

ChaoticByte:
I stumbled across a minor imperfection in the README: both SkiftOS and SkiffOS are mentioned, but the former has no entry in the compatibility list, and the latter has no profile linked in the demo list (but there apparently is one). This might confuse some users (like me) :D

I quickly fixed this by adding the two lines.


copy:
Cheers!

PR #946, opened 2023-11-22
Fix release artifact upload.

basicer:
Fixes publishing build artifacts into the latest release and also builds and includes the fallback wasm build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,9 @@ jobs:
       - name: Build all
         run: make all
 
+      - name: Build fallback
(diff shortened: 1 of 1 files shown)


basicer:
Also something is going on with the nasm tests on CI where they pass if they use the cache but fail for clean builds.

copy:
Thanks! I'm a bit busy right now, will look at this again in two weeks.

copy:
> Also something is going on with the nasm tests on CI where they pass if they use the cache but fail for clean builds.

If you rebase on master this should be fixed, and then good to merge.

copy:
Thanks!

PR #949, opened 2023-11-27
Implement Virt-io Console device.

basicer:
Implements a virtio console device. This allows reading and writing buffers in bulk as well as a channel to communicate terminal size updates.


src/virtio_console.js
```diff
@@ -0,0 +1,294 @@
...
+    this.Send(2, bufchain, buf);
+};
+
+VirtioConsole.prototype.Send = function (queue_id, bufchain, blob) {
+    bufchain.set_next_blob(blob);
+    this.virtio.queues[queue_id].push_reply(bufchain);
+    this.virtio.queues[queue_id].flush_replies();
+};
+
+VirtioConsole.prototype.Ack = function (queue_id, bufchain) {
```
copy, line 289: Try to follow the code style the project (newline before `{`)

src/virtio_console.js
```diff
@@ -0,0 +1,294 @@
...
+                        switch(event) {
+                            case VIRTIO_CONSOLE_DEVICE_READY:
+                                for (var i = 0; i < this.ports; ++i) {
+                                    this.SendEvent(i, VIRTIO_CONSOLE_DEVICE_ADD, 0);
+                                }                               
+                                break;
+                            case VIRTIO_CONSOLE_PORT_READY:
+                                this.Ack(queue_id, bufchain);
+                                this.SendEvent(port, VIRTIO_CONSOLE_CONSOLE_PORT, 1);
+                                this.SendName(port, 'virtio-' + port);
```
copy, line 145: Style: Use double quotes.

src/virtio_console.js
```diff
@@ -0,0 +1,294 @@
+
+"use strict";
```
copy, line 2: If you have used some document/reference for the protocol, you should post it here in a comment.

copy (commented):
Thanks! This is useful to check the existing virtio implemention, and eventually could serve as a better console for some of the modern OSes.

I have left a few stylistic remarks, but overall this looks good.

I've tested this on the archlinux profile and the device is detected and seems to work.

If you rebase, the test failure should go away.

What's missing (not necessarily in this PR):

- A test. Unfortunately the small Linux images haven't been configured with virtio-console support, so this will have to wait for later
- Integration into xtermjs (e.g. a `virtio_console_container_xtermjs` parameter, analogous to `serial_container_xtermjs`)
- Some documentation on how to use this, at least on Linux

src/virtio_console.js
basicer: https://docs.oasis-open.org/virtio/virtio/v1.2/csd01/virtio-v1.2-csd01.html#x1-2900003

basicer:
Thanks for the feedback.  I'll get the style cleaned up. For testing I made a custom Alpine Linux based image that weights about 26mb (6mb kernel, 20mb initramfs). Would a test based on that work, or is it still a bit too large?

src/virtio_console.js
```diff
@@ -0,0 +1,294 @@
...
+    /** @type {VirtIO} */
+    this.virtio = new VirtIO(cpu,
+    {
+        name: "virtio-console",
+        pci_id: 0x07 << 3,
+        device_id: 0x1043,
+        subsystem_device_id: 3,
+        common:
+        {
+            initial_port: 0xB800,
```
basicer, line 63: One thing I was unsure about was these values.  I just picked some based on the ones used for p9 and they seemed to work.

copy:
> For testing I made a custom Alpine Linux based image that weights about 26mb (6mb kernel, 20mb initramfs). Would a test based on that work, or is it still a bit too large?

Yes, that would be useful.

src/virtio_console.js
copy: Yes, this will likely create conflicts with the virtio-fs device. Could you change it?

basicer: I had added 0x1000 to them all, I just dont know what that means.  I tested it with both console and p9 up and both seem to work together.

copy:
Thanks! I've added a test in https://github.com/copy/v86/commit/609d79454d24f342f75c54d65d7348bcbde5cdff

PR #967, opened 2023-12-23
Index.html

spetterman66:


copy:
I think the current phrasing is fine :-)

PR #970, opened 2023-12-30
Feature: Add development environment setup

cognitivegears:
This PR adds enhanced support for github codespaces, as well as VS Code remote containers. With this change, a developer can create a new codespace and automatically have access to build and test the code without needing to setup their environment manually.


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,26 @@
...
+			"christian-kohler.npm-intellisense",
+			"dbaeumer.vscode-eslint",
+			"dustypomerleau.rust-syntax",
+			"serayuzgur.crates",
+			"ms-vscode.makefile-tools",
+			"rust-lang.rust-analyzer"
+		]
+	}
+}
+}
```
copy, line 26: Could you properly format this file? (replace tabs with spaces and add a trailing newline)

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt update && \
+    apt-get install -y nodejs nasm gdb unzip p7zip-full openjdk-17-jre wget python3 qemu-system-x86 git-core build-essential libc6-dev-i386-cross libc6-dev-i386 clang curl time
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+RUN export PATH="$HOME/.cargo/bin:$PATH" && \
+    rustup toolchain install stable && \
+    rustup target add wasm32-unknown-unknown && \
+    rustup component add rustfmt-preview
+RUN wget -P images/ https://k.copy.sh/{linux.iso,linux4.iso,buildroot-bzimage.bin,openbsd-floppy.img,kolibri.img,windows101.img,os8.img,freedos722.img}
```
copy, line 11: I don't think you can use `{}` in Dockerfiles:

```
 > [dev_container_auto_added_stage_label 5/7] RUN wget -P images/ https://k.********.sh/{linux.iso,linux4.iso,buildroot-bzimage.bin,openbsd-floppy.img,kolibri.img,windows101.img,os8.img,freedos722.img}:
0.473 --2024-01-02 22:20:11--  https://k.********.sh/%7Blinux.iso,linux4.iso,buildroot-bzimage.bin,openbsd-floppy.img,kolibri.img,windows101.img,os8.img,freedos722.img%7D
404 Not Found
0.887 2024-01-02 22:20:11 ERROR 404: Not Found.
0.887 
```

Also, I don't believe these files will end up in the correct `images/` directory. The `images/` directory is inside of the repo, which isn't available when the container is built.

.devcontainer/devcontainer.json
cognitivegears: Thanks - reformatted.

.devcontainer/Dockerfile
cognitivegears: This is done as well. I moved it to a task instead of building during Dockerfile startup. I also tested fresh installs with both a Docker remote container and a Github codespace.

.devcontainer/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM library/node:lts-bookworm
 
 ARG DEBIAN_FRONTEND=noninteractive
 RUN apt update && \
-    apt-get install -y nodejs nasm gdb unzip p7zip-full openjdk-17-jre wget python3 qemu-system-x86 git-core build-essential libc6-dev-i386-cross libc6-dev-i386 clang curl time
```
copy: `make qemutests` is now failing, could you check if this change is related?

```
gcc -m32 -Wall -O2 -g -fno-strict-aliasing -static -mmmx -msse  -o test-i386 \
              ./test-i386.c ./test-i386-code16.S ./test-i386-vm86.S -lm
In file included from /usr/include/features.h:392,
                 from /usr/include/assert.h:35,
                 from ./test-i386.c:21:
/usr/include/features-time64.h:20:10: fatal error: bits/wordsize.h: No such file or directory
   20 | #include <bits/wordsize.h>
```

.vscode/tasks.json
```diff
@@ -0,0 +1,12 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "Fetch images",
+            "type": "shell",
+            "command": "wget -P images/ https://k.copy.sh/{linux.iso,linux4.iso,buildroot-bzimage.bin,openbsd-floppy.img,kolibri.img,windows101.img,os8.img,freedos722.img}"
```
copy, line 9: `linux3.iso` is missing from this, which causes `make tests` to fail.

copy (commented):
This works pretty well now (and it's a nice feature for windows users or people who don't want to install all of these build tools).

I've left two small suggestions which should allow running all tests, then it's good to merge.

copy:
Thanks!

PR #971, opened 2024-01-02
change multiboot to boot as an option rom, add initrd, cmdline

russor:
This fleshes out the multiboot support a bit more. My hobby os relies on multiboot, but also needs some ACPI tables set. Using the existing multiboot support means seabios doesn't run at all, and the tables are missing.

The multiboot content is loaded in an i/o write hook rather than before booting, because otherwise seabios may overwrite the loaded data (happens with the kvm-test).

load_multiboot still works, although the mechanism changed, because nasmtests uses it directly and doesn't load seabios anyway. Now it generates the option rom, loads it at address 0 (seemed like a convenient place) and adjusts execution to start at address 3 (where an option rom is jumped to in a BIOS context).

There's a not entirely related log fix in elf section header debug info (log said program headers, but was reporting about section headers); happy to put that separate, but it came up while I was working on this.

### DIFF
--- a/src/cpu.js
+++ b/src/cpu.js
@@ -1060,12 +1060,12 @@ CPU.prototype.load_multiboot_option_rom = function(buffer, initrd, cmdline)
         // bit 1 : provide a memory map (which we always will)
         dbg_assert((flags & ~MULTIBOOT_HEADER_ADDRESS & ~3) === 0, "TODO");
 
-        let multiboot_info_addr = 0x7C00;
         // do this in a io register hook, so it can happen after BIOS does its work
         var cpu = this;
 
         this.io.register_read(0xF4, this, function () {return 0;} , function () { return 0;}, function () {
-            // actually do the load and return the entrypoint address
+            // actually do the load and return the multiboot magic
+            let multiboot_info_addr = 0x7C00;
             let multiboot_data = multiboot_info_addr + MULTIBOOT_INFO_STRUCT_LEN;
             let info = 0;
 
@@ -1160,7 +1160,7 @@ CPU.prototype.load_multiboot_option_rom = function(buffer, initrd, cmdline)
 
                 let elf = read_elf(buffer);
 
-                entrypoint = elf.header.entry | 0;
+                entrypoint = elf.header.entry;
 
                 for(let program of elf.program_headers)
                 {
@@ -1172,9 +1172,6 @@ CPU.prototype.load_multiboot_option_rom = function(buffer, initrd, cmdline)
                     {
                         // load
 
-                        // Since multiboot specifies that paging is disabled,
-                        // virtual and physical address must be equal
-                        dbg_assert(program.paddr === program.vaddr);
                         dbg_assert(program.filesz <= program.memsz);
 
                         if(program.paddr + program.memsz < cpu.memory_size[0])
@@ -1186,6 +1183,13 @@ CPU.prototype.load_multiboot_option_rom = function(buffer, initrd, cmdline)
                             }
                             top_of_load = Math.max(top_of_load, program.paddr + program.memsz);
                             dbg_log("prg load " + program.paddr + " to " + (program.paddr + program.memsz), LOG_CPU);
+
+                            // Since multiboot specifies that paging is disabled, we load to the physical address;
+                            // but the entry point is specified in virtual addresses so adjust the entrypoint if needed
+
+                            if (entrypoint == elf.header.entry && program.vaddr <= entrypoint && (program.vaddr + program.memsz) > entrypoint) {
+                                entrypoint = (entrypoint - program.vaddr) + program.paddr;
+                            }
                         }
                         else
                         {
--- a/src/cpu.js
+++ b/src/cpu.js
@@ -980,8 +980,13 @@ CPU.prototype.init = function(settings, device_bus)
 
         if(option_rom)
         {
-            dbg_log("adding option rom for multiboot", LOG_CPU);
-            this.option_roms.push(option_rom);
+            if (this.bios.main) {
+                dbg_log("adding option rom for multiboot", LOG_CPU);
+                this.option_roms.push(option_rom);
+            } else {
+                dbg_log("loaded multiboot", LOG_CPU);
+                this.reg32[REG_EAX] = this.io.port_read32(0xF4);
+            }
         }
     }
 
@@ -1001,11 +1006,7 @@ CPU.prototype.load_multiboot = function (buffer)
     if(option_rom)
     {
         dbg_log("loaded multiboot", LOG_CPU);
-        this.write_blob(option_rom.data, 0);
-        // set CS segment to 0
-        this.segment_offsets[1] = 0;
-        this.sreg[1] = 0;
-        this.instruction_pointer[0] = this.get_seg_cs() + 0x3 | 0;
+        this.reg32[REG_EAX] = this.io.port_read32(0xF4);
     }
 };
 
@@ -1019,6 +1020,7 @@ CPU.prototype.load_multiboot_option_rom = function(buffer, initrd, cmdline)
     const MULTIBOOT_HEADER_MAGIC = 0x1BADB002;
     const MULTIBOOT_HEADER_MEMORY_INFO = 0x2;
     const MULTIBOOT_HEADER_ADDRESS = 0x10000;
+    const MULTIBOOT_BOOTLOADER_MAGIC = 0x2BADB002;
     const MULTIBOOT_SEARCH_BYTES = 8192;
     const MULTIBOOT_INFO_STRUCT_LEN = 116;
     const MULTIBOOT_INFO_CMDLINE = 0x4;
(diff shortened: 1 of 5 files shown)


src/cpu.js
```diff
@@ -1163,12 +1251,165 @@ CPU.prototype.load_multiboot = function(buffer)
...
+        data8[i++] = 0x2b;
+        // place multiboot header location in ebx
+        data8[i++] = 0xbb; // mov ebx, multiboot_info_addr
+        data8[i++] = multiboot_info_addr;
+        data8[i++] = multiboot_info_addr >> 8;
+        data8[i++] = multiboot_info_addr >> 16;
+        data8[i++] = multiboot_info_addr >> 24;
+        // jump to multiboot entrypoint
+        data8[i++] = 0xFF; // jmp rcx
+        data8[i++] = 0xE1;
```
copy, line 1390: Since we're using a (somewhat ugly) io port hook anyway, wouldn't it be simpler to implement all of this in JavaScript in the hook?

It would also remove this extra work from nasmtests (i.e. load_multiboot could be implement as exactly this callback plus setting cs/eip, if that makes sense)

src/cpu.js
```diff
@@ -1163,12 +1251,165 @@ CPU.prototype.load_multiboot = function(buffer)
             this.io.register_write(0x2000 + i, this, handle_write, handle_write, handle_write);
         }
 
-        this.update_state_flags();
+        // This rom will be executed by seabios after its initialisation
+        // It sets up the multiboot environment.
+        const SIZE = 0x200;
+
+        const data8 = new Uint8Array(0x100);
```
copy, line 1258: I believe this should be `new Uint8Array(SIZE);` (also in kernel.js)

src/cpu.js
```diff
@@ -974,7 +981,25 @@ CPU.prototype.init = function(settings, device_bus)
...
+{
+    if (this.bios.main) {
+        dbg_assert(false, "load_multiboot not supported with BIOS");
+    }
+
+    const { option_rom } = this.load_multiboot_option_rom(buffer, false, "");
+    if(option_rom)
+    {
+        dbg_log("loaded multiboot", LOG_CPU);
+        this.write_blob(new Uint8Array(option_rom.data), 0);
```
copy, line 994: I believe the `new Uint8Array` makes an unnecessary copy of the option rom and can be removed.

src/cpu.js
```diff
@@ -1015,123 +1040,186 @@ CPU.prototype.load_multiboot = function(buffer)
...
+                    cpu.write32(multiboot_data + 16, 0);
+                    cpu.write32(multiboot_data + 20, 1); // type (MULTIBOOT_MEMORY_AVAILABLE)
+                    multiboot_data += 24;
+                    multiboot_mmap_count += 24;
+                    was_memory = false;
+                } else if (!was_memory && cpu.memory_map_read8[addr >>> MMAP_BLOCK_BITS] === undefined) {
+                    start = addr;
+                    was_memory = true;
+                }
+            }
```
copy, line 1091: Out of curiosity, is this used/required by your OS or Linux (or both)?

src/cpu.js
```diff
@@ -1015,123 +1040,186 @@ CPU.prototype.load_multiboot = function(buffer)
...
         let multiboot_info_addr = 0x7C00;
-        this.reg32[REG_EBX] = multiboot_info_addr;
-        this.write32(multiboot_info_addr, 0);
+        // actually do the load and return the entrypoint address
+        // do this in a io register hook, so it can happen after BIOS does its work
+        var cpu = this;
+
+        this.io.register_read(0xF4, this, function () {return 0;} , function () { return 0;}, function () {
+            let multiboot_data = multiboot_info_addr + 116;
+            cpu.write32(multiboot_info_addr, ((1 << 2) |  // cmdline set
```
copy, line 1055: Should this be 0 if no cmdline was passed?

src/cpu.js
```diff
@@ -1015,123 +1040,186 @@ CPU.prototype.load_multiboot = function(buffer)
...
 
         let multiboot_info_addr = 0x7C00;
-        this.reg32[REG_EBX] = multiboot_info_addr;
-        this.write32(multiboot_info_addr, 0);
+        // actually do the load and return the entrypoint address
+        // do this in a io register hook, so it can happen after BIOS does its work
+        var cpu = this;
+
+        this.io.register_read(0xF4, this, function () {return 0;} , function () { return 0;}, function () {
+            let multiboot_data = multiboot_info_addr + 116;
```
copy, line 1054: Could you give this constant (116) a name?

src/cpu.js
```diff
@@ -1015,123 +1040,186 @@ CPU.prototype.load_multiboot = function(buffer)
...
+            // memory map
+            let multiboot_mmap_count = 0;
+            cpu.write32(multiboot_info_addr + 44, 0);
+            cpu.write32(multiboot_info_addr + 48, multiboot_data);
+
+            // Create a memory map for the multiboot kernel
+            // does not exclude traditional bios exclusions
+            let start = 0;
+            let was_memory = false;
+            for (let addr = 0; addr < MMAP_MAX; addr += MMAP_BLOCK_SIZE) {
```
copy, line 1076: I think this loop could stop at `this.memory_size[0]`.

copy (commented):
Thanks!

Testing-wise, I would like to add your OS to the testsuite (in tests/full/run.js). Could you upload a (preferably small) multiboot image? Also, I assume this could also be used to boot Linux kernels. Have you tested that?

Overall the code looks good. My main bigger suggestion would be to remove the assembly and implement it directly in the io port hook.

src/cpu.js
russor: That's a good idea! We need a small boot rom so that we can trigger the io port hook after bios does its thing, but it can be really short.

russor:
> Thanks!
> 
> Testing-wise, I would like to add your OS to the testsuite (in tests/full/run.js). Could you upload a (preferably small) multiboot image? 

I hadn't released it yet, but there's a time for everything. The multiboot image is < 1M; it's more interesting if it has the ~20M initrd though... but that's then a bit big. Either way, I'll need to do some cleanup beforehand. 

> Also, I assume this could also be used to boot Linux kernels. Have you tested that?
> 

I have not. So far, I only tested my OS and the tests in the repo. I'll see what's out there that I can get to boot with this after I address the code comments and remove the bulk of the assembly.

> Overall the code looks good. My main bigger suggestion would be to remove the assembly and implement it directly in the io port hook.

russor:
I think this addresses most of the comments, but I have some more work to do now; which I'll try to get done over the next few days.

FreeBSD and Linux don't appear to have multiboot kernels (at least the ones I tried)... but NetBSD does.

When attempting to load an image that isn't a multiboot image, at least a proper console message would be appropriate, instead of a TypeError.

When attempting to load NetBSD, it's hitting an assert in this multiboot loader; and I'm going to dig into why, and see if I can fix it. :D

russor:
Ok, so NetBSD and GNU Hurd kernels both have virtual address != physical address. I looked around, and it seems like the thing to do is load at the physical address, adjust the entry point address from a virtual address to the physical address and jump to that. I didn't see that in the specification (my kernel, and the v86 unit tests have physical == virtual), but it's what grub does.

With that change, Hurd does something --- lots of debug output from cpuid and friends, but no output on screen or com1 (probably needs some config or command line). Behavior seems similar with or without bios.

NetBSD does start and outputs some boot logs to VGA; then it hits an unexpected overlapped read in ne2k. Excitingly, NetBSD only works if run with SeaBIOS... without a bios, it crashes the emulator with a bad instruction; so that's a good case for me. NetBSD does throw a warning about the memory map, so there's probably improvements to be made there at some point; but I think this is OK for now.

I also tweaked things so multiboot and linux kernel loading don't crash if they fail to load. With seabios, that means it'll try to load something else, which doesn't work, but I think that's reasonable... better than just getting stuck on a error?

Let me know if you want me to squash this into fewer commits, or feel free to adjust as you like.

russor:
> Testing-wise, I would like to add your OS to the testsuite (in tests/full/run.js). Could you upload a (preferably small) multiboot image?

I've put up https://crazierl.org/crazierl.elf and https://crazierl.org/initrd -- using the initrd, a command line of "kernel /libexec/ld-elf32.so.1" is required, works in my demo page https://crazierl.org/demo.html in case any other parameters are important. Expectation if you boot without initrd is it will start and eventually print "end of kernel!" on vga and com1, and halt.  If you boot with the initrd, it should end up with a prompt like "1>" where you can type "5 + 5." and hit enter, and get back 10 and a new prompt.

VGA output is wonky; my demo page uses com1 only. I may provide a future PR to make vga work better; I set the CRTC Offset Register http://www.osdever.net/FreeVGA/vga/crtcreg.htm#13 so text lines are 128 characters in memory (but still 80 on screen), which makes scrolling with wrap around much simpler.

copy:
Thanks, merged!

I'll add NetBSD (even though it doesn't fully boot, I believe unrelated to multiboot) and crazierl as tests soon. I'll also add crazierl to the website, as it's pretty cool :-)

Issue #138, opened 2017-03-14
DOS dialup terminal to websocket gateway


tracker1:
Just curious if anyone has any tips if I wanted to do FreeDOS with an old dos dialup terminal program and use a websocket gateway to a telnet or similar endpoint.

copy:
This is technically possible, basically have a look at https://github.com/copy/v86/blob/master/src/browser/serial.js

This function handles interaction between a serial console and a `<textarea>` element, you basically need to replace the `<textarea>` stuff with websocket calls.

Alternatively, if you find an ne2k driver for FreeDOS, you could use the normal network stack. The backend we're currently using is https://github.com/benjamincburns/websockproxy.

cgorringe:
I had this same idea. I'd like to run some old BBS software in DOSBox/FreeDOS on a server, using a websocket gateway to RIPTerm 1.54 running from the browser.

jfmherokiller:
if there is any interest in this still remaining you might want to try looking at this https://en.wikipedia.org/wiki/Hayes_command_set its the general standard for modems.

tracker1:
@jfmherokiller I'm definitely still interrested, just totally lost track and hadn't looked in a while... I don't think I really need the full hayes support, as I should be able to ignore any `AT*` other than `ATDT *`  where * is host:port ... and even then, I can probably ignore that if I can activate the terminal in a connected mode.

copy:
This is implemented in #1523 (with some context in #1512). Happy testing!

PR #980, opened 2024-01-19
Add modem status functionality to UART class

cognitivegears:
This PR adds modem status functionality to UART.js. A couple of use cases for this:

1) Running a retro BBS system or BBS door game within a browser or node.js
2) Running old DOS modem terminal programs - part of what is needed for Issue #138

...or anything else that uses the UART that expects flags such as carrier detect, etc to be set.

### DIFF
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -1122,6 +1122,46 @@ V86.prototype.serial_send_bytes = function(serial, data)
 };
 
 /**
+ * Set the modem status of a serial port.
+ */
+V86.prototype.serial_set_modem_status = function(serial, status)
+{
+    this.bus.send("serial" + serial + "-modem-status-input", status);
+}
+
+/**
+ * Set the carrier detect status of a serial port.
+ */
+V86.prototype.serial_set_carrier_detect = function(serial, status)
+{
+    this.bus.send("serial" + serial + "-carrier-detect-input", status);
+}
+
+/**
+ * Set the ring indicator status of a serial port.
+ */
+V86.prototype.serial_set_ring_indicator = function(serial, status)
+{
+    this.bus.send("serial" + serial + "-ring-indicator-input", status);
+}
+
+/**
+ * Set the data set ready status of a serial port.
+ */
+V86.prototype.serial_set_data_set_ready = function(serial, status)
+{
+    this.bus.send("serial" + serial + "-data-set-ready-input", status);
+}
+
+/**
+ * Set the clear to send status of a serial port.
+ */
+V86.prototype.serial_set_clear_to_send = function(serial, status)
+{
+    this.bus.send("serial" + serial + "-clear-to-send-input", status);
+}
+
+/**
  * Mount another filesystem to the current filesystem.
  * @param {string} path Path for the mount point
  * @param {string|undefined} baseurl
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -1127,7 +1127,7 @@ V86.prototype.serial_send_bytes = function(serial, data)
 V86.prototype.serial_set_modem_status = function(serial, status)
 {
     this.bus.send("serial" + serial + "-modem-status-input", status);
-}
+};
 
 /**
  * Set the carrier detect status of a serial port.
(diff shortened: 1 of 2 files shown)


src/browser/starter.js
```diff
@@ -1121,6 +1121,46 @@ V86.prototype.serial_send_bytes = function(serial, data)
...
+    this.bus.send("serial" + serial + "-data-set-ready-input", status);
+};
+
+/**
+ * Set the clear to send status of a serial port.
+ */
+V86.prototype.serial_set_clear_to_send = function(serial, status)
+{
+    this.bus.send("serial" + serial + "-clear-to-send-input", status);
+};
```
copy, line 1162: This wouldn't prevent merging, but do you have a test for this? (either some program that actually calls these or a unit test that calls these functions and checks some output from the guest OS)

src/uart.js
```diff
@@ -98,6 +109,53 @@ function UART(cpu, port, bus)
...
+    }, this);
+
+    this.bus.register("serial" + this.com + "-clear-to-send-input", function(data)
+    {
+        if(data) {
+            this.set_modem_status(this.modem_status | (1 << UART_MSR_CTS) | (1 << UART_MSR_DCTS));
+        }
+        else {
+            this.set_modem_status(this.modem_status & ~(1 << UART_MSR_CTS) & ~(1 << UART_MSR_DCTS));
+        }
```
copy, line 156: I'd say these are better written like this:

```js
const status = data ?
    this.modem_status | (1 << UART_MSR_CTS) | (1 << UART_MSR_DCTS) : 
    this.modem_status & ~(1 << UART_MSR_CTS) & ~(1 << UART_MSR_DCTS);
this.set_modem_status(status);
```

copy (commented):
Thanks!

> Running a retro BBS system or BBS door game within a browser or node.js

That sounds pretty cool, is this something you've been working on?

The code looks good, I've left two small suggestions. Before merging, could you do a quick pass to make sure the style is the same as adjacent code (i.e. indentation, newline between `)` and `{`?

cognitivegears:
> Thanks!
> 
> > Running a retro BBS system or BBS door game within a browser or node.js
> 
> That sounds pretty cool, is this something you've been working on?
> 


It is :) I'm a contributor to Enigma 1/2 BBS - https://github.com/NuSkooler/enigma-bbs which is a BBS written in node.js. One of the toughest things to do with it is setting up BBS Door games. It has to shell out to an emulator. So instead, I'd like to use v86 to run the BBS DOS door game directly.

foodfight

This requires a change like this though, because most Door games check the carrier detect to see if someone has hung up.

> The code looks good, I've left two small suggestions. Before merging, could you do a quick pass to make sure the style is the same as adjacent code (i.e. indentation, newline between `)` and `{`?

Will do, I ended up a little busy with work this week but should be able to finish this off this weekend. More to come.

src/uart.js
cognitivegears: Changed, thanks!

src/browser/starter.js
cognitivegears: I spent tome time on this, but haven't been able to find an easy way to test this. It's easy enough to add an api-test along with serial.js, but the problem is that I am having trouble finding current guest OS that have support for checking these flags built-in (i.e. minicom, etc.) I could either write a C program to check them, or create a new/modified image that has a program that cares about these flags, but I'm having trouble finding one that already has something that would work for this (at least with the LInux and DOS images, I'm not as familiar with some of the others to see.) If you have any suggestions for how to proceed I'd be happy to do so, otherwise I'm afraid I don't have one. (I do have external programs using my own images that I've been testing with, but none of the current ones from v86.)

copy: No problem!

copy:
> It is :) I'm a contributor to Enigma 1/2 BBS - https://github.com/NuSkooler/enigma-bbs which is a BBS written in node.js. One of the toughest things to do with it is setting up BBS Door games. It has to shell out to an emulator. So instead, I'd like to use v86 to run the BBS DOS door game directly.

That's very cool :-)

Thanks for the contribution, merged!

PR #992, opened 2024-02-12
vga: support line_compare and offset_register in text mode

russor:
There's not a whole lot of use of these registers in text mode in the wild, but crazierl uses it, so I'm motivated. :)

Math.max(this.max_cols, this.offset_register * 2) is a bit of a cop-out, I think VGA user should likely be properly setting the offset register, but I didn't want things to fail in case they don't.


russor:
This tests good for me. And your cleanups are much nicer than what I wrote. :)

copy:
Cheers!

PR #1002, opened 2024-03-13
Virtio console: Send bytes to the right queues

jmezzacappa:
I discovered that all virtio console data was getting sent to the same port (`/dev/hvc0`) when doing something like this:

```js
const encoder = new TextEncoder()
emulator.bus.send('virtio-console1-input-bytes', encoder.encode('hello world 1\n'))
emulator.bus.send('virtio-console2-input-bytes', encoder.encode('hello world 2\n'))
emulator.bus.send('virtio-console3-input-bytes', encoder.encode('hello world 3\n'))
```

When looking at the code, I noticed that the variable `queue` (the first one) is unused. I renamed it to `queue_id` (because there's another `queue` variable inside the event listener), plugged it into where it looked like it should go, and it worked :)

### DIFF
--- a/src/virtio_console.js
+++ b/src/virtio_console.js
@@ -204,12 +204,12 @@ function VirtioConsole(cpu, bus)
     });
 
     for (let port = 0; port < this.ports; ++port) {
-        let queue = port == 0 ? 0 : port * 2 + 2;
+        let queue_id = port == 0 ? 0 : port * 2 + 2;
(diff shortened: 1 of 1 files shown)


copy:
Thanks!

basicer:
Woops! 🙃

jmezzacappa:
> Woops! 🙃

You handled the hard parts :D The terminal feels faster and resizing is simpler to handle. Thanks for this feature!

PR #1008, opened 2024-04-01
kernel.js: cmdline_size is only valid for linux boot protocol 2.06+

rouseabout:


rouseabout:
I have used v86's linux boot protocol loader to successfully boot into my hobby/toy operating system. Thanks.

<https://pross.sdf.org/sandpit/os/>

copy:
Thanks!

PR #1021, opened 2024-04-18
Fix PS/2 keyboard identify response

Bananymous:
Add hexadecimal prefix to identify response's second byte. Now PS/2 keyboard responds with the correct identification bytes 0xAB, 0x83.


copy:
Thanks!

copy:
Tested and doesn't break any OSes.

Thanks!

PR #1022, opened 2024-04-18
PageFault error code now reports page presence correctly

Bananymous:
If page directory was present and ring3 access or write protection violation caused page fault on page directory level, page presence was not checked and it was assumed to be present.

This PR adds this check to correctly report present bit in page fault error codes.

I found this bug as my OS uses the present bit in error code of page fault as the initial requirement for demand paging.

This is my first time writing rust, so please tell me if something should be done differently.

### DIFF
--- a/src/rust/cpu/cpu.rs
+++ b/src/rust/cpu/cpu.rs
@@ -1928,26 +1928,20 @@ pub unsafe fn do_page_walk(
         }
 
         let kernel_write_override = !user && 0 == cr0 & CR0_WP;
-        if page_dir_entry & PAGE_TABLE_RW_MASK == 0 && !kernel_write_override && for_writing {
-            if side_effects {
-                trigger_pagefault(addr, true, for_writing, user, jit);
-            }
-            return Err(());
-        }
+        let mut allow_write = page_dir_entry & PAGE_TABLE_RW_MASK != 0;
 
-        if page_dir_entry & PAGE_TABLE_USER_MASK == 0 {
-            allow_user = false;
-            if user {
-                // Page Fault: page table accessed by non-supervisor
+        allow_user &= page_dir_entry & PAGE_TABLE_USER_MASK != 0;
+
+        if 0 != page_dir_entry & PAGE_TABLE_PSE_MASK && 0 != cr4 & CR4_PSE {
+            // size bit is set
+
+            if for_writing && !allow_write && !kernel_write_override {
                 if side_effects {
                     trigger_pagefault(addr, true, for_writing, user, jit);
                 }
                 return Err(());
             }
-        }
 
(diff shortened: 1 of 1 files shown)


copy:
Interesting find, this is a pretty serious bug.

> This is my first time writing rust, so please tell me if something should be done differently.

I think you could avoid the extra function by moving the write/user checks down and combining them with the page table checks. I'll leave a suggestion in the review.

I'll try to write some tests for this.

src/rust/cpu/cpu.rs
```diff
@@ -1940,7 +1943,10 @@ pub unsafe fn do_page_walk(
...
                 // Page Fault: page table accessed by non-supervisor
                 if side_effects {
-                    trigger_pagefault(addr, true, for_writing, user, jit);
+                    // NOTE: page presence has to be checked since we
+                    //       don't know about page table entry
+                    let is_present = is_page_present_in_page_directory(addr, page_dir_entry);
+                    trigger_pagefault(addr, is_present, for_writing, user, jit);
                 }
                 return Err(());
             }
```
copy, line 1952: Change this entire block to `allow_user &= page_dir_entry & PAGE_TABLE_USER_MASK != 0;`

Then, in the page table entry rewrite the user check to:

```rust
allow_user &= page_table_entry & PAGE_TABLE_USER_MASK != 0;
if user && !allow_user {
    if side_effects {
        trigger_pagefault(addr, true, for_writing, user, jit);
    }
    return Err(());
}
```

The size bit set case also needs a similar check then.

The same should work for write checks (need to introduce a variable similar to `allow_user`).

Bananymous: Thank you, I will do the changes soon.

Bananymous:
Done.

src/rust/cpu/cpu.rs
```diff
@@ -1928,26 +1928,20 @@ pub unsafe fn do_page_walk(
...
+        allow_user &= page_dir_entry & PAGE_TABLE_USER_MASK != 0;
+
+        if 0 != page_dir_entry & PAGE_TABLE_PSE_MASK && 0 != cr4 & CR4_PSE {
+            // size bit is set
 
-        if page_dir_entry & PAGE_TABLE_USER_MASK == 0 {
-            allow_user = false;
-            if user {
-                // Page Fault: page table accessed by non-supervisor
+            if for_writing && !allow_write && !kernel_write_override {
```
copy, line 1938: The user check is missing here now. I'd suggest merging it into the write check, since the body is the same (and also merge it below).

Bananymous: I somehow completely forgot the user check there. These changes are now done.

copy:
Thanks!

PR #1023, opened 2024-04-21
Report current PS/2 scan code set as 1

Bananymous:
PS/2 keyboard reported current scan code set as 2 while only scan code set 1 was emulated.


copy:
Thanks!

PR #1030, opened 2024-05-04
Readme: Changes in compatibility section

theonlyasdk:
Add clarification on which versions of SerenityOS works in v86.


TCFFan123:
Should not need to clarify that it's 32-bit, most people already know that v86 can only emulate 32-bit.

TCFFan123:
Also don't need to clarify that it's old, because well... it's also emulating a Pentium 4.
Pull request should be closed.

PR #1043, opened 2024-05-17
Create manifest.json

LavedenC1:


spetterman66:
You have left the default values, and there is no manifest specified in the index.html file.

copy:
Feel free to reopen with an explanation of how this is useful.

PR #1052, opened 2024-05-26
Text Mode enhancement

vip-delete:
1. Add BLINK flag support (3rd bit in VGA Attribute Mode Control Register).

> When this bit is set to 0, the most significant bit of the attribute selects the background intensity (allows 16 colors for background). When set to 1, this bit enables blinking.

blink

A test program which writes a blinking text to the text buffer

```
org 100h

mov ax, 0xB800 ; text buffer address
mov es, ax     ; segment register points to the text buffer
xor di, di     ; text buffer offset

mov ah, 0x92   ; XRGBIRGB, X-blink, RGB-bg, I-intensity, RGB-fg
xor bx, bx     ; text buffer index
mov cx, len    ; message length
mov bp, msg    ; message address

print:
mov al, [bp]            ; put next char from msg to al
mov word [es:di+bx], ax ; direct write to the text buffer
add bx, 2               ; video memory cell is 2-bytes long
inc bp                  ; move to the next char from msg
loop print

mov ah, 94h             ; red on blue (blink)
mov al, 3               ; heart char
mov word [es:di+bx], ax ; direct write to the text buffer

xor ah, ah     ; press any key to continue
int 16h

mov ah, 4Ch    ; exit
int 21h

msg db 'Hello + World = '
len equ $ - msg

```

A test program to clear/set BLINK flag
```
org 100h

mov dx, 3DAh
in al, dx

mov dx, 3C0h
mov al, 30h
out dx, al

inc dx
in al, dx
and al, 11110111b ; clear BLINK flag
;or al, 00001000b ; set BLINK flag
dec dx
out dx, al

mov ah, 4Ch    ; exit
int 21h
```

SuperCalc 4 clears BLINK on start to use high intensity yellow for background:

sc

on exit, it sets the flag back.

2. fill the cursor with the foreground color of the character at the cursor's current location:
cursor1
cursor2
cursor3

### DIFF
--- a/v86.css
+++ b/v86.css
@@ -307,5 +307,22 @@ h4 {
 }
 
 .blink {
-    animation: blink calc(1s / 4.380) step-start infinite;
+    animation: blink 0.6s step-start infinite;
 }
+
+.cursor {
+    animation: cursor 0.2s step-start infinite;
+}
+
+@keyframes blink {
+    50% {
+        color: transparent;
+    }
+}
+
+@keyframes cursor {
+    50% {
+        background-color: transparent;
+    }
+}
+
--- a/v86.css
+++ b/v86.css
@@ -130,6 +130,22 @@ h4 {
 #terminal {
     max-width: 1024px;
 }
+.blink {
+    animation: blink 0.6s step-start infinite;
+}
+@keyframes blink {
+    50% {
+        color: transparent;
+    }
+}
+.cursor {
+    animation: cursor 0.6s step-start infinite;
+}
+@keyframes cursor {
+    50% {
+        background-color: transparent;
+    }
+}
 
 /* the code below was copied from xterm.css */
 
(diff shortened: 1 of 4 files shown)


v86.css
```diff
@@ -305,3 +305,24 @@ h4 {
     z-index: 2;
     position: relative;
 }
+
+.blink {
+    animation: blink 0.6s step-start infinite;
+}
+
+.cursor {
+    animation: cursor 0.2s step-start infinite;
```
copy, line 314: I don't understand this code. Is the cursor background colour somehow animated?

copy (commented):
Thanks! I've left a few comments.

Just like for the other PR, could you send me a disk image with the programs to reproduce the issues? (ideally uploaded in github)

v86.css
vip-delete: yes, 0.1s background is as it is, 0.1s background is transparent (cursor is not visible), then all repeats.

vip-delete:
> (ideally uploaded in github)

uploaded https://github.com/copy/v86/pull/1055#issuecomment-2142656642

vip-delete:
is there any action required from my side to mark the PR somehow that all is done?

copy:
> is there any action required from my side to mark the PR somehow that all is done?

I'm a bit busy right now and haven't gotten around to testing this yet. I'll try to get it done some time next week.

v86.css
copy: Makes sense. FYI, I'll change this to 0.6, because 0.2 is too fast for my taste, and 0.6 roughly matches qemu's cursor blinking speed.

PR #1060, opened 2024-06-15
Remove the duplicate line.

Joinhack:


copy:
The two lines are not identical.

PR #1061, opened 2024-06-20
network: Fetch based browser networking

### LINKED ISSUE #198: Use networking without the websocket proxy?
Currently, networking is run through https://github.com/benjamincburns/websockproxy which needs a server to be set up and have v86 pointed at it. It also limits the speed. I was wondering if we could run the networking through javascript calls instead or by some other local solution. (websockproxy also doesn't support https so it would be nice if there was a solution that could use that)

basicer:
Emulate a networking stack, intercept HTTP requests, serve them with fetch().  Enable by setting networking_proxy=fetch

Somewhat fixes #198

Just enough networking for http to work for package managers and net-boot. The networking stack just reflects the mac address of packets it receives, so shouldn't be bothered by mac address changes. Services include:

- ARP replies by establishing the gateways IP
- DHCP server sending gateway, dns, and client-ip
  - Gateway: 192.168.86.1
  - Client: 192.168.86.100
  - Netmask: 255.255.255.0
- UDP DNS server, all addresses resolve to: 192.168.87.1
- NTP Server, giving out the current time from Date()
- Replies to ICMP pings to any IP address
- UDP echo server on port 8

Limitations:
- Request are limited by CORS rules, but CORS-proxies can work around this.
- Redirects are handled transparently by fetch. The upside is sites that force an http->https upgrade appear to the VM to work over http.

Future Work:
- Allow HTTP requests from browser to be served by VM.
- HTTPS support (maybe with help of a rust based TLS stack in WASM)
- Support for POSTing binary data


SuperMaxusa:
> * UDP DNS server, all addresses resolve to: 192.168.87.1

How about use a [DNS over HTTPS](https://en.wikipedia.org/wiki/DNS_over_HTTPS) (https://datatracker.ietf.org/doc/html/rfc8484) or this not suitable for this case?

basicer:
I thought about doing that, or making it an option.  It's neat for the vm
to know the real ip, but doesn't really unlock anything interesting.  Also,
1.1.1.1 didn't seem to support CORS.

src/browser/fetch_network.js
```diff
@@ -0,0 +1,1103 @@
...
+    view.setUint32(12, spec.dhcp.ciaddr);
+    view.setUint32(16, spec.dhcp.yiaddr);
+    view.setUint32(20, spec.dhcp.siaddr);
+    view.setUint32(24, spec.dhcp.giaddr);
+
+    for (let i = 0; i < spec.dhcp.chaddr.length; ++i) {
+        view.setUint8(28+i, spec.dhcp.chaddr[i]);
+    }
+
+    view.setUint32(236, 0x63825363);
```
SuperMaxusa, line 750: How about idea to replace it to:
```js
view.setUint32(236, DHCP_MAGIC_COOKIE);
```
and set in upper:
```js
const DHCP_MAGIC_COOKIE = 0x63825363;
```
as small improvement of readability code (like `IPV4_PROTO_*` and `ETHERTYPE_*` variables)?

tools/fetch_network_harness.js
```diff
@@ -0,0 +1,41 @@
+const dgram = require('node:dgram');
+const server = dgram.createSocket('udp4');
+const fs = require('node:fs');
+const path = require('node:path');
+
+//  qemu-system-i386 -m 2G -nographic -hda ~/disk.qcow2 -netdev dgram,id=net0,remote.type=inet,remote.host=127.0.0.1,remote.port=6677,local.host=127.0.0.1,local.port=7744,local.type=inet -device e1000,netdev=net0
+
+globalThis.dbg_assert = require('node:assert');
+
+let FetchNetworkAdapter = eval(fs.readFileSync(path.join(__dirname, 'browser', 'fetch_network.js'), 'utf-8') + ";FetchNetworkAdapter");
```
copy, line 10: At the bottom fetch_network.js, you could write:

```js
if(typeof module !== "undefined" && typeof module.exports !== "undefined")
{
    module.exports["FetchNetworkAdapter"] = FetchNetworkAdapter;
}
```

and use `require` instead.

src/browser/fetch_network.js
```diff
@@ -0,0 +1,1103 @@
...
+ * @param {BusConnector} bus
+ */
+function FetchNetworkAdapter(bus)
+{
+    let adapater = this;
+    this.bus = bus;
+    this.router_mac = new Uint8Array([82, 84, 0, 1, 2, 3]);
+    this.router_ip = new Uint8Array([192, 168, 86, 1]);
+    this.tcp_conn = {};
+    // this.cors_proxy = 'https://corsproxy.io/?'
```
copy, line 30: This should be configurable from the outside.

src/browser/fetch_network.js
```diff
@@ -0,0 +1,1103 @@
...
+        this.send(data);
+    }, this);
+
+    this.fetch = function(url, options) {
+        if (adapater.cors_proxy) url = adapater.cors_proxy + encodeURIComponent(url);
+        return fetch(url, options).then(function(resp) {
+            return resp.arrayBuffer().then(function(ab) {
+                return [resp, ab];
+            });
+        }).catch(err => {
```
copy, line 43: Should probably log this error to the console too.

copy (commented):
Thanks, this very cool.

From my first look:

- Some parts are rough performance-wise. This doesn't block a merge, but some TODO(perf) comments would be nice
- I think this should be exposed as an experimental option in index.html
- This needs a test (nodejs has fetch now). See [tests/devices/virtio_console.js
](https://github.com/copy/v86/blob/master/tests/devices/virtio_console.js) for an example to script a Linux guest through the serial console. Just a simple udhcpcd and curl should be sufficient for now.
- Some documentation should be added to docs/networking.md

> Allow HTTP requests from browser to be served by VM.

That would be really cool.

> HTTPS support (maybe with help of a rust based TLS stack in WASM)

Could be interesting, but it could also increase the size of v86.wasm significantly. I'd suggest sticking to http for this PR.

copy (commented):
Thanks, this looks mostly good now. I've left a few comments on the test.

Before merging, could you rebase on master, run `eslint src tests gen lib examples tools --fix` and fix the remaining issues manually?

tools/fetch_network_harness.js
```diff
@@ -0,0 +1,43 @@
...
+const fs = require('node:fs');
+const path = require('node:path');
+
+//  qemu-system-i386 -m 2G -nographic -hda ~/disk.qcow2 -netdev dgram,id=net0,remote.type=inet,remote.host=127.0.0.1,remote.port=6677,local.host=127.0.0.1,local.port=7744,local.type=inet -device e1000,netdev=net0
+
+globalThis.dbg_assert = require('node:assert');
+globalThis.dbg_log = (what, level) => console.log(what);
+globalThis.LOG_NET = 0;
+
+let FetchNetworkAdapter = require(path.join(__dirname, '..', 'src', 'browser', 'fetch_network.js'));
```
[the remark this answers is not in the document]
basicer: The downside of that is you have to build the bundle every time you make a change you want to test.

tests/devices/fetch_network.js
```diff
@@ -0,0 +1,358 @@
...
+            emulator.serial0_send("wget -q -O - example.org\n");
+            emulator.serial0_send("echo -e done\\\\thttpbin\n");
+        },
+        end_trigger: "done\thttpbin",
+        end: (capture, done) =>
+        {
+            assert(/This domain is for use in illustrative examples in documents/.test(capture), "got example.org text");
+            done();
+        },
+    },
```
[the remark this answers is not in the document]
basicer: You'd need to know the local ip I think for this to work out.  I think just stubbing out fetch with a version that's mocks might be the way to go.

PR #1074, opened 2024-07-07
Implement virtio network device

basicer:
Implement a virtio networking card.  Currently only supports one set of buffer pairs but seems to be a bit faster than ne2k.

Tested with linux4 and the alpine linux_virt kernel.


copy:
FYI, I'm reworking the `V86` constructor (to make more device configurations possible, e.g. disabling some default devices and allowing multiple network cards (with different backends)). I'm planning to merge this afterwards.

basicer:
> FYI, I'm reworking the `V86` constructor (to make more device configurations possible, e.g. disabling some default devices and allowing multiple network cards (with different backends)). I'm planning to merge this afterwards.

Awesome! I was going to suggest that.  The main interface was really showing its inflexibility.

copy:
cherry-picked into the wip branch. FYI, the api change is here: 60e0d5c31afce7301b15a0bcf521b4a68cf7b234 (I haven't implemented multiple network cards yet). Feedback would be welcome.

Issue #1091, opened 2024-07-25
fetch-based networking follow-ups


copy:
- Doesn't work on https (running `curl copy.sh` in buildroot):

```
Mixed Content: The page at 'https://copy.sh/v86/?profile=buildroot' was loaded over HTTPS, but requested an insecure resource 'http://copy.sh/'. This request has been blocked; the content must be served over HTTPS.
```

- In the arch profile, dhcp seems to have an issue:

```
root@localhost:~# ./networking.sh 
dhcpcd-9.4.1 starting
DUID 00:01:00:01:2e:34:fb:88:00:22:15:58:de:2b
eth0: IAID 15:58:de:2b
eth0: soliciting a DHCP lease
eth0: offered 192.168.86.100 from 192.168.86.1
eth0: probing address 192.168.86.100/24
eth0: 52:54:00:01:02:03(52:54:00:01:02:03) claims 192.168.86.100
eth0: DAD detected 192.168.86.100
eth0: soliciting a DHCP lease
eth0: offered 192.168.86.100 from 192.168.86.1
eth0: probing address 192.168.86.100/24
eth0: 52:54:00:01:02:03(52:54:00:01:02:03) claims 192.168.86.100
eth0: DAD detected 192.168.86.100
eth0: soliciting a DHCP lease
eth0: offered 192.168.86.100 from 192.168.86.1
eth0: probing address 192.168.86.100/24
eth0: 52:54:00:01:02:03(52:54:00:01:02:03) claims 192.168.86.100
eth0: DAD detected 192.168.86.100
...
```

- write some docs, in particular mention `chromium --user-data-dir=/tmp/temp-chromium-profile --disable-web-security`
- make corsproxy/masquerade, etc. configurable in the V86 constructor and in the UI

SuperMaxusa:
> * Doesn't work on https (running `curl copy.sh` in buildroot):

On Firefox with HTTPS-Only mode it just automatically upgrades to HTTPS and works without problems:

![firefox-https](https://github.com/user-attachments/assets/5be350ea-992f-4c9a-99a1-70d5c5202339)


Anyways, I think we could force changing protocol for `fetch()` if page with emulator was opened over HTTPS:
```diff
--- a/src/browser/fetch_network.js
+++ b/src/browser/fetch_network.js
@@ -1174,6 +1174,9 @@ TCPConnection.prototype.on_data_http = async function(data) {
         if(/^https?:/.test(first_line[1])) {
             target = new URL(first_line[1]);
[...]

PR #1097, opened 2024-07-30
wisp_network adapter

ProgrammerIn-wonderland:
WISP is a proxy protocol similar to SOCKS5 made by Mercury Workshop. It's documentation and a list of server implementation can be found at https://github.com/MercuryWorkshop/wisp-protocol and a demo server at wisps://wisp.mercurywork.shop/

This PR adds the ability to specify a wisp URL to use as a network proxy in the place of jor1k relay. The main benefit of this is that wisp is a lot less chatty since its not sending direct ethernet frames but rather just TCP and UDP payloads as needed, usually in 65k frame sizes instead of 1.5k ethernet frames. This results in a pretty good speed increase.

Here is some real world performance I get:
image

This PR only adds the TCP support and not UDP yet though UDP can be added in the future. Additionally it has a rudimentary DNS over HTTPS client which sends requests to google's DOH server to find the IP. It may not be a bad idea to make the DNS over HTTPS server a config option in the future?

Special Thanks:
@basicer For making fetch_network.js. This PR reuses a lot of what was introduced in that PR such as the TCP parser 
@ading2210 For creating the wisp protocol and hosting wisp.mercurywork.shop
the rest of Mercury Workshop for helping me as I needed help
and ofc copy and the other v86 contributors for making this amazing project I've looked up to since I was a child


src/browser/wisp_network.js
```diff
@@ -0,0 +1,571 @@
...
+function validateIPaddress(ipaddress) {
+    if(/^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/.test(ipaddress)) {
+      return true;
+    }
+    return false;
+}
+
+// DNS over HTTPS fetch, recursively fetch the A record until the first result is an IPv4
+async function dohdns(q) {
+    const prefered_fetch = (window.anura?.net?.fetch) || fetch;
```
copy, line 184: What is this about?

src/browser/wisp_network.js
```diff
@@ -0,0 +1,571 @@
...
+                {
+                    "name": q.name.join("."),
+                    "type": 1,
+                    "TTL": 600,
+                    "data": "127.0.0.1"
+                }
+            ]
+        };
+    } else {
+        throw new Error("DNS Server returned error code");
```
copy, line 218: Probably better to write `dbg_log("DNS Server returned error code: " + e, LOG_NET);`

src/browser/wisp_network.js
```diff
@@ -0,0 +1,571 @@
...
+            flags |= 0x0180; // Recursion
+            // flags |= 0x0400; Authoritative
+            for(let i = 0; i < packet.dns.questions.length; ++i) {
+                let q = packet.dns.questions[i];
+                // Sometimes this results in no answer at all, like if dohdns crashes, but UDP is unreliable in nature so this shouldn't be that big of an issue
+                const res = await dohdns(q);
+                switch(q.type){
+                    case 1: // A recrod
+
+                        // for (const ans in res.Answer) {    // v86 DNS server crashes and burns with multiple answers, not quite sure why
```
copy, line 347: > v86 DNS server crashes and burns with multiple answers, not quite sure why

What dns server does this refer to? The one in the alpine profile?

I think it's more likely a bug in make_packet

src/browser/wisp_network.js
```diff
@@ -0,0 +1,571 @@
...
+                connections[streamID].congested = false;
+            }
+
+            break;
+        case 4: // CLOSE
+            if(connections[streamID])
+                connections[streamID].closeCallback(view.getUint8(5));
+            delete connections[streamID];
+            break;
+    }
```
copy, line 55: There should be a catch-all case with a warning message. (also in the `switch(frameObj.type) {` below)

src/browser/wisp_network.js
```diff
@@ -0,0 +1,571 @@
...
+    if(/^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/.test(ipaddress)) {
+      return true;
+    }
+    return false;
+}
+
+// DNS over HTTPS fetch, recursively fetch the A record until the first result is an IPv4
+async function dohdns(q) {
+    const prefered_fetch = (window.anura?.net?.fetch) || fetch;
+    const req = await prefered_fetch(`https://dns.google/resolve?name=${q.name.join(".")}&type=${q.type}`);
```
copy, line 185:
```suggestion
    const req = await prefered_fetch(`${DOH_SERVER}?name=${q.name.join(".")}&type=${q.type}`);
```

src/browser/wisp_network.js
```diff
@@ -0,0 +1,571 @@
...
+        this.vm_mac = new Uint8Array(mac.split(":").map(function(x) { return parseInt(x, 16); }));
+    }, this);
+    this.bus.register("net" + this.id + "-send", function(data)
+    {
+        this.send(data);
+    }, this);
+}
+
+WispNetworkAdapter.prototype.destroy = function()
+{
```
copy, line 171: The websocket should be closed here

copy (commented):
Thanks, this looks very useful.

Could you do one pass to make the style more like in the rest of the codebase? In particular:

- `{}` after if and newline betweem `)` and `{`
- snake_case for identifiers

There should be at least one test case, see [`tests/devices/fetch_network.js`](https://github.com/copy/v86/blob/master/tests/devices/fetch_network.js) for a reference. I'd suggest not using mocks and testing against the real wisp server, if that's okay. It's not necessary to run the test automatically on ci.

Also, it would be nice if you could add a short section at the bottom of [`docs/networking.md`](https://github.com/copy/v86/blob/master/docs/networking.md)

ProgrammerIn-wonderland:
Yeah I can work on that

src/browser/wisp_network.js
ProgrammerIn-wonderland: That was an accident I left from a seperate project, will fix

src/browser/wisp_network.js
```diff
@@ -0,0 +1,571 @@
...
+            streamID: this.streamID,
+            data: data
+        });
+    }
+};
+
+if(typeof module !== "undefined" && typeof module.exports !== "undefined")
+{
+    module.exports["WispNetworkAdapter"] = WispNetworkAdapter;
+}
```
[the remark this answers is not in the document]
ProgrammerIn-wonderland: Was referring to the internal DNS server in fetch_networking.js, if I pushed all the answers it wasn't very happy

src/browser/wisp_network.js
ProgrammerIn-wonderland: Was referring to the internal DNS server in fetch_networking.js, if I pushed all the answers it wasn't very happy, it most likely is a make_packet bug I agree

src/browser/wisp_network.js
ProgrammerIn-wonderland: where should DOH_SERVER be defined?

copy: At the top of this file

ProgrammerIn-wonderland:
further testing reveals that not all DNS over HTTPS servers are so kind to cross origin requests, dns.google worked but 1.1.1.1 gives a 400 error if its not on the same origin. Quad9, adguard, and the others I tried just fails with a cors error. I'm not as sure if it should be a configurable option since dns.google seems to be the only one friendly to cross origin

ProgrammerIn-wonderland:
I'm also running into this weird bug where 
```js
    const req = await fetch(`https://dns.google/resolve?name=${"google.com."}&type=${1}`,  {headers: [["accept", "application/dns-json"]]});
const res = await req.json()
```
will give res.Answer most of the time but suddenly it will disappear and not come back until you clear cookies, trying to figure that out as well

ProgrammerIn-wonderland:
I'm wondering if using corsproxy.io to fix this issue would be appropriate, though it would get rid of the main point of DNS over HTTPS

SuperMaxusa:
> 1.1.1.1 gives a 400 error if its not on the same origin

From https://developers.cloudflare.com/1.1.1.1/encryption/dns-over-https/make-api-requests/#return-codes: 
> `400` DNS query not specified or too small.

Cloudflare uses a [RFC 8488](https://datatracker.ietf.org/doc/html/rfc8484#section-4.1.1)-compatible API which different then [Google DNS JSON API](https://developers.google.com/speed/public-dns/docs/doh/json), but Google responds with `application/dns-message` on `/dns-query` endpoint.  
Anyways, they are same of functionallity (and works with CORS too).

```js
> google = await fetch("https://dns.google/resolve?name=example.com", { headers: {"Accept": "application/dns-json"}});
> await google.json()
{
  Status: 0,
  TC: false,
  RD: true,
  ...
}

> cloudflare = await fetch("https://1.1.1.1/dns-query?name=example.com", { headers: {"Accept": "application/dns-json"}});
> await cloudflare.json()
{
  Status: 0,
  TC: false,
  RD: true,
  ...
}
```

ProgrammerIn-wonderland:
I tried that and was complaining to one of my friends about how it feels like it should work but it doesn't. He advised me to check my network tab and pointed out the serviceworkers could intercept cross origin requests. As it turned out, that was my issue. None of the issues I described with Answer not happening existed, that was a testing error on my part, apologies

copy:
@ProgrammerIn-wonderland Now that I think about it, couldn't dns requests be handled like regular wisp traffic? (that is, delete the special dns handling in this code and let it pass through as regular udp traffic)?

ProgrammerIn-wonderland:
its possible, its just also common practice to disable UDP on wisp servers to disincentivize torrenting, I believe the public wisp.mercurywork.shop is one of these

ProgrammerIn-wonderland:
I think other than the tests, the TCP + DOH is ready to go now

ProgrammerIn-wonderland:
It seems to test Wisp network from nodejs I'd need the https://www.npmjs.com/package/ws npm package

ProgrammerIn-wonderland:
committed the test though, it does pass

ProgrammerIn-wonderland:
I asked the hoster of wisp.mercurywork.shop, he is somewhat concerned about bandwidth usage since its hosted on a server with limited bandwidth monthly. Is it possible for you to be able to host one? If not I can find someone else who doesn't have metered bandwidth. Hosting a wisp server is a lot easier than hosting a jor1k relay since it doesn't need docker or dnsmasq, just a up to date version of nodejs or python (or rust) depending on which server you choose

basicer:
This is super cool.

I wonder if a wisp server could be built out of a cloudflare worker.

For DNS, if google can give a message already encoded in the DNS wire format, why not just send that message right back down.

It seems like we should refactor out the TCP/Ethernet/UDP/etc stack from fetch instead of copying a bunch of the code.  I'm pretty sure there's bound to be bugs / things to improve, especially in TCP and it would be a pain to have to do that in two places.

copy:
> edit 2: A concern I still have for UDP is DHCP, the server probably wouldn't be too happy with handling that, should there be certain ports handled internally and others sent?

Yes, dhcp (and arp) should be handled internally and dns should be sent off to whisp.

> there is extremely recent nodejs WebSocket support but it would require you to have a very recent (made after 2023, so node 20.10.xx) nodejs version to run the test, I'm not sure if thats appropriate or acceptable for this project https://github.com/nodejs/node/pull/49830

Yes, depending on node 20.x is fine (it's the current LTS).

> I asked the hoster of wisp.mercurywork.shop, he is somewhat concerned about bandwidth usage since its hosted on a server with limited bandwidth monthly. Is it possible for you to be able to host one? Hosting a wisp server is a lot easier than hosting a jor1k relay since it doesn't need docker or dnsmasq, just a up to date version of nodejs or python (or rust) depending on which server you choose

If that's a concern, we could avoid mentioning wisp.mercurywork.shop, so most users wouldn't use it. I don't have any plans of hosting a proxy service myself, but if anybody else does I would add those to a dropdown on the website.

They are also free to apply stricter throttling to v86 users.

ProgrammerIn-wonderland:
> This is super cool.
> 
> I wonder if a wisp server could be built out of a cloudflare worker.
> 
> For DNS, if google can give a message already encoded in the DNS wire format, why not just send that message right back down.
> 
> It seems like we should refactor out the TCP/Ethernet/UDP/etc stack from fetch instead of copying a bunch of the code. I'm pretty sure there's bound to be bugs / things to improve, especially in TCP and it would be a pain to have to do that in two places.

We actually do have an impl that works on cloudflare workers, the main thing is the usage limits can make it hard to work with https://github.com/MercuryWorkshop/wisp-server-workers, we ran into limits like cf not being able to handle more than 6 concurrent connections iirc

ProgrammerIn-wonderland:
undrafting since I think its fairly ready now

src/browser/wisp_network.js
```diff
@@ -0,0 +1,570 @@
...
+    } else {
+        dbg_log("DNS Server returned error code: " + await req.text(), LOG_NET);
+    }
+}
+
+/**
+ * @param {Uint8Array} data
+ */
+WispNetworkAdapter.prototype.send = function(data)
+{
```
basicer, line 234: Its' mainly this logic that I worry should be refactored out.

SuperMaxusa:
> is there a way to update the nodejs?

GitHub CI has a v20.15.1 in cached dependencies: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#nodejs

And I think we could do a check version or WebSocket object at the beginning of the script like that:
```js
// checking the major version
if (parseInt(process.versions.node.slice(0, 2)) < 20) {
	console.error("This version of Node.JS doesn't support running this test.");
	console.error("Minimal version: v22.x.x (or v20.10-21.x with --experimental-websocket flag)");
	process.exit(1);
}

// or websocket availability
if (typeof WebSocket === "undefined") {
	console.error("Your Node.JS doesn't support WebSocket.");
	console.error("If Node.JS version is v20.10-21.x, make sure the script was run with --experimental-websocket flag.");
	process.exit(1);
}
```

ProgrammerIn-wonderland:
I added the --expirimental-websocket flag I'm hoping that exposes the object

src/browser/wisp_network.js
ProgrammerIn-wonderland: What do you have in mind? In general I think the way the wisp client will handle TCP and UDP too differently for that function to be shared between fetch_network and wisp_network, but I'm open to ideas on how it should be organized

ProgrammerIn-wonderland: it looks similar at first glance but if you look a bit more at packet.tcp.syn you'll see that I modified a bit of what goes into this.tcp_conn\[tuple\] like adding stream_id and send_wisp_frame, those weren't originally a part of the function in the fetch_network I believe

ProgrammerIn-wonderland:
also a slight update on wisp-server-workers, I gave some misleading information but its not 6 total streams, its 6 streams per wisp connection (each v86 Emulator connected to wisp-server-worker would get 6 concurrent TCP streams)

src/browser/wisp_network.js
basicer: I think a good place to start would be to extract the duplicate logic for things like dhcp and ntp into functions in fetch-networking and then just call them in the handler.

ProgrammerIn-wonderland: I can do that, should I make a new file for this?

ading2210:
> If that's a concern, we could avoid mentioning wisp.mercurywork.shop, so most users wouldn't use it. I don't have any plans of hosting a proxy service myself, but if anybody else does I would add those to a dropdown on the website.

I'm honestly fine with wisp.mercurywork.shop being the default, but I would have to decrease the rate limits so I don't hit the 10TB free bandwidth limit that Oracle Cloud gives me. Perhaps v86 users could use an alternate endpoint (`wss://wisp.mercurywork.shop/v86/`) with low rate limits (something like 300kb/s?) so that I don't have to throttle everyone else.

ProgrammerIn-wonderland:
I'd also probably encourage you to keep a global counter that shuts down all endpoints at 9.5TB usage to avoid the oracle overage bill

src/browser/wisp_network.js
basicer: Yeah I was debating on putting all the protocol utility stuff it its own file independent of the "clever" fetch impersonation bits, but ultimately decided against.  If more than one file is going to use them it might make sense though.

SuperMaxusa:
I submitted a PR for updating Node.js: #1100 

With Node.js v20.15.1 on local machine, `Test #5: Curl example.org` fails:
```
Preparing test #5: Curl example.org
Starting test #5: Curl example.org
Capturing...
    Captured: wget -T 10 -O - example.org
    Captured: echo -e done\\texample.org
    Captured: ~% wget -T 10 -O - example.org
    Captured: Connecting to example.org (93.184.215.14:80)
node:internal/event_target:1094
  process.nextTick(() => { throw err; });
                           ^

RangeError: Offset is outside the bounds of the DataView
    at DataView.getUint32 (<anonymous>)
    at $WispNetworkAdapter$$.process_incoming_wisp_frame (.../v86-wisp-net/build/libv86-debug.js:10241:104)
    at wispws.onmessage (.../v86-wisp-net/build/libv86-debug.js:10309:12)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:820:20)
    at WebSocket.dispatchEvent (node:internal/event_target:755:26)
    at fireEvent (node:internal/deps/undici/undici:10966:14)
    at websocketMessageReceived (node:internal/deps/undici/undici:10988:7)
    at ByteParser.run (node:internal/deps/undici/undici:11439:17)
    at ByteParser._write (node:internal/deps/undici/undici:11315:14)
    at writeOrBuffer (node:internal/streams/writable:564:12)

Node.js v20.15.1

```

On v20.16.x and v22.5.1 the test passes successfully.

src/browser/wisp_network.js
```diff
@@ -0,0 +1,567 @@
...
+                }
+
+                break;
+            case 4: // CLOSE
+                if(this.connections[stream_id])
+                    this.connections[stream_id].close_callback(view.getUint8(5));
+                delete this.connections[stream_id];
+                break;
+            case 5: // PROTOEXT
+                // Not responding, this is wisp v1 client not wisp v2;
```
SuperMaxusa, line 65: Perhaps we could add a `dbg_log()` here with message like "Wisp server returned PROTOEXT packet that is not supported by this version of client protocol"? (or catch in `default` case as unknown)

ProgrammerIn-wonderland: It isn't technically unknown so I'm hesitant to mark it as that but if there's no response to a protoext packet the server is supposed to downgrade to wisp V1. I think a message like "got a wisp V2 upgrade request, ignoring" could work

ProgrammerIn-wonderland:
Is devicestest frozen in the CI? All the tests seemed to have passed but it's still counting minutes

src/browser/wisp_network.js
ProgrammerIn-wonderland: @copy thoughts before I (or if I should)  split the shared "utility" functions into net-utils.js?

copy:
> Is devicestest frozen in the CI? All the tests seemed to have passed but it's still counting minutes.
I may just add a process.exit(0) that runs after the all tests passed message, but I wonder what's keeping the test running

It means there are still resources that can't be garbage-collected. Try setting the websocket to null or removing the event handlers from the websocket in `destruct`.

src/browser/wisp_network.js
ProgrammerIn-wonderland: > Yeah I was debating on putting all the protocol utility stuff it its own file independent of the "clever" fetch impersonation bits, but ultimately decided against. If more than one file is going to use them it might make sense though.

I'm refactoring both right now and I see FetchNetworkAdapter.prototype.tcp_connect but it seems to be fully unused, tests still run perfectly without it too, should I get rid of it? or are there future plans for it

ProgrammerIn-wonderland:
> > Is devicestest frozen in the CI? All the tests seemed to have passed but it's still counting minutes.
> > I may just add a process.exit(0) that runs after the all tests passed message, but I wonder what's keeping the test running
> 
> It means there are still resources that can't be garbage-collected. Try setting the websocket to null or removing the event handlers from the websocket in `destruct`.

it seems my destructor is never actually being called 
image

ProgrammerIn-wonderland:
I see why, I copied most of my test from fetch_network and it never called emulator.destroy, this is fixed by just adding a emulator.destroy after emulator.stop
Screenshot_20240802_102344

ProgrammerIn-wonderland:
figured it out! onclose() was opening a new websocket, had to set that event handler to null, thanks copy

ProgrammerIn-wonderland:
@basicer moved the common/shared code to a different file

ProgrammerIn-wonderland:
> For DNS, if google can give a message already encoded in the DNS wire format, why not just send that message right back down.

I've been trying for a while now to use the DNS wire format, cant seem to get it working, not sure if its within my abilities sorry

ProgrammerIn-wonderland:
latest commit fixes the dhcpcd issue in #1091

basicer:
> latest commit fixes the dhcpcd issue in #1091

Is this the correct fix? Reading the issue it seems like the problem is that the fake gateway gives an arp reply of its mac for all ips in subnet.  Instead shouldn't it just stay quiet for ips in the dhcp client ip?

ProgrammerIn-wonderland:
> > latest commit fixes the dhcpcd issue in #1091
> 
> Is this the correct fix? Reading the issue it seems like the problem is that the fake gateway gives an arp reply of its mac for all ips in subnet. Instead shouldn't it just stay quiet for ips in the dhcp client ip?

I wasn't fully sure of how ARP handled a response when the IP wasn't assigned, so I just replied back with the source's mac address as the one assigned when it asked for 192.168.86.100, I'm not sure if this is the correct way to handle it (it most likely isn't) but it made dhcpcd happy, I can try just not replying and see how dhcpcd likes that

ProgrammerIn-wonderland:
Screenshot_20240802_214613
yeah just ignoring works fine too for dhcpcd, I'll get rid of that part then

copy (commented):
Thanks, this is taking shape :-)

I find the use of callbacks in TCPConnection a bit confusing. It would be nice if you could make TCPConnection free of http/whisp handling, and instead make its public api (process, accept and connect) return packets for the upper layer (http/wisp) and the lower layer (ethernet).

SuperMaxusa:
> I've been trying for a while now to use the DNS wire format, cant seem to get it working

Works for me `nslookup -type=<a/aaaa/mx/cname/text> example.com`. 

By the way, you can try `POST` raw packet with `content-type: application/dns-message`, but with `nslookup example.com` it's stuck on `*** Can't find example.com: No answer` after getting DNS A response:

```diff
--- a/src/browser/wisp_network.js
+++ b/src/browser/wisp_network.js
@@ -266,9 +266,9 @@ WispNetworkAdapter.prototype.send = function(data)
                 dest: packet.ipv4.src,
             };
             reply.udp = { sport: 53, dport: packet.udp.sport };
-            const fetchURL = "https://cloudflare-dns.com/dns-query?dns=" + await bufferToBase64(packet.udp.data);
+            const fetchURL = "https://cloudflare-dns.com/dns-query";
             console.log("fetching " + fetchURL);
-            const result = await ((await fetch(fetchURL, {headers: [["accept", "application/dns-message"]]})).arrayBuffer())
+            const result = await ((await fetch(fetchURL, {method: "POST", headers: [["accept", "application/dns-message"], ["content-type", "application/dns-message"]], body: new Blob([packet.udp.data])})).arrayBuffer());
             console.log(result);
             reply.udp.data = new Uint8Array(result);
             console.log(reply.udp.data)

```

ProgrammerIn-wonderland:
> Thanks, this is taking shape :-)
> 
> I find the use of callbacks in TCPConnection a bit confusing. It would be nice if you could make TCPConnection free of http/whisp handling, and instead make its public api (process, accept and connect) return packets for the upper layer (http/wisp) and the lower layer (ethernet).

I can do that, the main reason I initially had wisp specific callbacks in TCP connection was because fetch network did the same

ProgrammerIn-wonderland:
> > I've been trying for a while now to use the DNS wire format, cant seem to get it working
> 
> Works for me `nslookup -type=<a/aaaa/mx/cname/text> example.com`. 
> 
> By the way, you can try `POST` raw packet with `content-type: application/dns-message`, but with `nslookup example.com` it's stuck on `*** Can't find example.com: No answer` after getting DNS A response:
> 
> ```diff
> --- a/src/browser/wisp_network.js
> +++ b/src/browser/wisp_network.js
> @@ -266,9 +266,9 @@ WispNetworkAdapter.prototype.send = function(data)
>                  dest: packet.ipv4.src,
>              };
>              reply.udp = { sport: 53, dport: packet.udp.sport };
> -            const fetchURL = "https://cloudflare-dns.com/dns-query?dns=" + await bufferToBase64(packet.udp.data);
> +            const fetchURL = "https://cloudflare-dns.com/dns-query";
>              console.log("fetching " + fetchURL);
> -            const result = await ((await fetch(fetchURL, {headers: [["accept", "application/dns-message"]]})).arrayBuffer())
> +            const result = await ((await fetch(fetchURL, {method: "POST", headers: [["accept", "application/dns-message"], ["content-type", "application/dns-message"]], body: new Blob([packet.udp.data])})).arrayBuffer());
>              console.log(result);
>              reply.udp.data = new Uint8Array(result);
>              console.log(reply.udp.data)
> 
> ```

I was somehow getting the same issue you describe but with the normal GET handler, I'll see if I can resolve it later today or early tomorrow

src/browser/wisp_network.js
```diff
@@ -0,0 +1,353 @@
...
+WispNetworkAdapter.prototype.receive = function(data)
+{
+    this.bus.send("net" + this.id + "-receive", new Uint8Array(data));
+};
+
+/**
+ *
+ * @param {Uint8Array} data
+ */
+TCPConnection.prototype.on_data_wisp = async function(data) {
```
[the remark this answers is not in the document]
ProgrammerIn-wonderland: Should I do the same for fetch networking?

ProgrammerIn-wonderland:
I do hope to get DNS over wire working since windows seems to have a problem with the current DNS setup

src/browser/wisp_network.js
```diff
@@ -0,0 +1,353 @@
...
+        this.wispws = new WebSocket(wisp_url.replace("wisp://", "ws://").replace("wisps://", "wss://"));
+        this.wispws.binaryType = "arraybuffer";
+        this.wispws.onmessage = (event) => {
+            this.process_incoming_wisp_frame(new Uint8Array(event.data));
+        };
+        this.wispws.onclose = () => {
+            register_ws();
+        };
+    };
+    register_ws();
```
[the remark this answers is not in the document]
ProgrammerIn-wonderland: should I also move the initial state to the prototype?

ProgrammerIn-wonderland:
I moved some of the initial state to prototypes too

SuperMaxusa:
> I was somehow getting the same issue you describe but with the normal GET handler

Not sure but that `nslookup` in BusyBox (if you are testing on Alpine Linux) might cause this because on newer versions (`v1.28.4` - works properly, `v.1.3x` - has a bug) it sends all DNS query packets with the same transaction ID: http://lists.busybox.net/pipermail/busybox/2023-May/090318.html

wireshark-dns

ProgrammerIn-wonderland:
> > I was somehow getting the same issue you describe but with the normal GET handler
> 
> Not sure but `nslookup` in BusyBox (if you are testing on Alpine Linux) might cause this because on newer versions (`v1.28.4` - works properly, `v.1.3x` - has a bug) it sends all DNS query packets with the same transaction ID: http://lists.busybox.net/pipermail/busybox/2023-May/090318.html
> 
> wireshark-dns

I was (lazily) just using curl http://google.com in archlinux and noticing it failed at DNS resolution while direct IP access worked

SuperMaxusa:
> using curl http://google.com in archlinux and noticing it failed at DNS resolution

Have you tried run with verbose log (`curl -v google.com`) or trace (`curl --trace - google.com`)?

ProgrammerIn-wonderland:
not yet, I'll reapply the patch and try it out

ProgrammerIn-wonderland:
I was wrong @SuperMaxusa the way I was testing was make devices-test but it was failing so I assumed something was broken, but archlinux itself actually works fine
Screenshot_20240803_160504-1

which makes me wonder if it was a nodejs issue? I'm genuinely unsure why the test fails but the actual archlinux emu works. Or I wonder if its the aformentioned busybox dns bug

ProgrammerIn-wonderland:
Screenshot_20240803_162216
nodejs make tests issue

ProgrammerIn-wonderland:
[failed.log](https://github.com/user-attachments/files/16482797/failed.log)

ProgrammerIn-wonderland:
Screenshot_20240803_162216-1
windows actually likes the DNS this time, the older solution broke windows

ProgrammerIn-wonderland:
Screenshot_20240803_162216-3
buildroot also works fine with this solution, it seems the issue is just with the test and I was misleading myself into thinking it was defective

ProgrammerIn-wonderland:
also, updated diff for the latest commit (the old one wont merge since the latest refactor)
[updated.diff.txt](https://github.com/user-attachments/files/16482833/updated.diff.txt)

ProgrammerIn-wonderland:
It seems other than the test, the DNS over https wire protocol adapter is much better 🥴

ProgrammerIn-wonderland:
image
testing with linux4 (same image used in the test) more or less shows me that my issue seemed to be nodejs only

SuperMaxusa:
> nodejs make tests issue (+ failed.log)

Are you sure that it is DNS packets and then what you are sending to the DoH? Because when I decode one of them, the length is 32KB.

ProgrammerIn-wonderland:
that seems a bit too big, let me recheck

ProgrammerIn-wonderland:
I made my version use POST and it works now? (it also removes the need for the base64 converter). I guess my issue was there

ProgrammerIn-wonderland:
also linux4 still seems to work with my POST based version, without the issue you described
Screenshot_20240803_162216-5

ProgrammerIn-wonderland:
thanks for the POST tip, it also saves time between converting to base64 when its not necessary

ProgrammerIn-wonderland:
I'm thinking as for finishing touches, would it be a good idea to allow the user to specify the DNS over HTTPS server through the contructor? (and maybe the corsproxy for fetch since this PR refactors both)

SuperMaxusa:
Can you try on https://copy.sh/v86?profile=buildroot too? 
Linux 4 has Buildroot v1.28.4 which (from my tests) does not have this problem.

ProgrammerIn-wonderland:
Screenshot_20240803_162216-6
works as well

ProgrammerIn-wonderland:
this was an issue since the beginning but now that windows98 networking works perfectly I wonder why windows 2000 dhcp is still stalled
Screenshot_20240803_162216-7

ProgrammerIn-wonderland:
Screenshot_20240803_162216-8
the issue is just with dhcp too since if you specify the IP manually it doesn't have an issue

ProgrammerIn-wonderland:
haiku, reactos, win98, archlinux, linux4, buildroot all seem to work now, its really just windows 2000 remaining

ProgrammerIn-wonderland:
okay this is bizzare, it works fine with `https://copy.sh/v86/?profile=windows2000-boot`, just not snapshot windows 2k? I am not very sure why

ProgrammerIn-wonderland:
image
this is dhcp, no static trickery going on here

ProgrammerIn-wonderland:
issue seems to be with the snapshot itself, the current snapshot works fine with widgetry relay but not with fetch or wisp_networking, if I take a new snapshot it works with both

ProgrammerIn-wonderland:
[win2k-networking.bin.bz.txt](https://github.com/user-attachments/files/16485375/win2k-networking.bin.bz.txt)
here is a new state that works (I called it .txt to upload to github but remove that, its a xz compressed file)

src/browser/wisp_network.js
```diff
@@ -0,0 +1,320 @@
...
+}
+
+WispNetworkAdapter.prototype.register_ws = function (wisp_url) {
+    this.wispws = new WebSocket(wisp_url.replace("wisp://", "ws://").replace("wisps://", "wss://"));
+    this.wispws.binaryType = "arraybuffer";
+    this.wispws.onmessage = (event) => {
+        this.process_incoming_wisp_frame(new Uint8Array(event.data));
+    };
+    this.wispws.onclose = () => {
+        this.register_ws(wisp_url);
```
copy, line 41: This should probably have a `setTimeout` of 10 or so seconds, just so we don't hammer the server if there are problems with the connection.

src/browser/wisp_network.js
```diff
@@ -0,0 +1,320 @@
...
+//
+// FrameObj.type -- DATA
+//      FrameObj.data (Uint8Array)
+//
+// FrameObj.type -- CLOSE
+//      FrameObj.reason (number)
+//
+// FrameObj.type -- RESIZE
+//      FrameObj.cols (number)
+//      FrameObj.rows (number)
```
ading2210, line 123: This comment seems to be left over from an unrelated project.

src/browser/wisp_network.js
```diff
@@ -0,0 +1,320 @@
...
+    this.wispws.onmessage = (event) => {
+        this.process_incoming_wisp_frame(new Uint8Array(event.data));
+    };
+    this.wispws.onclose = () => {
+        this.register_ws(wisp_url);
+    };
+};
+WispNetworkAdapter.prototype.last_stream = 1;
+
+WispNetworkAdapter.prototype.connections = {0: {congestion: 0}};
```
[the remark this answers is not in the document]
ProgrammerIn-wonderland: To be honest I had put it there for the typescript language server's auto complete feature, but I can move it to the constructor

copy:
This needs a rebase after #1113 and addressing the last two suggestions, then it should be good to merge.

ProgrammerIn-wonderland:
:+1:, I was busy this week due to standardized testing so I didn't have time to rebase, but I should be able to handle it soon

ProgrammerIn-wonderland:
accident

ProgrammerIn-wonderland:
It seems there some parts i cant reuse from fake_network this handle_fake_networking uses the fake DNS server, but other than that function I believe I can make it mostly work

ProgrammerIn-wonderland:
should be good now I think

copy.sh/v86/index.html
SuperMaxusa, line 1: Was this file created by mistake?

ProgrammerIn-wonderland: Yes

src/browser/wisp_network.js
```diff
@@ -0,0 +1,302 @@
...
+
+        this.tcp_conn[tuple].process(packet);
+    }
+
+    if(packet.arp && packet.arp.oper === 1 && packet.arp.ptype === ETHERTYPE_IPV4) {
+        arp_whohas(packet, this);
+    }
+
+    if(packet.dns) {
+        // TODO: remove when this wisp client supports udp
```
copy, line 263: FYI, I would like to avoid DOH (behind a switch that is disabled by default), in order to avoid leaking hostnames to cloudflare. However, since this wisp implementation doesn't have UDP yet, I'll leave it with this comment for now.

src/browser/wisp_network.js
```diff
@@ -281,28 +273,22 @@ WispNetworkAdapter.prototype.send = function(data)
...
         handle_fake_ntp(packet, this);
         return;
     }
 
-    // ICMP Ping
-    if(packet.icmp && packet.icmp.type === 8) {
-        handle_fake_ping(packet, this);
-        return;
-    }
-
```
copy: I removed the fake ping implementation here, it's not very useful (and possibly confusing) when combined with a real networking proxy.

copy:
> > If that's a concern, we could avoid mentioning wisp.mercurywork.shop, so most users wouldn't use it. I don't have any plans of hosting a proxy service myself, but if anybody else does I would add those to a dropdown on the website.
> 
> I'm honestly fine with wisp.mercurywork.shop being the default, but I would have to decrease the rate limits so I don't hit the 10TB free bandwidth limit that Oracle Cloud gives me. Perhaps v86 users could use an alternate endpoint (`wss://wisp.mercurywork.shop/v86/`) with low rate limits (something like 300kb/s?) so that I don't have to throttle everyone else.

Sure. We could also add a custom header (then other wisp relays could use apply throttling too). In any case, this relay is currently only used for tests, which causes very little traffic.

copy:
Thanks everyone!

src/browser/wisp_network.js
ProgrammerIn-wonderland: Understandable, I'm a bit busy this week but I'll see if I can make out time in the near future to add in UDP

PR #1123, opened 2024-08-14
Add VGA graphical text mode

### LINKED ISSUE #1098: Support for other code pages in text terminal?
I need to present the information in the text terminal in other language than English. Is there any support for the text terminal to display letters with diacritics? If not, I am going to implement it ASAP. I miss this feature a lot.

chschnell:
Added alternate text mode "graphical text mode" that uses internal VGA fonts and settings in text mode with the canvas that is otherwise used in graphical mode.

Added new configuration option `screen_options` to V86 constructor, which is needed for two screen-related options:

- `use_graphical_text`: Boolean, enable VGAScreen's graphical text mode if true
- `disable_autoscale`: Boolean, disable ScreenAdapter's auto-scale feature if true

VGAScreen's current behaviour is unmodified unless `use_graphical_text` is set to true, this PR is intended to be a non-breaking change.

Added new bus event `screen-settings-changed`, it reports details about the current video mode settings when graphical text mode is enabled, event arguments:

- in text mode: `[false, txt_width, text_height, font_width, font_height]`
- in graphical mode: `[true, gfx_width, gfx_height, virtual_width, virtual_height]`

Added new VGAScreen function `grab_text_content()` that returns an array of strings containing the current text screen content (in CP437).

**Supported VGA text mode features:**

- monochrome and 16 colour modes
- 40 and 80 columns
- 25, 28 and 50 rows
- font height 1-32 pixel
- font width 8, 9 or 16 pixel
- cursor emulation
- blinking text
- font page A/B (512 character mode)
- line graphics enable (continuous horizontal line drawing characters)

**Usage example for new optional setting `screen_options`:**

    window.emulator = new V86({
        wasm_path: "v86.wasm",
        memory_size: 16 * 1024 * 1024,
        vga_memory_size: 2 * 1024 * 1024,
        screen_container: document.getElementById("screen_container"),
        screen_options: {
            use_graphical_text: true,
            disable_autoscale:  true
        },
        bios: { 
            url: "seabios.bin" 
        },
        vga_bios: { 
            url: "vgabios.bin" 
        },
        fda: { 
            url: "freedos13.img" 
        },
        autostart: true, 
    });

**Open issues:**

- split screen (looking for example)
- 43 rows is refused by both MS-DOS and FreeDOS
- Monochrome text-mode is refused by MS-DOS

This PR should resolve #1098 (more details there).


src/vga.js
```diff
@@ -65,7 +65,7 @@ var VGA_HOST_MEMORY_SPACE_SIZE = Uint32Array.from([
  * @param {BusConnector} bus
  * @param {number} vga_memory_size
  */
-function VGAScreen(cpu, bus, vga_memory_size)
+function VGAScreen(cpu, bus, vga_memory_size, screen_options = undefined)
```
copy, line 68: I think the code would be a bit simpler if you force this to be an object. (VGAScreen is internal to v86, so we can change its interface arbitrarily).

src/cpu.js
```diff
@@ -929,7 +929,8 @@ CPU.prototype.init = function(settings, device_bus)
         this.devices.dma = new DMA(this);
 
         this.devices.vga = new VGAScreen(this, device_bus,
-                settings.vga_memory_size || 8 * 1024 * 1024);
+                settings.vga_memory_size || 8 * 1024 * 1024,
+                settings.screen_options);
```
copy, line 933:
```suggestion
                settings.screen_options || {});
```

src/vga.js
```diff
@@ -409,7 +408,66 @@ function VGAScreen(cpu, bus, vga_memory_size)
...
 
+    if(!! screen_options?.use_graphical_text)
+    {
+        this.graphical_text = new GraphicalText(bus, me);
+    }
+    this.text_to_graphical_mode_flag = false;   // flag that indicates a previous text-to-graphical mode transition to screen_fill_buffer()
+
     cpu.devices.pci.register_device(this);
+
+    setTimeout(() => { me.switch_screen_mode(this.graphical_mode); }, 0);
```
copy, line 419:
```suggestion
    setTimeout(() => { this.switch_screen_mode(this.graphical_mode); }, 0);
```

(`me` is a leftover from before arrow functions)

src/vga.js
```diff
@@ -409,7 +408,66 @@ function VGAScreen(cpu, bus, vga_memory_size)
...
+
+    if(!graphical_mode && this.graphical_text)
+    {
+        this.graphical_text.resize_canvas();
+    }
+
+    this.bus.send("screen-set-mode", graphical_mode || this.graphical_text);
+}
+
+VGAScreen.prototype.grab_text_content = function(keep_whitespace)
```
copy, line 446: This looks unused.

src/browser/screen.js
```diff
@@ -6,7 +6,7 @@
  *
  * @param {BusConnector} bus
  */
-function ScreenAdapter(screen_container, bus)
+function ScreenAdapter(screen_container, bus, screen_options = undefined)
```
copy, line 9: I think the code would be a bit simpler if you force this to be an object.

copy (commented):
Thanks!

I've left a few stylistic comments. However, looking at the bigger picture, many of the changes in vga.js look similar to this:

```diff
-    this.bus.send("screen-update-cursor-scanline", [start, end, visible]);
+    if(this.graphical_text)
+    {
+        this.graphical_text.set_cursor_attr(start, end, visible);
+    }
+    else
+    {
+        this.bus.send("screen-update-cursor-scanline", [start, end, visible]);
+    }
```

Which indicates that it might be better to merge src/vga_text.js into src/browser/screen.js. It would also be nice to reuse the `screen-put-char` machinery.

src/vga.js
chschnell: Patched it, I wasn't sure whether VGAScreen was considered internal.

src/cpu.js
chschnell: Patched.

src/vga.js
chschnell: Sure!

src/browser/screen.js
chschnell: Patched.

src/vga.js
chschnell: Early on you mentioned the current ability to select text from the DIV using the mouse.

This is my poor-man's replacement for this, but I left out the accompanying method in V86's user interface, currently I call it using v86.cpu.devices.vga.grab_text_content() in my V86 embedding.

I could add mouse-controlled text-selection though, roughly like this:

- Text selection is simply a linear copy from the 16 bit VGA text buffer, starting at (start-row, start-col) to (end-row, end-col).
- If any character within the current selection is changed, selection gets cancelled and reset to nothing.
- Swap forground and background color when rendering selected text.
- Immediately copy to clipboard when user changes selected text range.
- Any events from mouse and keyboard related to text selection must be hidden from the virtualized OS. Not sure yet how to modify MouseAdapter and KeyboardAdapter here.

But someone would have to establish a mouse-button + key combination that triggers text selection in graphical text mode. For example, I use the DOS mouse driver in FreeDOS, and I can select/copy text *within* FreeDOS's EDIT.EXE using my mouse. So the normal left mouse button is not an option, text-based applications may use it in graphical text mode. IIRC already DOS supported mice with 3 buttons, and certainly the 104/105 keyboards.

Or, make text selection in graphical text mode a special state that has to be triggered through the HTML user interface, naively like a button "Begin Text Selection" which changes its label to "End Text Selection" when clicked, and during that state the left mouse button would be used for text selection normally and all mouse events would be hidden from the virtualized OS. I'd prefer that.

What do you think?

chschnell:
> Thanks!
> 
> I've left a few stylistic comments. However, looking at the bigger picture, many of the changes in vga.js look similar to this:
> 
> ```diff
> -    this.bus.send("screen-update-cursor-scanline", [start, end, visible]);
> +    if(this.graphical_text)
> +    {
> +        this.graphical_text.set_cursor_attr(start, end, visible);
> +    }
> +    else
> +    {
> +        this.bus.send("screen-update-cursor-scanline", [start, end, visible]);
> +    }
> ```
> 
> Which indicates that it might be better to merge src/vga_text.js into src/browser/screen.js. It would also be nice to reuse the `screen-put-char` machinery.

You're welcome, and thank you very much for taking the time for a close look at all this, I really appreciate that.

I agree that the new code should be merged into existing classes, it is structured that way (minimal changes to existing code) to best oversee how.

Here's where I ended up: In graphical text mode, the entire DIV-based text implementation in ScreenAdapter is currently dead on purpose as I did not see any use for it, the code is 100% "living off the land" in VGAScreen, specifically without the 2nd text buffer ScreenAdapter.text_mode_data[] and all the code that goes with it.

Are there other bus listeners besides ScreenAdapter for any of the 4 text-related bus events screen-put-char, screen-update-cursor, screen-set-size-text and screen-update-cursor-scanline that need to be taken into consideration? Could you please describe a bit more to me what behaviour you would like to see restored?

Currently, the only essential function of ScreenAdapter in graphical text mode is that it controls the rendering loop at ~60fps, faithfully believing it is in graphical mode (hehe).

I haven't made up my mind, though I currently think GraphicalText (in vga_text.js) is much closer to VGAScreen than ScreenAdapter. Maybe GraphicalText could be split across VGAScreen and ScreenAdapter since its function does sit somewhere in between the two. The rendering should happen in VGAScreen, do you agree? I need to think about this, not sure where to draw the line.

By the way, even though it is not used in graphical text mode, the HTML DIV element is still required because ScreenAdapter still assumes its existence, this could be easily changed as ScreenAdapter also knows about graphical text mode in its constructor through the screen_options constructor argument. Should I?

src/vga.js
copy: > I could add mouse-controlled text-selection though, roughly like this:

This sounds reasonable. I would change the following:

> If any character within the current selection is changed, selection gets cancelled and reset to nothing.

In my experience, this makes copying regularly changing text annoying, and the alternative is better. 

> Immediately copy to clipboard when user changes selected text range.

I don't think this is a good idea, as users expect the "browser-like" behaviour of copying with ctrl+c. It's also likely that browser's don't allow changing the clipboard without user interaction.

An alternative implementation could use regular DOM text elements with the current vga text implementation, that are "above" the canvas screen but opaque. I believe xterm.js does something like that, but I can't find the implementation right now.

> Or, make text selection in graphical text mode a special state that has to be triggered through the HTML user interface, naively like a button "Begin Text Selection" which changes its label to "End Text Selection" when clicked, and during that state the left mouse button would be used for text selection normally and all mouse events would be hidden from the virtualized OS. I'd prefer that.

That sounds reasonable (much better than a mouse-button + key combination), and it's also similar how text selection in the existing text screen works: When the guest OS uses the mouse, clicking on the screen doesn't select text, it locks the mouse. When the user clicks on "disable mouse" (or the guest OS doesn't use the mouse in the first place), text selection is provided by the browser.

copy:
> Are there other bus listeners besides ScreenAdapter for any of the 4 text-related bus events screen-put-char, screen-update-cursor, screen-set-size-text and screen-update-cursor-scanline that need to be taken into consideration? Could you please describe a bit more to me what behaviour you would like to see restored?
> ...
> I haven't made up my mind, though I currently think GraphicalText (in vga_text.js) is much closer to VGAScreen than ScreenAdapter. Maybe GraphicalText could be split across VGAScreen and ScreenAdapter since its function does sit somewhere in between the two. The rendering should happen in VGAScreen, do you agree? I need to think about this, not sure where to draw the line.

Looking at this again, you're right that the line is blurry. I think we can merge this in its current design, and I'll think about potential refactoring later (see the next paragraph).

Regarding bus, this is a leftover from a very old version of v86 that supported web workers (with the idea that "bus" was the boundary between web worker and main thread). It's now mostly used as an event handler for the public V86 class. I'm thinking of removing it and replacing it with more direct calls.

> By the way, even though it is not used in graphical text mode, the HTML DIV element is still required because ScreenAdapter still assumes its existence, this could be easily changed as ScreenAdapter also knows about graphical text mode in its constructor through the screen_options constructor argument. Should I?

No, an invisible div doesn't harm anyone (and code that switches between modes might assume its existence).

src/vga.js
chschnell: Fully agree (explicit select/copy-state and Ctrl+C to copy).

The canvas overlay idea is good and I will look into it when I find time. Maybe one way to go would be to use [context.fillRect()](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fillRect) with [context.globalCompositeOperation = "xor"](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation) to invert text in the browser canvas at low costs (as oppsed to the render() function of GraphicalText).

A HTML DIV overlay in graphical text mode would make the font appearance mismatch. It also touches on the issue of the dreaded 8-bit codepages and their Unicode-mappings.

However, I'd vote to leave this issue for another day.

chschnell:
Yesterday I took a very detailled look at this PR's changeset to ensure everyhing was ok (which it wasn't), and also took notes on the design for an overview.

Here are the current relations between VGAScreen, ScreenAdapter and GraphicalText:

    +---------------+         +---------------+         +---------------+
    | ScreenAdapter |<= BUS =>|   VGAScreen   |<= API =>| GraphicalText |
    +---------------+         +---------------+         +---------------+

In detail (functions that suppress a bus message are annotated with the message's name):

    1. VGAScreen uses these public methods of GraphicalText:

       GraphicalText(bus, vga)
       GraphicalText.set_character_map(char_map_select)
       GraphicalText.set_size(rows, cols)                  // "screen-set-size-text"
       GraphicalText.set_cursor_pos(row, col)              // "screen-update-cursor"
       GraphicalText.set_cursor_attr(start, end, visible)  // "screen-update-cursor-scanline"
       GraphicalText.invalidate_row(row)                   // "screen-put-char"
       GraphicalText.invalidate_font_shape()
       GraphicalText.resize_canvas()
       GraphicalText.invalidate_all()
       GraphicalText.render()

    2. GraphicalText uses these public members and methods of VGAScreen:

       VGAScreen.vga_memory[]
       VGAScreen.plane2[]
       VGAScreen.offset_register
       VGAScreen.line_compare
       VGAScreen.start_address
       VGAScreen.dac_mask
       VGAScreen.dac_map[]
       VGAScreen.vga256_palette[]
       VGAScreen.clocking_mode
       VGAScreen.attribute_mode
       VGAScreen.max_scan_line
       VGAScreen.scan_line_to_screen_row(VGAScreen.line_compare)

    3. ScreenAdapter and VGAScreen interface each other through the bus only.

    4. ScreenAdapter and GraphicalText do not interface each other.

I also cobbled a crude hack together where I changed the design to this:

    +---------------+         +---------------+         +---------------+
    |   VGAScreen   |<= BUS =>| ScreenAdapter |<= API =>| GraphicalText |
    +---------------+         +---------------+         +---------------+

This proof-of-concept worked, VGAScreen was left with almost no changes (compared to the current master) and it was as ignorant about GraphicalText as ScreenAdapter currently is in my PR. ScreenAdapter becomes the owner of the private GraphicalText instance (currently VGAScreen is the owner). There are only a few hurdles for this transformation, I guess it would take me a couple of hours to make it right and proper. As before, GraphicalText needs a valid reference to VGAScreen, but that's solvable.

If this design issue comes up in the future I'd be happy to help.

chschnell:
> Can you give me a quick sketch about what needs to be considered when saving/restoring state?

Never mind, I did a few tests and it seems to work as it is, allthough I'm not sure I'm testing right.

I saved the state, reloaded my browser, rebooted using the same image and then restored the state. The screen restored to the exact same state (content and cursor) that I had saved.

Then I changed mode to 50 rows, trashed the screen and restored state again. Also the second time the screen was restored perfectly.

I guess it works because I'm working off the VGA state which does get saved/restored properly.

src/vga.js
chschnell: See this [little demo](https://www.rekoba.de/~cs/v86/select-demo/) about highlighting selected text in a canvas using xor. That on the top is a canvas with a static image, the raw text data is simply stored fixed in the js file.

context.globalCompositeOperation isn't what it seems.

copy:
FYI, I'm reworking the bus interface, and will try to get this merged afterwards.

chschnell:
> FYI, I'm reworking the bus interface, and will try to get this merged afterwards.

Ok!

The other day I made a thorough cross-check with qemu, and the text modes look and behave the same as in this PR.

So from my perspective it's complete and ready to be merged.

bit-lang:
just found out yesterday there's another repo named doswasmx, seemingly had solutions to a few issues here.
But that is mainly aiming at gaming, I suppose.

chschnell:
> just found out yesterday there's another repo named doswasmx, seemingly had solutions to a few issues here. But that is mainly aiming at gaming, I suppose.

If I run a text mode application in DOS Wasm X which intensely dumps to stdout the screen gets quite unresponsive and kinda breaks (I ran nmake over the MS-DOS 4.0 source tree with a Pentium II at 300Mhz), see here:

dos-wasm-x-nmake

It's a bit hard to catch in a screenshot but take a look at the 3rd row from the bottom for what I mean.

A bit worse, if I run the Norton Commander installer in DOS Wasm X I don't see the font-based graphical mouse-cursor emulation (see the [3rd screebshot and text below here](https://github.com/copy/v86/issues/1098#issuecomment-2307921157)), and then Firefox crashes.

I only know a few simulators and thus don't have much of an opinion, but perhaps [Em-DOSBox](https://github.com/dreamlayers/em-dosbox) is a better wasm-port of DOSBox than "DOS Wasm X", Em-DOSBox runs all the simulations on archive.org, and those work quite well in my experience.

copy:
Merged in c758d6da40c79d8ee648e00ff66ee66e44b33501 and f92e6b4b55c3f6e5a327b836375df20fedd45fa4. I meant to push to wip instead of master, but oh well.

There's still more refactoring I'd like to do, in particular:

- Get rid of the `if(this.graphical_text) this.graphical_text.foo() else this.screen.foo()`, by merging graphical_text into ScreenAdapter
- Pass the font data from VGAScreen to GraphicalText (and remove this.vga from GraphicalText)
- Reusing the screen-put-char logic (e.g. storing the screen contents similar to ScreenAdapter
- Move the invalidation logic out of redraw
- Re-enable blinking with a proper timing (at the moment it blinks twice as fast on a 120hz screen)

However I have other things to work on. It would be nice if you could submit some of the refactorings as a PR, but it's good enough to be merged as is.

Could you do a quick test to see if everything is still working? I moved several things around, especially around the font size checks.

chschnell:
Can you undo this merge?

chschnell:
Else I would like to begin with the state of vga.js in the 2nd-to-last commit [c758d6d](https://github.com/copy/v86/commit/c758d6da40c79d8ee648e00ff66ee66e44b33501) and just override what was then merged in the last commit [f92e6b4](https://github.com/copy/v86/commit/f92e6b4b55c3f6e5a327b836375df20fedd45fa4). Would that be safe or did you make more changes in addition to merging GraphicalText that would get lost if I did this?

I have a solid plan how to implement what you listed in one go, I'm only looking for a better starting position.

Compared to the previous state of vga.js, I only need to introduce 2 new bus messages (called "screen-set-font-bitmap" and "screen-set-font-page", let me know if I should expand on them) and make sure that the 6 VGA register-additions I made are good, these are the only changes to vga.js, very few.

For the rest, GraphicalText will get merged into ScreenAdapter, and ScreenAdapter won't need a link to VGAScreen. ScreenAdapter will be the only class with knowledge about graphical text mode and will be only using received bus messages to implement this mode. Graphical text mode implementation in ScreenAdapter will reuse as much as possible from classical text mode (all the bus messages, the character and attribute buffer, cursor attributes, etc.).

copy:
> Can you undo this merge?

I can't, but feel free to include `git revert f92e6b4b55c3f6e5a327b836375df20fedd45fa4` in your PR, then I'll merge (or squash) everything in one go.

chschnell:
Ok I have started the 2nd iteration of this.

Please do not modify vga.js or screen.js in the next couple of days.

chschnell:
Success!

I can see it booting as it's supposed to, so far I've tested: 50 row-mode, 8/9px font width and font pages A/B. Fontraption looks as it should.

So the main part is working, only minor things like the cursor are left to do.

chschnell:
The redesign is complete, it's ready for review.

All previous features are implemented and tested:
- 80 and 40 columns
- 25, 28 and 50 rows
- 16-color and monochrome mode
- 8/9px font width
- 16px font width (doubled width)
- 512-character set mode (font pages A/B)
- switching between graphical text mode and graphical mode (shared canvas)
- blinking text and cursor emulation (using clock-based interval timing)

I also made sure that the "classical" (DIV-based) text mode ist still working.

chschnell:
There is one issue, I'm trying to be as brief as possible but I must give a bit of context.

* The text attribute byte is divided into two nibbles. The lower four bits are the foreground attributes (mostly color bits, see below), and the higher four are the background attributes.
* If the screen-wide BLINK_ENABLE bit (bit 4 of the VGA [Attribute Mode Control Register 10h](http://www.osdever.net/FreeVGA/vga/attrreg.htm#10)) is set, the highest bit in the fourground nibble becomes the per-character BLINK attribute, reducing the available foreground colors from 16 to 8.
* If the screen-wide PAGE_AB_ENABLE bit is set ([Character Map Select Register Index 03h](http://www.osdever.net/FreeVGA/vga/seqreg.htm#03), FONT_PAGE_SELECT is True when CHAR_SET_A_SELECT != CHAR_SET_B_SELECT), the highest bit in the background nibble becomes the per-character FONT_PAGE_SELECT attribute, reducing the available background colors from 16 to 8.

So in order for a character to blink two things must to be true **at the same time**, the same goes for a character that is to be selected from one of the two font pages A/B.

When VGAScreen calls ScreenAdapter.put_char() it uses the current value of the screen-wide bits (BLINK_ENABLE and PAGE_AB_ENABLE) to build a character's attributes and colors, and that is too early, this should be computed whenever a character is drawn to screen (not when it's stored into a buffer as it is now). The applicaton is free in the order in which it writes characters/attributes to VGA memory and when it modifes the screen-wide bits.

I do not have an good solution for this (yet), maybe you have an idea.

copy:
> The applicaton is free in the order in which it writes characters/attributes to VGA memory and when it modifes the screen-wide bits.

Presumably the palettes and screen-wide bits are not modified very often, would that be a fair assumption? If so, I'd suggest calling `VGAScreen.prototype.complete_redraw` when any of them change (which, under the hood, refills the entire screen with `put_char` calls).

chschnell:
> > The applicaton is free in the order in which it writes characters/attributes to VGA memory and when it modifes the screen-wide bits.
> 
> Presumably the palettes and screen-wide bits are not modified very often, would that be a fair assumption? If so, I'd suggest calling `VGAScreen.prototype.complete_redraw` when any of them change (which, under the hood, refills the entire screen with `put_char` calls).

Now that's a really well-balanced proposition, thank you. This will also change the code only minimally.

I'll add another commit tomorrow.

chschnell:
The second CI check for my last commit has now been stuck at `Run api tests: Restoring: filesystem` for more than 2 hours, and I'm not sure what to do about it.

    Run make api-tests
    ./tests/api/clean-shutdown.js
    Calling stop()
    Called stop()
    ./tests/api/state.js
    Saving: async cdrom
    Restoring: async cdrom
    Done: async cdrom
    Saving: sync cdrom
    Restoring: sync cdrom
    Done: sync cdrom
    Saving: filesystem
    Restoring: filesystem

In code, it's stuck at [tests/api/state.js](https://github.com/copy/v86/blob/master/tests/api/state.js) in a call to [emulator.restore_state()](https://github.com/copy/v86/blob/f92e6b4b55c3f6e5a327b836375df20fedd45fa4/tests/api/state.js#L73).

I have not modifed the state[] of VGAScreen in this redesign PR.

copy:
> The second CI check for my last commit has now been stuck at Run api tests: Restoring: filesystem for more than 2 hours, and I'm not sure what to do about it.

It's a spurious failure, feel free to ignore it.

src/vga.js
```diff
@@ -167,6 +167,7 @@ function VGAScreen(cpu, bus, screen, vga_memory_size, options)
 
     /** @type {boolean} */
     this.graphical_mode = false;
+    this.screen.set_mode(this.graphical_mode);
```
copy, line 170: Is this needed?

src/browser/screen.js
```diff
@@ -204,16 +411,45 @@ function ScreenAdapter(options, screen_fill_buffer)
...
+                let line = "";
+                for(var col = 0; col < text_mode_width; col++, i_chr += TEXT_BUF_COMPONENT_SIZE)
+                {
+                    line += codepage[text_mode_data[i_chr]];
+                }
+                text_rows.push(keep_whitespace ? line : line.trimEnd());
+            }
+        }
+        return text_rows;
+    };
```
copy, line 441: Could you merge this function into get_text_row/get_text_screen. I'd suggest removing the `keep_whitespace` parameter, the caller could do that if they want.

src/browser/screen.js
```diff
@@ -375,7 +725,10 @@ function ScreenAdapter(options, screen_fill_buffer)
 
     function update_scale_graphic()
     {
-        elem_set_scale(graphic_screen, scale_x * base_scale, scale_y * base_scale, false);
+        if(!options.disable_autoscale)
+        {
```
copy, line 729: FYI, I reverted this change in my cleanup, because it breaks `this.set_scale`, and I was under the impression that you wanted to disable the automatic 2x scaling of small resolutions.

Later I realised, that you want v86 to not touch the width/height style at all, because you're using `object-fit: contain` with `width: 100%` or similar.

For now, I'm fine with merging this as-is, but I'm not 100% happy with this API, and will probably make some breaking changes in the future. In particular:

- The 2x auto-scaling should be disableable (or be moved into the caller)
- The `disable_autoscale` option in this PR should be changeable at runtime. I'm thinking of changing it to `options.scale = 0`, so that `set_scale(0)` disables autoscale and `set_scale(1)` restores it

Note that calling `elem_set_scale` is important to get proper 1x resolution on screens where `window.devicePixelRatio` isn't 1. Otherwise, canvas pixels don't render 1-to-1 as screen pixels.

src/vga.js
```diff
@@ -676,13 +650,17 @@ VGAScreen.prototype.vga_memory_write = function(addr, value)
...
+        if(this.plane_write_bm & 0x4)
         {
+            // write to plane 2 (font-bitmap)
             this.plane2[addr] = value;
-            return;
         }
+    }
+    else
+    {
+        // write to planes 0 and 1
```
copy, line 663:
```suggestion
    else if(this.plane_write_bm & 0x3)
    {
        // write to planes 0 and 1
        this.vga_memory_write_text_mode(addr, value);
    }
    else if(this.plane_write_bm & 0x4)
    {
        // write to plane 2 (font-bitmap)
        this.plane2[addr] = value;
```

src/browser/screen.js
```diff
@@ -12,6 +12,22 @@ function ScreenAdapter(options, screen_fill_buffer)
...
+    const FLAGS_INDEX = 1;
+    const BG_COLOR_INDEX = 2;
+    const FG_COLOR_INDEX = 3;
+    const TEXT_BUF_COMPONENT_SIZE = 4;
+
+    const FLAG_BLINKING = 0x01;
+    const FLAG_FONT_PAGE_B = 0x02;
+
+    this.FLAG_BLINKING = FLAG_BLINKING;
+    this.FLAG_FONT_PAGE_B = FLAG_FONT_PAGE_B;
```
copy, line 29: I think these are better as global constants. If those cause a problem in the future, I'll consider using es6 imports.

src/browser/screen.js
```diff
@@ -67,86 +101,259 @@ function ScreenAdapter(options, screen_fill_buffer)
         return "#" + "0".repeat(6 - n.length) + n;
     }
 
+    function render_font_bitmap(src_bitmap)
+    {
+        const dst_size = 8 * 256 * font_width * font_height;
+        const dst_bitmap = font_bitmap && font_bitmap.length === dst_size ?
+            font_bitmap : new Uint8ClampedArray(dst_size);
```
copy, line 108: This inspects the global `font_bitmap`, returns it and then the caller overwrites the same global, which is a bit tedious. How about:

```js
if(font_bitmap.length !== dst_size) font_bitmap = new Uint8ClampedArray(dst_size)
// use font_bitmap in the rest of this function
```

If it doesn't cause problems with the rest of the code, we could also change the condition to `font_bitmap.length < dst_size`, so repeated font resizes don't cause allocations.

src/browser/screen.js
```diff
@@ -67,86 +101,259 @@ function ScreenAdapter(options, screen_fill_buffer)
...
+                            graphical_text_buffer[gfx_i+2] = fg_b;
+                        }
+                        else
+                        {
+                            graphical_text_buffer[gfx_i]   = bg_r;
+                            graphical_text_buffer[gfx_i+1] = bg_g;
+                            graphical_text_buffer[gfx_i+2] = bg_b;
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
chschnell: That was the initial idea two or so months ago, but I soon came to realize that [`putImageData()`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/putImageData) does not support changing the colors of the monochrome source bitmap, so this would unfortunately only work in monochrome mode.

I'm fully aware that `render_changed_rows()` implements a costly set of nested loops. There are two alternate ideas I have for this method in the back of my mind that would reduce costs: WebFonts or WASM.

* The first idea is to render VGA fonts dynamically into browser fonts, and then use classical text mode with these dynamically created web fonts. MIT-licensed library [opentype.js](https://opentype.js.org/) would seem a good starting point, and in their [issue 62](https://github.com/opentypejs/opentype.js/issues/62#issuecomment-2067820197) is a nice discussion about how to load such dynamically created fonts dynamically back into the browser. There are some downsides to this which kept me from pursuing this further:  
  * The discrete VGA glyph bitmaps would need to be vectorized to each glyph's outline, it's not enough to map pixels to squares, neighbouring squares need to be merged into larger shapes or it looks like a pixel cloud.
  * Browser's font scaling is not always precise, and
  * the recommended method to load the dynamically created font into the web page looks very costly, however font updates by the guest OS are usually extremely rare.
  
  Mabye the new [FontFace API](https://developer.mozilla.org/en-US/docs/Web/API/FontFace) could be useful for this, too.

* The second idea would be to rewrite this algorithm in WASM/rust. I'm still too new to the rust language and ecosystem to properly reason about this. I understand this would have to be asynchronous? It would be nice to write a synchronous WASM function where I could pass arguments by reference (to avoid copying these large buffers around). Writing the actual function would not be an issue for me, but I'd like to discuss this upfront cause I'm not sure how to go about it.

The second idea would be much more streamlined with the current code, the first would blow up everything again.

Any ideas and opinions would be very welcome here.

src/browser/screen.js
chschnell: First, I never got around to mention it properly, but there is a bug in V86's autoscale. Secondly, the way it is written prevents embedders to apply their own scaling rules to canvas using CSS styles and classes.

The bug is not obvious: Enter fullscreen mode, switch to a graphical mode with small dimensions (320x200, anything that triggers autoscale), and then directly exit fullscreen mode again: The canvas stays zoomed and does not fall back into its non-fullscreen dimensions. The bug is unrelated to graphical text mode, you can test it at https://copy.sh/v86/ using my image [FreeDOS-20m-demo2.img](https://www.rekoba.de/~cs/v86/demo/img/FreeDOS-20m-demo2.img) (20M), steps to reproduce:

    # first boot up image and enter fullscreen
    cd monkey
    monkey.exe
    # now press ESC to exit fullscreen

Before I added code to disable autoscale, I tried to fix this bug. Adding an event listener for [`fullscreenchange`](https://developer.mozilla.org/en-US/docs/Web/API/Document/fullscreenchange_event) event did not help, I believe it's a case for [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) which I felt overblown. Simply disabling it by opt-out seemed like a good middleground to me.

To the second point: The problem crystallizes in [these two lines](https://github.com/copy/v86/blob/d49cb4c65a7a126e45f25141fad3181d79c49359/src/browser/screen.js#L383-L384) in screen.js. Function elem_set_scale() unconditionally rolls over whatever is in these CSS attributes like a train (the same is true about forcing display attribute [`block`](https://github.com/copy/v86/blob/d49cb4c65a7a126e45f25141fad3181d79c49359/src/browser/screen.js#L132) for visibilty, there are more than a dozen other values besides `block` that one could want to use here, like `flex` or `grid`).

In my personal opinion (no one needs to agree) this destructive forcing can easily be avoided by use of CSS classes (which can be overridden by the embedder) instead of direct write access from code to individual element's CSS style attributes for DOM objects that were not created by V86 (like for example the cursor DIV which is out of scope for embedders).

> The disable_autoscale option in this PR should be changeable at runtime. I'm thinking of changing it to options.scale = 0, so that set_scale(0) disables autoscale and set_scale(1) restores it

Good idea, I'll look into it tomorrow.

> Note that calling `elem_set_scale` is important to get proper 1x resolution on screens where `window.devicePixelRatio` isn't 1. Otherwise, canvas pixels don't render 1-to-1 as screen pixels.

I wasn't aware of this and will include it.

src/browser/screen.js
chschnell: Should I move the two consts to const.js? Are their names ok for a global scope?

copy: No, leave them in this file.

src/browser/screen.js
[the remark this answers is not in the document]
copy: > That was the initial idea two or so months ago, but I soon came to realize that [putImageData()](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/putImageData) does not support changing the colors of the monochrome source bitmap, so this would unfortunately only work in monochrome mode.

I see. There's still a possibility to do it using [drawImage](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage) with certain [globalCompositeOperation](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Compositing#globalcompositeoperation) options, e.g.: https://stackoverflow.com/a/58077837

But I'd also merge it as it is now.

> Any ideas and opinions would be very welcome here.

Both of those suggestions sound a bit overkill for me.

src/vga.js
```diff
@@ -1456,11 +1406,11 @@ VGAScreen.prototype.port3C0_write = function(value)
                     var previous_mode = this.attribute_mode;
                     this.attribute_mode = value;
 
-                    const is_graphical = (value & 0x1) !== 0;
+                    var is_graphical = (value & 0x1) > 0;
```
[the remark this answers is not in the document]
chschnell: Sorry, I didn't mean to do that, resolved with 861dfb4.

src/browser/screen.js
```diff
@@ -67,86 +101,259 @@ function ScreenAdapter(options, screen_fill_buffer)
...
+        const gfx_col_step = (font_width - font_height * gfx_width) * 4;
+        // move forward to start of column's next line in graphical_text_buffer
+        const gfx_line_step = (gfx_width - font_width) * 4;
+        // current cursor linear position in canvas coordinates (top left of its row/col)
+        const cursor_gfx_i = (cursor_row*gfx_width*font_height + cursor_col*font_width) * 4;
+
+        let txt_i, chr, chr_flags, chr_blinking, chr_font_offset, chr_bg_rgba, chr_fg_rgba;
+        let fg, bg, fg_r=0, fg_g=0, fg_b=0, bg_r=0, bg_g=0, bg_b=0;
+        let gfx_i, gfx_end_y, gfx_end_x, gfx_ic, glyph_i;
+        let draw_cursor, row, col, n_rows_rendered=0;
```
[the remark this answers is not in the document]
chschnell: Resolved in ff632a0.

src/browser/screen.js
```diff
@@ -339,12 +688,14 @@ function ScreenAdapter(options, screen_fill_buffer)
 
         graphic_screen.width = width;
         graphic_screen.height = height;
-        // graphic_context loses its configuration when its graphic_screen gets resized, reinitialize
+
+        // graphic_context must be reconfigured whenever its graphic_screen is resized
+        graphic_context.webkitImageSmoothingEnabled = false;
+        graphic_context.mozImageSmoothingEnabled = false;
```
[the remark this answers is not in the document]
chschnell: My bad, that's a leftover from a test I had forgotten about. Resolved in d35da67.

src/vga.js
chschnell: No, resolved in 48b9975.

src/browser/screen.js
chschnell: Done, resolved in 9f2d029 and e58e02b.

src/vga.js
chschnell: Resolved in 0e76c0f.

src/browser/screen.js
chschnell: Ok, I left this unchanged.

src/browser/screen.js
chschnell: Sure, both resolved in 403d5e5.

src/browser/screen.js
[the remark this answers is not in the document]
chschnell: > There's still a possibility to do it using [drawImage](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage) with certain [globalCompositeOperation](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Compositing#globalcompositeoperation) options, e.g.: https://stackoverflow.com/a/58077837

Interesting, I thought that wasn't possible because I had wrong ideas about globalcompositeoperation.

I might take a closer look some day but leave it as it is for now.

> Both of those suggestions sound a bit overkill for me.

Agreed :)

src/browser/screen.js
chschnell: I hadn't fully understood elem_set_scale() until now, thanks for the clarifications, except one question regarding this small section:

https://github.com/copy/v86/blob/d49cb4c65a7a126e45f25141fad3181d79c49359/src/browser/screen.js#L418-L423

What is the assumed range of window.devicePixelRatio here?

Anyway, I chose the most simple solution to address this in ea2119f for now, though that's still not really what you were describing. The solution is to disable all aspects of automatic scaling when options.scale is set to 0 (as before with disable_autoscale).

An application that sets options.scale to 0 has the full responsibilty for scaling, but is free in how it implements that. That's my propsal for this.

src/browser/screen.js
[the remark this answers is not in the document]
chschnell: I've fully reimplemented rendering using drawImage() and globalCompositeOperation, it works really great!

I'll commit this tomorrow.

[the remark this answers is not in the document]
chschnell: See commit fcb634b.

It passes all relevant tests with the current releases of Firefox and Chrome under Windows.

To make it a bit more obvious how this rendering method works I've put in a few comments.

src/browser/screen.js
copy: > window.devicePixelRatio

It can be any number bigger or equal to 1, including non-integers. The division looks a bit weird, but it indeed does the right thing for all devicePixelRatios. 

> Anyway, I chose the most simple solution to address this in https://github.com/copy/v86/commit/ea2119f094ecbeb30a1070104db5ef27aa0dc992 for now, though that's still not really what you were describing. The solution is to disable all aspects of automatic scaling when options.scale is set to 0 (as before with disable_autoscale).

Yes, that's what I had in mind.

copy:
@chschnell This is good to merge now, do you mind if I squash?

chschnell:
> @chschnell This is good to merge now, do you mind if I squash?

Yes, it's good to go.

src/browser/screen.js
chschnell: > > window.devicePixelRatio
> 
> It can be any number bigger or equal to 1, including non-integers. The division looks a bit weird, but it indeed does the right thing for all devicePixelRatios.

Some made up numbers to make my point:

- if scale is 10 and device_pixel_ratio is any positive integer, scale remains 10
- if scale is 10 and device_pixel_ratio is 2.5, scale becomes 4

If you don't mind, can you elaborate a bit on why these cases are treated so differently? What browser behaviour is being fought here?

I must admit that device_pixel_ratio and HiDPI screens are still a bit of a mystery to me.

Here's my understanding of device_pixel_ratio for you to correct: its value is a combination of the number P of physical pixels per css pixel (which should always be a positive integer, right?) and some small fractional value Z (between 0 and 1, correct?) stemming from the browser's zoom level, so device_pixel_ratio = P + Z.

And about HiDPI screens: On older "96-DPI" screens (like mine) browsers set P to 1, and on HiDPI screens to 2, 3 or 4.

I'm just trying to get my head around that.

> 
> > Anyway, I chose the most simple solution to address this in [ea2119f](https://github.com/copy/v86/commit/ea2119f094ecbeb30a1070104db5ef27aa0dc992) for now, though that's still not really what you were describing. The solution is to disable all aspects of automatic scaling when options.scale is set to 0 (as before with disable_autoscale).
> 
> Yes, that's what I had in mind.

Great, we'll leave it at that, then.

copy:
@chschnell Thanks for the contribution!

chschnell:
You're welcome, and I'm really pleased with the outcome.

Issue #1133, opened 2024-08-24
Adding speed input


donno2048:
What about being able to control the delay between any two instructions executed to be able to debug closer to single-stepping?

PR #1134, opened 2024-08-24
Add delay slider

donno2048:
solves #1133

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -15,6 +15,9 @@ function v86(bus, wasm)
     /** @type {boolean} */
     this.idle = true;
 
+    /** @type {number} */
+    this.delay = 0;
+
     this.tick_counter = 0;
     this.worker = null;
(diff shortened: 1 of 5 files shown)


src/main.js
```diff
@@ -55,7 +58,7 @@ v86.prototype.do_tick = function()
     this.idle = false;
     const t = this.cpu.main_loop();
 
-    this.next_tick(t);
+    this.next_tick(this.delay || t);
```
donno2048, line 61: Maybe `this.delay + t` is better?

copy:
Let's close this, as it doesn't really achieve single stepping. See https://github.com/copy/v86/issues/1133#issuecomment-2312407357

donno2048:
I think it's useful not just to emulate single stepping, but okay (sorry if it sounds passive aggressive 😅)

PR #1135, opened 2024-08-25
I had < 64MB of ram and suffered

KernelDeimos:
I decided to put an error log here because the comment 6 frames up and 100 lines away wasn't noticed by me until after some time.

### DIFF
--- a/src/kernel.js
+++ b/src/kernel.js
@@ -156,6 +156,11 @@ function load_kernel(mem8, bzimage, initrd, cmdline)
     {
         ramdisk_address = INITRD_ADDRESS;
         ramdisk_size = initrd.byteLength;
+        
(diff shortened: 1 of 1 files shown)


copy:
I think this is a better fix: https://github.com/copy/v86/commit/2be07a1b9bade3f88cbec570a64a40dd5642d41a

KernelDeimos:
> I think this is a better fix: [2be07a1](https://github.com/copy/v86/commit/2be07a1b9bade3f88cbec570a64a40dd5642d41a)

Yep! That said, when I encountered this error I realized I needed more than 64MiB to get anywhere.

ProgrammerIn-wonderland:
Should this PR be closed now?

KernelDeimos:
> Should this PR be closed now?
I think so

PR #1139, opened 2024-08-28
migrate 9pfs to use TextEncoder to fix 3 byte utf bug

ProgrammerIn-wonderland:
3 byte utf8 characters such as `【` in 9pfs filenames are currently broken due to a bugged UTF implementation in /lib/utf8.js. This PR aims to get rid of most of the code in utf8.js and replace it with the modern TextEncoder and TextDecoder equivalents


lib/utf8.js
```diff
@@ -4,60 +4,9 @@
...
-        return -1;
-    };
-}
-
-function UnicodeToUTF8Stream(key)
-{
-        if(key < 0x80)  return [key];
-        if(key < 0x800) return [0xC0|((key>>6)&0x1F), 0x80|(key&0x3F)];
+function utf8_len(s) {
+    return texten.encode(s).length;
```
copy, line 11: I'd suggest to inline this function and delete this file.

ProgrammerIn-wonderland: :+1:, what's a good new home for it, 9p.js?

ProgrammerIn-wonderland: Got it

copy:
Thanks!

Issue #1127, opened 2024-08-18
AltGr-Key under Windows


chschnell:
When using FreeDos's EDIT.EXE with European codepage CP858 and a german keyboard layout KEYB GR under Windows, all key-combinations using the AltGr-key are dead (AltGr is also called "right Alt" or "AltGraph", the key to the right of the spacebar). Due to the [German keyboard layout](https://kbdlayout.info/kbdgr/overview+scancodes) this fatally deprives me of `\ | { } [ ]`, amongst other keys. I found a fix, and this may be of interest to others.

The problem has to do with Windows, not with my german setup. A characteristic of browsers under Windows (tested Firefox, Chrome and Edge) is that pressing the `AltGr` key generates two KeyboardEvent events instead of one, first one with KeyboardEvent.code `ControlLeft` directly followed by another with code `AltRight`, instead of a single AltRight event like under Linux and Mac (`AltRight` is the KeyboardEvent.code value of the `AltGr` key).

You can quickly try out Mozilla's [KeyboardEvent example](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#try_it_out) to see which KeyboardEvent(s) your platform generates when pressing the AltGr key (what matters is the value shown at `code=`). Here's what I get (the sequence is identical when releasing AltGr, ControlLeft directly followed by AltRight, which is not shown there):

    KeyboardEvent: key='Control'  | code='ControlLeft'
    KeyboardEvent: key='AltGraph' | code='AltRight'

I figured maybe the extra ControlLeft-key was causing the problems observed in EDIT.EXE, and indeed, when replacing the key sequence ControlLeft+AltRight with a single AltRight (for both pressing and releasing), every single key on my keyboard works as it should, and for the first time I can edit text files properly under FreeDOS!

Here's the patch for src/browser/keyboard.js: **[v86-keyboard-altgr.patch](https://github.com/user-attachments/files/16651751/v86-keyboard-altgr.patch)**. EDIT: Improved patch by using [KeyboardEvent.getModifierState()](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState) to ensure that ControlLeft is only removed from the keyboard stream when AltGr is pressed. 

I don't have enough information to know if this modification breaks expected behaviour under OSes other than FreeDOS when virtualized under Windows, maybe others can help out testing to see if this has any negative side effects.

[...]

PR #1144, opened 2024-08-30
Remove spurious ControlLeft-key when AltRight is pressed under Windows

chschnell:
Remove ControlLeft from keyboard input sequence `ControlLeft + AltRight` if platform is Windows and `AltRight` is pressed.

See issue #1127 for more details.

### DIFF
--- a/src/browser/keyboard.js
+++ b/src/browser/keyboard.js
@@ -433,7 +433,7 @@ function KeyboardAdapter(bus)
                     handle_event(deferred_event, deferred_keydown);
                     deferred_event = null;
                 }, 10);
-                return;
+                return false;
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/browser/keyboard.js
```diff
@@ -426,11 +442,47 @@ function KeyboardAdapter(bus)
...
+        }
+
+        if(code === 0x1D || code === 0x9D)
+        {
+            // defer ControlLeft-Down/-Up until the next invocation of this method or 10ms have passed, whichever comes first
+            deferred_code = code;
+            deferred_timeout_id = setTimeout(() => {
+                send_to_controller(deferred_code);
+                deferred_code = 0;
+            }, 10);
```
copy, line 485: I think this logic should be moved earlier in the chain, into `handler` before `translate`. That way we don't have to pass through the event object, `keys_pressed` doesn't get infected, and we can use `e.code === "AltLeft"` directly.

copy:
Thanks!

Issue #1088, opened 2024-07-21
Screen appears to "loop around" when resolution goes above 1024x768 on Alpine Linux


HunterBoy344:
Browser: Chromium version 126.0.6478.182
VM memory: 256MB
VM video memory: 128MB

Screenshot 1 (before resolution change, login screen, 1024x768):
![image](https://github.com/user-attachments/assets/3b43ee46-6f68-4c08-954f-4456926e3458)

Screenshot 2 (after resolution change, desktop, 1280x960):
![image](https://github.com/user-attachments/assets/6e3dda92-ecb4-4f48-a251-56741aa1e309)

The start button is supposed to be on the far left, and the sliver of my mouse you can see is where the right edge of the screen is supposed to be. Everything appears to be offset, and moving my mouse off the left edge of the screen causes it to loop over to the right.

I can avoid this issue by not raising the video memory of the VM above 8MB, preventing it from resizing above 1024x768. However, I want a higher resolution than that, and I also want more video memory, so this is a bit of a problem. This bug doesn't seem to happen when I use my disk image on QEMU, nor does it happen when changing resolutions in the official Windows 2000 VM on V86.

Disk image is [here](https://drive.google.com/file/d/1UfRCxOAvZtDPTp1tuYQxD5RUzTEUIO6j/view?usp=drive_link). It's 3GB and it contains a copy of Alpine Linux with linux-virt and the Xfce desktop environment. Default password is alp1neNSO.

ProgrammerIn-wonderland:
How do other kernels treat the display? Like linux-lts in the place of Linux virt?

chschnell:
I've taken a [log of V86's VGA debug messages](https://github.com/user-attachments/files/16868997/alpine-log.txt) when booting this image.

The switch from 1024x768 -> 1280x960 begins at timestamp 11:40:35, I've inserted 3 lines breaks in the log. I think this stands out:

    11:40:36.158 11:40:36+158 [VGA ] 1CF / dispi write 0x6: 0x500
    11:40:36.158 11:40:36+158 [VGA ] Unimplemented dispi write index: 0x6
    11:40:36.158 11:40:36+158 [VGA ] SVGA: enabled=false, 1280x960x32
    11:40:36.158 11:40:36+158 [VGA ] 1CF / dispi write 0x7: 0x6666
    11:40:36.158 11:40:36+158 [VGA ] Unimplemented dispi write index: 0x7
[...]

PR #1147, opened 2024-09-04
Add support for VBE_DISPI_INDEX_X_OFFSET

chschnell:
Adds support for horizontal offset in Bochs VESA BIOS Extension.

Should fix issue #1088.

### DIFF
--- a/src/vga.js
+++ b/src/vga.js
@@ -220,7 +220,7 @@ function VGAScreen(cpu, bus, vga_memory_size)
      * @type {number}
      */
     this.svga_offset = 0;
-
+    this.svga_offset_x = 0;
     this.svga_offset_y = 0;
 
     const pci_revision = 0; // set to 2 for qemu extended registers
@@ -2123,14 +2123,23 @@ VGAScreen.prototype.port1CF_write = function(value)
             dbg_log("SVGA bank offset: " + h(value << 16), LOG_VGA);
             this.svga_bank_offset = value << 16;
             break;
+        case 8:
+            // x offset
+            dbg_log("SVGA X offset: " + h(value), LOG_VGA);
+            if(this.svga_offset_x !== value)
+            {
+                this.svga_offset_x = value;
+                this.svga_offset = this.svga_offset_y * this.svga_width + this.svga_offset_x;
+                this.complete_redraw();
+            }
+            break;
         case 9:
             // y offset
-            const offset = value * this.svga_width;
-            dbg_log("SVGA offset: " + h(offset) + " y=" + h(value), LOG_VGA);
+            dbg_log("SVGA Y offset: " + h(value * this.svga_width) + " y=" + h(value), LOG_VGA);
             if(this.svga_offset_y !== value)
             {
                 this.svga_offset_y = value;
-                this.svga_offset = offset;
+                this.svga_offset = this.svga_offset_y * this.svga_width + this.svga_offset_x;
                 this.complete_redraw();
             }
             break;
@@ -2204,7 +2213,7 @@ VGAScreen.prototype.svga_register_read = function(n)
 
         case 8:
             // x offset
-            return 0;
+            return this.svga_offset_x;
         case 9:
             return this.svga_offset_y;
         case 0x0A:
--- a/src/vga.js
+++ b/src/vga.js
@@ -2163,6 +2163,9 @@ VGAScreen.prototype.port1CF_write = function(value)
 
     if(this.svga_enabled && this.dispi_index === 4)
     {
+        this.svga_offset = 0;
+        this.svga_offset_x = 0;
+        this.svga_offset_y = 0;
         this.set_size_graphical(this.svga_width, this.svga_height, this.svga_bpp, this.svga_width, this.svga_height);
         this.bus.send("screen-set-mode", true);
         this.graphical_mode = true;


copy:
I don't believe this change is correct. The vga code copies memory linearly (and thus only supports offsets on the `y` axis). See https://github.com/copy/v86/blob/master/src/rust/cpu/vga.rs

Note that virtual width/height are also not supported. Writes to the corresponding registers (6 and 7) are ignored, and reads simply return full width/height.

chschnell:
> I don't believe this change is correct. The vga code copies memory linearly (and thus only supports offsets on the `y` axis). See https://github.com/copy/v86/blob/master/src/rust/cpu/vga.rs

This PR does not change the linear memory layout in VGA RAM.

What it changes is the calculation of the fixed offset in VGA memory `this.svga_offset` where this linear block of memory begins.

As you pointed out, this.svga_offset can currently only be a multiple of the horizontal screen size, in [vga.js lines 2126-2136](https://github.com/copy/v86/blob/3c77b98bc4bc7a5d51a2056ea73d7666ca50fc9d/src/vga.js#L2126-L2136) it is effectively calculated as:

    this.svga_offset = this.svga_offset_y * this.svga_width;

This PR only extends this calculation to:

    this.svga_offset = this.svga_offset_y * this.svga_width + this.svga_offset_x;

This does not change the linear nature of the memory block that starts at this.svga_offset, only the fixed offset where it starts.

> Note that virtual width/height are also not supported. Writes to the corresponding registers (6 and 7) are ignored, and reads simply return full width/height.

Yes, and I have not changed that, this PR is independent of virtual width and height. I understand that SVGA in V86 is linear-only.

Let's take a step back and look at [what motivated this PR](https://github.com/copy/v86/issues/1088#issuecomment-2328561535).

To better see this defect I've made two new 1280x960 screenshots, both show a maximized terminal window running the `top` command (to compare it's best to open these two screenshots in separate browser tabs).

**Before this PR:**
alpine-1280x960-top-maximized
**After this PR:**
alpine-1280x960-top-maximized-fixed

It helps to zoom in into the top-right corner of the first screenshot, notice the top scanline and the 1px Y-offset:

alpine-1280x960-top-maximized-top-right

This looks exactly like the visual glitch I would expect if an offset into VGA memory was off by some value X_OFFSET being less than X_WIDTH (the length of a single scanline which is 1280 in this case). If the difference was larger than X_WIDTH then we would seen only a section or even nothing in the first screenshot, but we see all lines (although shifted by X_OFFSET).

When I looked into this I noticed two things:

1. the "cut" on the right side is at X = 1024, and
2. Alpine attempts to write 1024 into the VBE_DISPI_INDEX_X_OFFSET register when setting this 1280x960 video mode.

Hence I included VBE_DISPI_INDEX_X_OFFSET in the calculation of this.svga_offset as described above, which worked and eventually led to this PR.

I've looked over the rust code but fail to see how it is affected by this PR other than being passed a different value in argument `svga_dest_offset` (being [this.svga_offset](https://github.com/copy/v86/blob/3c77b98bc4bc7a5d51a2056ea73d7666ca50fc9d/src/vga.js#L2436)) which it simply uses as the linear offset for the copy destination buffer. My rust knowledge is probably insufficient, I'd kindly ask you to point me more directly to the problem that this PR causes in cpu/vga.rs.

chschnell:
Bochs handles VBE_DISPI_INDEX_Y_OFFSET and VBE_DISPI_INDEX_X_OFFSET structurally the same as in my interpretation (the details are a bit different because Bochs implements VBE_DISPI_INDEX_VIRT_WIDTH), see:

https://sourceforge.net/p/bochs/code/HEAD/tree/trunk/bochs/iodev/display/vga.cc#l1180

In addition, when the SVGA gets switched into enabled state in the handler of [VBE_DISPI_INDEX_ENABLE](https://sourceforge.net/p/bochs/code/HEAD/tree/trunk/bochs/iodev/display/vga.cc#l1068), Bochs resets offset_x (VBE_DISPI_INDEX_X_OFFSET) and offset_y to `0`, I think that's a good idea and missing from this PR, if you want I can add similar code to this PR at the [equivalent spot in vga.js](https://github.com/copy/v86/blob/3c77b98bc4bc7a5d51a2056ea73d7666ca50fc9d/src/vga.js#L2155).

copy:
I see, thanks for the explanation. I was under the impression that fixing the bug required virtual size support, and that x offset only made sense in conjunction with virtual size support. Alpine's behaviour here is a bit weird, but I see how the fix is correct.

> In addition, when the SVGA gets switched into enabled state in the handler of [VBE_DISPI_INDEX_ENABLE](https://sourceforge.net/p/bochs/code/HEAD/tree/trunk/bochs/iodev/display/vga.cc#l1068), Bochs resets offset_x (VBE_DISPI_INDEX_X_OFFSET) and offset_y to 0, I think that's a good idea and missing from this PR, if you want I can add similar code to this PR at the [equivalent spot in vga.js](https://github.com/copy/v86/blob/3c77b98bc4bc7a5d51a2056ea73d7666ca50fc9d/src/vga.js#L2155).

Yes, feel free to include that change in this PR, then I'll merge.

chschnell:
I added the three lines to the PR.

It's a big pleasure for me to contribute to this great project!

chschnell:
> Alpine's behaviour here is a bit weird, but I see how the fix is correct.

I was also wondering why it's doing that.

IIRC that X_OFFSET changes when you modify the screen resolution in Alpine, and it didn't appear plausible to me in any way.

However, this PR fixes that, all screen resolutions in Apline work now.

copy:
> It's a big pleasure for me to contribute to this great project!

Appreciate your contributions :-)

PR #1165, opened 2024-09-26
ci: set 1 hour timeout for "Build and test" job

SuperMaxusa:
Mentioned in Gitter chat, should prevent CI from getting stuck for 6 hours.


copy:
Thanks!

> Mentioned in Gitter chat

FYI, I didn't see any related message.

PR #1178, opened 2024-11-04
improve fake network write() efficiency

chschnell:
Several improvements in fake network code that deal with data flowing towards the guest OS (i.e. package encoding and downstream buffering) and TCP state mechanics.

- new: improved support for TCP connection state mechanics, implemented graceful shutdown, fixed FIN flag in generated TCP packet stream ([context](https://github.com/copy/v86/discussions/1175#discussioncomment-11147174))
- new: member eth_encoder_buf[] in classes FetchNetworkAdapter and WispNetworkAdapter, a buffer allocated once per adapter instance to encode single ethernet frames
- new: class GrowableRingbuffer, designed to replace array TCPConnection.send_buffer[] with a growable ringbuffer
- new: added internet checksum to synthesized UDP packets (required by mTCP) in function write_udp()
- replaced internet checksum calculation with more efficient algorithm from RFC1071 (chapter 4.1)
- increased TCP chunk size in TCPConnection.pump() from 500 to 1460 bytes (max. TCP payload size)
- added function to efficiently copy an array into a DataView using TypedArray.set(), replaces a bunch of for-loops
- added function to efficiently encode a string into a DataView using TextEncoder.encodeInto()
- refactored all write_...()-functions to use encoder buffer
- added several named constants for ethernet/ipv4/udp/tcp-related offsets
- removed several apparently unused "return true" statements

### DIFF
--- a/src/browser/fake_network.js
+++ b/src/browser/fake_network.js
@@ -58,7 +58,7 @@ function iptolong(parts) {
     return parts[0] << 24 | parts[1] << 16 | parts[2] << 8 | parts[3];
 }
 
-class Uint8Stream
+class GrowableRingbuffer
 {
     /**
      * @param {number} initial_capacity
@@ -66,23 +66,12 @@ class Uint8Stream
      */
     constructor(initial_capacity, maximum_capacity)
     {
-        this.maximum_capacity = maximum_capacity;
+        initial_capacity = Math.min(initial_capacity, 16);
+        this.maximum_capacity = maximum_capacity ? Math.max(maximum_capacity, initial_capacity) : 0;
         this.tail = 0;
         this.head = 0;
         this.length = 0;
         this.buffer = new Uint8Array(initial_capacity);
-    }
-
-    /**
-     * @param {number} capacity
-     */
-    resize(capacity)
-    {
-        const new_buffer = new Uint8Array(capacity);
-        this.peek(new_buffer, this.length);
-        this.tail = 0;
-        this.head = this.length;
-        this.buffer = new_buffer;
     }
 
     /**
@@ -98,22 +87,16 @@ class Uint8Stream
                 capacity *= 2;
             }
             if(this.maximum_capacity && capacity > this.maximum_capacity) {
-                console.error("stream capacity overflow in Uint8Stream.write()");
-                return;
+                throw new Error("stream capacity overflow in GrowableRingbuffer.write(), package dropped");
             }
-            this.resize(capacity);
+            const new_buffer = new Uint8Array(capacity);
+            this.peek(new_buffer);
+            this.tail = 0;
+            this.head = this.length;
+            this.buffer = new_buffer;
         }
         const buffer = this.buffer;
 
-        /*
-        let head = this.head;
-        for(let i=0; i<src_length; ++i) {
-            buffer[head] = src_array[i];
-            head = (head + 1) % capacity;
-        }
-        this.head = head;
-        this.length += src_length;
-        */
         const new_head = this.head + src_length;
         if(new_head > capacity) {
             const i_split = capacity - this.head;
@@ -129,22 +112,13 @@ class Uint8Stream
 
     /**
      * @param {Uint8Array} dst_array
-     * @param {number} length
      */
-    peek(dst_array, length)
+    peek(dst_array)
     {
-        if(length > this.length) {
-            length = this.length;
-        }
+        const length = Math.min(this.length, dst_array.length);
         if(length) {
             const buffer = this.buffer;
             const capacity = buffer.length;
-            /*
-            for(let i=0, tail = this.tail; i<length; ++i) {
-                dst_array[i] = buffer[tail];
-                tail = (tail + 1) % capacity;
-            }
-            */
             const new_tail = this.tail + length;
             if(new_tail > capacity) {
                 const buf_len_left = new_tail % capacity;
@@ -1017,7 +991,7 @@ function fake_tcp_connect(dport, adapter)
  */
 function TCPConnection()
 {
-    this.send_stream = new Uint8Stream(2048, 0);
+    this.send_stream = new GrowableRingbuffer(2048, 0);
     this.send_chunk_buf = new Uint8Array(TCP_PAYLOAD_SIZE);
     this.seq_history = [];
 }
@@ -1188,7 +1162,7 @@ TCPConnection.prototype.close = function() {
 TCPConnection.prototype.pump = function() {
     if(this.send_stream.length > 0 && !this.pending) {
         const data = this.send_chunk_buf;
-        const n_ready = this.send_stream.peek(data, data.length);
+        const n_ready = this.send_stream.peek(data);
         const reply = this.ipv4_reply();
         this.pending = true;
         if(this.state === TCP_STATE_FIN_WAIT_1 && this.send_stream.length === n_ready) {
--- a/src/browser/fake_network.js
+++ b/src/browser/fake_network.js
@@ -17,17 +17,38 @@ const TWO_TO_32 = Math.pow(2, 32);
 const DHCP_MAGIC_COOKIE = 0x63825363;
 const V86_ASCII = [118, 56, 54];
 
-//const TCP_STATE_CLOSED = "closed";
+/* TCP 4-way close handshake
+ *
+ *          Initiator    Receiver
+ *     (active close)    (passive close)
+ *              :           :
+ *  ESTABLISHED |           | ESTABLISHED
+ *              |           |
+ *   FIN_WAIT_1 |--- FIN -->| CLOSE_WAIT    [1]
+ *              |           |
+ *   FIN_WAIT_2 |<-- ACK ---|               [2]
+ *              |           |
+ *    TIME_WAIT |<-- FIN ---| LAST_ACK      [3]
+ *              |           |
+ *              |--- ACK -->| CLOSED        [4]
+ *       CLOSED |           |
+ *
+ * - Receiver MAY combine [2] and [3] into a single ACK+FIN packet.
+ * - State CLOSING (simultaneous close from both ends) not implemented.
+ * - In [3], instead of TIME_WAIT go diretly to CLOSED after sending ACK.
+ * - TODO: what about state LISTEN? Neither WISP nor FETCH seem to support it.
+ */
+const TCP_STATE_CLOSED = "closed";
+const TCP_STATE_SYN_RECEIVED = "syn-received";
+const TCP_STATE_SYN_SENT = "syn-sent";
 //const TCP_STATE_LISTEN = "listen";
 const TCP_STATE_ESTABLISHED = "established";
 const TCP_STATE_FIN_WAIT_1 = "fin-wait-1";
-//const TCP_STATE_CLOSE_WAIT = "close-wait";
-//const TCP_STATE_FIN_WAIT_2 = "fin-wait-2";
-//const TCP_STATE_LAST_ACK = "last-ack";
+const TCP_STATE_CLOSE_WAIT = "close-wait";
+const TCP_STATE_FIN_WAIT_2 = "fin-wait-2";
+const TCP_STATE_LAST_ACK = "last-ack";
 //const TCP_STATE_CLOSING = "closing";
 //const TCP_STATE_TIME_WAIT = "time-wait";
-const TCP_STATE_SYN_RECEIVED = "syn-received";
-const TCP_STATE_SYN_SENT = "syn-sent";
 
 // source: RFC6335, 6. Port Number Ranges
 const TCP_DYNAMIC_PORT_START = 49152;
@@ -54,7 +75,7 @@ function a2ethaddr(bytes) {
     return [0,1,2,3,4,5].map((i) => bytes[i].toString(16)).map(x => x.length === 1 ? "0" + x : x).join(":");
 }
 
-function siptolong(s) {
+function siptolong(s) { // TODO: this function is not used anywhere, can it be removed?
     let parts = s.split(".").map(function(x) { return parseInt(x, 10); });
     return parts[0] << 24 | parts[1] << 16 | parts[2] << 8 | parts[3];
 }
@@ -211,11 +232,12 @@ function view_setString(offset, str, view, out)
 function view_setInetChecksum(offset, length, checksum, view, out)
 {
     const uint16_end = view.byteOffset + (length & ~1);
+    const eth_frame = out.eth_frame;
     for(let i = view.byteOffset; i < uint16_end; i += 2) {
-        checksum += out.eth_frame[i] << 8 | out.eth_frame[i+1];
+        checksum += eth_frame[i] << 8 | eth_frame[i+1];
     }
     if(length & 1) {
-        checksum += out.eth_frame[uint16_end] << 8;
+        checksum += eth_frame[uint16_end] << 8;
     }
     while(checksum >> 16) {
         checksum = (checksum & 0xffff) + (checksum >> 16);
@@ -248,7 +270,7 @@ function handle_fake_tcp(packet, adapter)
 
     if(packet.tcp.syn) {
         if(adapter.tcp_conn[tuple]) {
-            dbg_log("SYN to already opened port", LOG_FETCH);
+            dbg_log("SYN to already opened port", LOG_FETCH);   // TODO: is LOG_FETCH a good choice for this module?
         }
         if(adapter.on_tcp_connection(adapter, packet, tuple)) {
             return;
@@ -295,7 +317,7 @@ function handle_fake_dns(packet, adapter)
         let q = packet.dns.questions[i];
 
         switch(q.type){
-            case 1: // A recrod
+            case 1: // A record
                 answers.push({
                     name: q.name,
                     type: q.type,
@@ -458,7 +480,7 @@ function parse_eth(data, o) {
 
     o.eth = eth;
 
-    // Remove CRC from the end of the packet maybe?
+    // TODO: Remove CRC from the end of the packet maybe?
     let payload = data.subarray(ETH_HEADER_SIZE, data.length);
 
     if(ethertype === ETHERTYPE_IPV4) {
@@ -547,7 +569,12 @@ function parse_ipv4(data, o) {
     };
 
     // Ethernet minmum packet size.
-    if(Math.max(len, 46) !== data.length) dbg_log(`ipv4 Length mismatch: ${len} != ${data.length}`, LOG_FETCH);
+    /* TODO: What's the overall reasoning behind this check, and where does the 46 come from? We get plenty of IP packets of size 40.
+    if(Math.max(len, 46) !== data.length) {
+    */
+    if(Math.max(len, 40) !== data.length) {
+        dbg_log(`ipv4 Length mismatch: ${len} != ${data.length}`, LOG_FETCH);
+    }
 
     o.ipv4 = ipv4;
     let ipdata = data.subarray(ihl * 4, len);
@@ -945,9 +972,10 @@ function write_tcp(spec, out) {
  */
 function TCPConnection()
 {
+    this.state = TCP_STATE_CLOSED;
     this.send_buffer = new GrowableRingbuffer(2048, 0);
     this.send_chunk_buf = new Uint8Array(TCP_PAYLOAD_SIZE);
-    this.seq_history = [];
+    this.delayed_active_close = false;
 }
 
 TCPConnection.prototype.ipv4_reply = function() {
@@ -969,7 +997,27 @@ TCPConnection.prototype.ipv4_reply = function() {
     return reply;
 };
 
+TCPConnection.prototype.packet_reply = function(packet, tcp_options) {
+    const reply_tcp = {
+        sport: packet.tcp.dport,
+        dport: packet.tcp.sport,
+        winsize: packet.tcp.winsize,
+        ackn: this.ack,
+        seq: this.seq
+    };
+    if(tcp_options) {
+        for(const opt in tcp_options) {
+            reply_tcp[opt] = tcp_options[opt];
+        }
+    }
+    const reply = this.ipv4_reply();
+    reply.tcp = reply_tcp;
+    return reply;
+};
+
+// TODO: Is this method used anywhere anymore? It used to be called from fake_tcp_connect() which was removed.
 TCPConnection.prototype.connect = function() {
+    dbg_log(`TCP[${this.tuple}]: connect(): sending SYN+ACK in state "${this.state}", next "${TCP_STATE_SYN_SENT}"`, LOG_FETCH);
     this.seq = 1338;
     this.ack = 1;
     this.start_seq = 0;
@@ -1002,7 +1050,6 @@ TCPConnection.prototype.accept = function(packet) {
     this.winsize = packet.tcp.winsize;
 
     let reply = this.ipv4_reply();
-
     reply.tcp = {
         sport: this.sport,
         dport: this.dport,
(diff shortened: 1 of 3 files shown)


src/browser/fake_network.js
```diff
@@ -43,6 +58,220 @@ function iptolong(parts) {
...
+     * @param {Uint8Array} src_array
+     */
+    write(src_array)
+    {
+        const src_length = src_array.length;
+        const total_length = this.length + src_length;
+        let capacity = this.buffer.length;
+        if(capacity < total_length) {
+            while(capacity < total_length) {
+                capacity *= 2;
```
copy, line 98: This can loop forever if capacity is zero. I'd suggest adding `Math.min(capacity, 16)` or `console.assert(capacity > 0)`

src/browser/fake_network.js
```diff
@@ -43,6 +58,220 @@ function iptolong(parts) {
...
 
+class Uint8Stream
+{
+    /**
+     * @param {number} initial_capacity
+     * @param {number} maximum_capacity
+     */
+    constructor(initial_capacity, maximum_capacity)
+    {
+        this.maximum_capacity = maximum_capacity;
```
copy, line 69: Unless we need ringbuffers with different capacities, I'd suggest making this a global constant.

src/browser/fake_network.js
```diff
@@ -796,67 +950,35 @@ function write_tcp(spec, data) {
...
-        adapter.router_ip.join("."),
-        sport
-    ].join(":");
+    const vm_ip_str = adapter.vm_ip.join(".");
+    const router_ip_str = adapter.router_ip.join(".");
+    let sport, tuple;
+    do {
+        sport = 1000 + Math.random() * 64535 | 0;
+        tuple = `${vm_ip_str}:${dport}:${router_ip_str}:${sport}`;
+    } while(adapter.tcp_conn[tuple]);
```
copy, line 981: I'd suggest something along the lines of (untested):

```javascript
let sport = 1000 + Math.random() * 64535 | 0;
for(let i = 0; i < 65536; i++)
{
    tuple = `${vm_ip_str}:${dport}:${router_ip_str}:${sport}`;
    if(!adapter.tcp_conn[tuple]) break;
    sport = Math.min(1000, (sport + 1) % 65536);
}
if(adapter.tcp_conn[tuple])
{
    dbg_log("warning: no free ports");
    return;
}
````

To prevent infinite loops if ports are exhausted.

File: src/browser/fake_network.js [hunk 4]
```diff
@@ -43,6 +58,220 @@ function iptolong(parts) {
     return parts[0] << 24 | parts[1] << 16 | parts[2] << 8 | parts[3];
 }
 
+class Uint8Stream
```
copy, line 61: I'd say (Growable)Ringbuffer would be a better name.

src/browser/fake_network.js
```diff
@@ -43,6 +58,220 @@ function iptolong(parts) {
...
+        this.head = new_head % capacity;
+        this.length += src_length;
+    }
+
+    /**
+     * @param {Uint8Array} dst_array
+     * @param {number} length
+     */
+    peek(dst_array, length)
+    {
```
copy, line 135: I think we could write `let length = dst_array.length` here and remove the `length` parameter.

src/browser/fake_network.js
```diff
@@ -43,6 +58,220 @@ function iptolong(parts) {
...
+        this.eth_frame = eth_frame;
+        this.eth_frame_view = new DataView(eth_frame.buffer);
+        this.eth_payload_view = new DataView(eth_frame.buffer, offset + ETH_PAYLOAD_OFFSET, ETH_PAYLOAD_SIZE);
+        this.ipv4_payload_view = new DataView(eth_frame.buffer, offset + IPV4_PAYLOAD_OFFSET, IPV4_PAYLOAD_SIZE);
+        this.udp_payload_view = new DataView(eth_frame.buffer, offset + UDP_PAYLOAD_OFFSET, UDP_PAYLOAD_SIZE);
+
+        for(const view of [this.eth_frame_view, this.eth_payload_view, this.ipv4_payload_view, this.udp_payload_view]) {
+            view.setArray = this.view_setArray.bind(this, view);
+            view.setString = this.view_setString.bind(this, view);
+            view.setInetChecksum = this.view_setInetChecksum.bind(this, view);
```
copy, line 194: `view_{setArray,setString,setInetChecksum}` should just be global functions, rather than doing this dance with monkeypatching built-in types.

src/browser/fake_network.js
```diff
@@ -43,6 +58,220 @@ function iptolong(parts) {
...
+        return this.eth_frame.subarray(0, length);
+    }
+}
+
+const ethernet_encoder = new EthernetPacketEncoder();
+
+function make_packet(spec)
+{
+    return ethernet_encoder.encode_packet(spec);
+}
```
copy, line 273: I'm really not a fan of the `EthernetPacketEncoder` class. Is there any way we can write

```javascript
function make_packet(spec, out)
{
     // encode spec into out, which is an ArrayBuffer(View)
     // pure function
}
```

And have the caller cache `out`?

We could also consider changing `receive` to take a high-level packet description and tcp/udp payload, and construct the packet in place (in ne2k/virtio), which would be even fewer copies.

copy (commented):
Thanks! Overall, this looks like a reasonable set of changes.

I've left some comments.

src/browser/fake_network.js
chschnell: I changed it to guarantee a minimum capacity of 16 bytes.

src/browser/fake_network.js
chschnell: I rewrote that as you suggested, I had also introduced a wrong dynamic port range and changed it to the proper range I found in RFC6335.

I however think the right thing to do when the pool of dynamic TCP ports is exhausted is to throw an Exception to inform the caller about it (to teardown the connection), if you don't mind (even though complete error handling is still TODO).

src/browser/fake_network.js
chschnell: Done, and GrowableRingbuffer it is (this was the initial name of this class).

src/browser/fake_network.js
chschnell: Yes, I had overlooked that and changed it.

src/browser/fake_network.js
chschnell: Fully agree, class EthernetPacketEncoder is gone now.

src/browser/fake_network.js
chschnell: I removed EthernetPacketEncoder, it was never meant to stay.

The current state is somewhere in the middle ground towards changing `receive` to take a package descriptor object containing a reference to a GrowableRingbuffer (TCP) or ArrayBuffer (UDP). I'll do this next.

More details in the patch description.

src/browser/fake_network.js
chschnell: Class [NetworkAdapter](https://github.com/copy/v86/blob/ba7f9405840b6ae476bf01a5b4dfb7f1a6e48055/src/browser/network.js#L21) might be another candidate that could benefit from class GrowableRingbuffer, as well as the inverse TCP data flow direction (from guest to internet) which I haven't much looked at so far.

I have yet to make up my mind what to use for `maximum_capacity`. It is clear that when it gets hit (or if Uint8Array constructor throws a RangeException) the whole TCP connection needs to be torn down, dropping traffic is not an option because there is no resend mechanism at this point.

I'll come back to this.

src/browser/fake_network.js
```diff
@@ -796,67 +921,38 @@ function write_tcp(spec, data) {
...
+        (spec.ipv4.src[2] << 8 | spec.ipv4.src[3]) +
+        (spec.ipv4.dest[0] << 8 | spec.ipv4.dest[1]) +
+        (spec.ipv4.dest[2] << 8 | spec.ipv4.dest[3]) +
+        IPV4_PROTO_TCP +
+        total_length;
+    view_setInetChecksum(16, total_length, pseudo_header, view, out);
+    return total_length;
 }
 
 function fake_tcp_connect(dport, adapter)
```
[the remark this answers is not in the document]
chschnell: Yes, this isolated pair of functions can be deleted. I'll do it.

src/browser/fake_network.js
```diff
@@ -18,23 +17,65 @@ const TWO_TO_32 = Math.pow(2, 32);
...
+const TCP_PAYLOAD_OFFSET  = IPV4_PAYLOAD_OFFSET + TCP_HEADER_SIZE;
+const TCP_PAYLOAD_SIZE    = IPV4_PAYLOAD_SIZE - TCP_HEADER_SIZE;
+const ICMP_HEADER_SIZE    = 4;
 
 function a2ethaddr(bytes) {
     return [0,1,2,3,4,5].map((i) => bytes[i].toString(16)).map(x => x.length === 1 ? "0" + x : x).join(":");
 }
 
-function siptolong(s) {
+function siptolong(s) { // TODO: this function is not used anywhere, can it be removed?
```
copy, line 78: Nice spot, let's remove it.

src/browser/fake_network.js
```diff
@@ -43,6 +84,179 @@ function iptolong(parts) {
...
+        checksum += eth_frame[i] << 8 | eth_frame[i+1];
+    }
+    if(length & 1) {
+        checksum += eth_frame[uint16_end] << 8;
+    }
+    while(checksum >> 16) {
+        checksum = (checksum & 0xffff) + (checksum >> 16);
+    }
+    view.setUint16(offset, ~checksum);
+}
```
copy, line 246: I'd suggest renaming this to `calc_inet_checksum` (or similar), returning the checksum and doing the `view.setUint16` in the caller.

src/browser/fake_network.js
```diff
@@ -43,6 +84,179 @@ function iptolong(parts) {
...
+ *
+ * @param {number} offset
+ * @param {string} str
+ * @param {DataView} view
+ * @param {Object} out
+ */
+function view_setString(offset, str, view, out)
+{
+    const ofs = view.byteOffset + offset;
+    const result = out.text_encoder.encodeInto(str, ofs ? out.eth_frame.subarray(ofs) : out.eth_frame);
```
copy, line 218: Considering that `ofs` is almost never `0`, I'd remove the ternary.

src/browser/fake_network.js
```diff
@@ -43,6 +84,179 @@ function iptolong(parts) {
...
+        checksum = (checksum & 0xffff) + (checksum >> 16);
+    }
+    view.setUint16(offset, ~checksum);
+}
+
+/**
+ * @param {WispNetworkAdapter|FetchNetworkAdapter} adapter
+ * @param {Object} spec
+ */
+function adapter_receive(adapter, spec)
```
copy, line 252: Could be changed to:

```javascript
function make_packet(out, spec)
{
    dbg_assert(spec.eth);
    out.eth_frame.fill(0);
    return out.eth_frame.subarray(0, write_eth(spec, out));
}
```

And in the caller:

```javascript
   this.receive(make_packet(this.eth_encoder_buf, reply));
```

src/browser/fake_network.js
```diff
@@ -43,6 +84,179 @@ function iptolong(parts) {
...
+
+/**
+ * Copy given data array into view starting at offset, return number of bytes written.
+ *
+ * @param {number} offset
+ * @param {ArrayBuffer|ArrayBufferView} data
+ * @param {DataView} view
+ * @param {Object} out
+ */
+function view_setArray(offset, data, view, out)
```
copy, line 201:
```suggestion
function view_set_array(offset, data, view, out)
```

src/browser/fake_network.js
```diff
@@ -43,6 +84,179 @@ function iptolong(parts) {
...
+
+/**
+ * UTF8-encode given string into view starting at offset, return number of bytes written.
+ *
+ * @param {number} offset
+ * @param {string} str
+ * @param {DataView} view
+ * @param {Object} out
+ */
+function view_setString(offset, str, view, out)
```
copy, line 215:
```suggestion
function view_set_string(offset, str, view, out)
```

src/browser/fake_network.js
chschnell: Done.

There's another one: TCPConnection.connect(), see my comment there.

It was orphaned when we recently removed fake_tcp_connect() and FetchNetworkAdapter.tcp_connect() in [this commit](https://github.com/copy/v86/pull/1178/commits/f8af3c6c868557ac9b6a91847bf9070fbb9a75b3).

Just so I understand correctly, all these funcions are related to public server sockets (opened in the guest) which are not supported by either wisp- or fetch-based networking?

src/browser/fake_network.js
chschnell: Good idea, done.

src/browser/fake_network.js
chschnell: Done.

src/browser/fake_network.js
chschnell: That's better, I refactored it in this way.

src/browser/fake_network.js
chschnell: Done.

src/browser/fake_network.js
chschnell: Done.

chschnell:
I have a small idea for fetch-based networking that I would first like to check with you before committing.

Currently, the full HTTP response (header and body) from fetch() is cached before being forwarded to the originating HTTP client.

My proposal is to only wait for the HTTP response headers from fetch(), and then to forward the response body asynchronously chunk by chunk to the originating HTTP client using [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) and HTTP [Chunked transfer encoding](https://en.wikipedia.org/wiki/Chunked_transfer_encoding).

To test this, I changed on_data_http() in fetch_network.js based on a ReadableStream usage pattern from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_readable_streams#reading_the_stream), it's pretty straight-forward and it works well with my tests (all HTTP applications support chunked Transfer-Encoding to my knowledge). This would also remove the need for FetchNetworkAdapter.fetch(), I merged its relevant code into on_data_http().

The upside is that this much, much increases both the responsiveness as well as the effective speed of fetch-based networking, all the more with increasing size of the HTTP response.

The downside is that the originating HTTP client never gets to see the Content-Length of the response it is receiving, thus it cannot show any meaningful progress statistics anymore other than about what it has received so far. Any Content-Length [must be ignored](https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2) if any Transfer-Encoding is used, and the Content-Length from the fetch response header can be incorrect (for example, for compressed response bodies). The only way around this seems to be the prefetching like it currently is. I might be wrong, but to my knowledge it's categorically either prefetching or the loss of Content-Length (for any request from the originating HTTP client). I wasn't aware of that until I had finished implementing it.

I would of course wait with this until #1182 is merged, and include any changes from there.

I'm really undecided which side outweighs the other here. What do you think, thumbs up or down?

copy:
> I'm really undecided which side outweighs the other here. What do you think, thumbs up or down?

Generally thumbs up, but I'm unsure about the downsides. I wonder if HTTP Chunked transfer encoding is necessary at all. Can't the data be streamed in separate ethernet packets?

> and the Content-Length from the fetch response header can be incorrect (for example, for compressed response bodies).

I believe for compressed bodies (via [Content-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding)), `Content-Length` indicates the length of the compressed body. As far as I know, compression via `Transfer-Encoding` isn't really used or widely implemented.

chschnell:
If I could rely on `Content-Length` in the received fetch() response headers then I could just pass the unmodified chunks on to the originating client, passing the very same `Content-Length` to it.

I build and send the response headers to the originating client when I receive the first response body chunk, but I don't have a `Content-Length` to set in these response headers (well I do, but it is wrong when compression is used).

In fetch(), my browser negotiates gzip compression with my Apache server and transparently decompresses the response before passing me the decompressed chunks (and passes me the compressed size in Content-Length, which is useless). I never get to know how long the response is going to be myself.

I tried to suppress compression but [`Accept-Encoding`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) is amongst the [forbidden header names](https://fetch.spec.whatwg.org/#forbidden-header-name) in fetch, so I can't set that header in the fetch request.

I wonder, is there any other method besides Chunked transfer encoding that I could use when I don't know the length of the response?

copy:
> In fetch(), my browser negotiates gzip compression with my Apache server and transparently decompresses the response before passing me the decompressed chunks (and passes me the compressed size in Content-Length, which is useless). I never get to know how long the response is going to be myself.

I see the problem now. `fetch` passes you already decompressed chunks, but `Content-Length` indicates the compressed size on the wire.

> I wonder, is there any other method besides Chunked transfer encoding that I could use when I don't know the length of the response?

It [seems to be legal](https://stackoverflow.com/questions/11375507/getting-response-of-http-request-without-content-length#comment126626653_11375745) to send an http reponse with neither `Content-Length` nor `Transfer-Encoding: chunked`, in which case the server closes the TCP connection after writing all data.

chschnell:
> > I wonder, is there any other method besides Chunked transfer encoding that I could use when I don't know the length of the response?
> 
> It [seems to be legal](https://stackoverflow.com/questions/11375507/getting-response-of-http-request-without-content-length#comment126626653_11375745) to send an http reponse with neither `Content-Length` nor `Transfer-Encoding: chunked`, in which case the server closes the TCP connection after writing all data.

I actually do all this to learn and also try out new things, so thank you very much for this information!

It should fit in well since I'm already closing the HTTP connection to the originating client once the transaction is complete (current fetch-based networking can't do that because of the FIN packet that is sent too early).

chschnell:
It works without Content-Length and chunked encoding! :)

Of course, the originating client still doesn't get to learn the Content-Length of the response body, so it cannot show any meaningfull download progress information besides the total bytes received so far.

src/browser/fake_network.js
```diff
@@ -961,125 +1042,222 @@ TCPConnection.prototype.accept = function(packet) {
...
-        if(this.send_buffer.length < 1) reply.tcp.fin = true;
+TCPConnection.prototype.pump = function() {
+    if(this.send_buffer.length && !this.pending) {
+        const data = this.send_chunk_buf;
+        const n_ready = this.send_buffer.peek(data);
+        const reply = this.ipv4_reply();
         reply.tcp.psh = true;
-        reply.tcp_data = data;
-        this.net.receive(make_packet(reply));
+        reply.tcp_data = data.subarray(0, n_ready);
```
copy, line 1258: I believe this line is not necessary

copy:
Looks good to merge to me (one more small comment above). @chschnell Any objections? (asking because it's still marked as a draft)

src/browser/fake_network.js
chschnell: I'm not sure I understand. `reply.tcp_data` is needed, it's the TCP payload for the packet we create with `make_packet()` here. It is eventually used further down the call chain in `write_tcp()`.

chschnell:
> Looks good to merge to me (one more small comment above). @chschnell Any objections? (asking because it's still marked as a draft)

I'm still not sure whether you want the [fetch()-improvement](https://github.com/copy/v86/pull/1178#issuecomment-2471667497) or not, I can check in the code any time, it's not that much code and also well-tested.

The question is whether you prefer to have the performance boost or to keep the Content-Length intact.

I tend to the performance boost, but I'm in no position to make that decision.

SuperMaxusa:
> The question is whether you prefer to have the performance boost or to keep the Content-Length intact.

How about using the default way for small responses or when the custom `x-fetch...` header is set (like a fallback)?

src/browser/fake_network.js
copy: Sorry, I meant the `.subarray()` shouldn't be needed.

copy:
> The question is whether you prefer to have the performance boost or to keep the Content-Length intact.

Yeah, let's do proper chunking and give up the `Content-Length`. (perhaps leave the old path in the code, so it could be toggled if there are any problems).

src/browser/fake_network.js
chschnell: Well, `data` has a fixed capacity of 1460 bytes of which only `n_ready` bytes are used at this point, which can be less than the capacity.

If we remove this `.subarray()` then `write_tcp()` would somehow need to know `n_ready` explicitely together with `data`. It would need an extra field `reply.tcp_data_length` next to `reply.tcp_data` or such.

As it is only called once per generated TCP packet I would just leave it like it is. Also, `write_tcp()` passes this subarray to `view_set_array()` which also treats it as an array with implicit size.

chschnell:
> > The question is whether you prefer to have the performance boost or to keep the Content-Length intact.
> 
> Yeah, let's do proper chunking and give up the `Content-Length`. (perhaps leave the old path in the code, so it could be toggled if there are any problems).

Ok, I hope this was the last commit, from my side this PR is ready to go.

I've commented out two blocks of code in fetch_network.js (for fallback), and put a single new one in. Some kind of software switch could be added to toggle between the two implementations, but I tried to keep it simple.

@SuperMaxusa It should be easy to extend the code as it is now to a two-way solution, for example to prefetch the body when it is not too large, or when the originating client's HTTP request contained a specific header. If anyone wants it I could add it in a later PR.

SuperMaxusa:
> It should be easy to extend the code as it is now to a two-way solution, for example to prefetch the body when it is not too large, or when the originating client's HTTP request contained a specific header.

Yes, since most HTTP clients can handle these responses without `Content-Length` (e.g. on KolibriOS, DSL with Dillo and Firefox, Windows 95 and Internet Explorer 4.0 this works properly), let's leave your implementation with ReadableStream.

By the way, in many CI tries, [one test failed](https://github.com/copy/v86/actions/runs/11900235683/job/33160663531?pr=1178#step:24:2340) (as an exception, this behaviour is allowed), while the test with real `example.org` is more reliable, and after the recent changes, this test doesn't work. Can we remove this test?

chschnell:
> e.g. on KolibriOS, DSL with Dillo and Firefox, Windows 95 with Internet Explorer 4.0, this works properly

@SuperMaxusa Thanks *a lot* for testing, that is really helpful of you!

By the way, this transfer model (transfer without Content-Length and to use end-of-connection to indicate end-of-response-body) is how HTTP 1.0 used to work, and support for it is still required in HTTP 1.1 (here the [relevant section](https://greenbytes.de/tech/webdav/rfc2616.html#message.length)). I guess for these two reasons it is widely support.

chschnell:
> > It should be easy to extend the code as it is now to a two-way solution, for example to prefetch the body when it is not too large, or when the originating client's HTTP request contained a specific header.
> 
> By the way, in many CI runs, [the `mocked.example.org` test failed](https://github.com/copy/v86/actions/runs/11900235683/job/33160663531?pr=1178#step:24:2340) (as an exception, this behaviour is allowed), while the test with real `example.org` is more reliable, and after the recent changes, this test doesn't work. Can we remove this test?

```
Starting test #6: Curl mocked.example.org
    Captured: wget -T 10 -O - mocked.example.org
    Captured: echo -e done\\tmocked.example.org
    Captured: ~% wget -T 10 -O - mocked.example.org
    Captured: Connecting to mocked.example.org (192.168.87.1:80)
Fetch Failed: http://mocked.example.org/
TypeError: fetch failed
    Captured: wget: server returned error: HTTP/1.1 502 Fetch Error
    Captured: ~% echo -e done\\tmocked.example.org
    Captured: done	mocked.example.org
[...]
```

Unless there's more to it, [Test #6](https://github.com/copy/v86/blob/f767d4d73a7abe4300fa5dad1e68f15c250fc20b/tests/devices/fetch_network.js#L99-L112) in tests/devices/fetch_network.js fails correctly, it attempts to fetch() from http://mocked.example.org/ which is an invalid DNS name. The next test nr. 7 attempts to fetch() from http://example.org/ wich is a valid URL, and this test succeeds.

I am not sure, but in case you mean test nr. 6, maybe it is intended to just test that a failure in fetch() is handled correctly?

chschnell:
> > it attempts to fetch() from http://mocked.example.org/ which is an invalid DNS name
> 
> This test has a [custom fetch](https://github.com/copy/v86/blob/f767d4d73a7abe4300fa5dad1e68f15c250fc20b/tests/devices/fetch_network.js#L183-L195) for which `http://mocked.example.org` is an existed website, see also [#1061 (comment)](https://github.com/copy/v86/pull/1061#discussion_r1676848522) and [6518770](https://github.com/copy/v86/commit/6518770365d65ef8fe9ea9d06630c61539858ca2).
> 
> When the `FetchNetworkAdapter.prototype.fetch` has been [commented out](https://github.com/copy/v86/pull/1178/commits/b3721ef96390cd04e61d19f4b870d0dc7f366d55#diff-0de1f47c8b93ec88de579f44b87a38c245629c4ff19f02fa2cca8d9ef232d792R180-R206) and replaced by fetch with ReadableStream, it's not called and the test fails.

Ah I see, thanks for the explanation.

It's true that this test cannot work anymore because I removed `FetchNetworkAdapter.fetch()`, I wasn't aware of this side-effect.

I would also argue (with my limited oversight) that this test is no longer needed with the changed class layout.

ProgrammerIn-wonderland:
Related note I just noticed since this touches wisp_network. If a connection was closed in wisp and the v86 VM tried sending a TCP packet it crashes the whole thing. It may be worth just adding some extra checks around that code just to make sure everything's behaving fine

ProgrammerIn-wonderland:
Wisp does support bidirectional stream closure, is there some issue you encouraged with it?

chschnell:
> Related note I just noticed since this touches wisp_network. If a connection was closed in wisp and the v86 VM tried sending a TCP packet it crashes the whole thing. It may be worth just adding some extra checks around that code just to make sure everything's behaving fine
> 
> This PR might fix it since it touches closing logic, I'm just not sure

I did not come across this when writing the PR. I mostly changed fake_network.js, wisp_network.js only minimally where neccessary.

As a test just now, I killed my WISP server (epoxy) while V86 was running without using the network, and after a few seconds it throws an exception in the browser console. Then I restarted the WISP server, and the same V86 instance picked it up just fine, network was ok.

Then I killed the WISP server while having a telnet session open in V86, caused some traffic, restarted the WISP server, tried to exit telnet which took around half a minute (two exceptions or so), but then it came back and I was able to start a new telnet session without problems, stil in the same V86 instance.

It looks stable to me (V86 survived), but maybe I'm not testing right. How can I provoke this situation you mean?

ProgrammerIn-wonderland:
I've never heard of half close, okay I see. Does full closing still work?

ProgrammerIn-wonderland:
close 0x02 closes the TCP stream entirely, no more receiving or transmitting can be done

chschnell:
> I've never heard of half close, okay I see, wisp doesn't support that. Does full closing still work? I thought that commit removed full closing and that's why I was concerned

> close 0x02 closes the TCP stream entirely, no more receiving or transmitting can be done

Only WispServerCpp implements CLOSE/0x02 like that, neither epoxy-tls nor wisp-js do, in my observation. There was no CLOSE/0x02 in V86's WISP client before I added it a couple of days ago.

So in the case of WispServerCpp that commit did remove full closing, but in the case of the other two WISP servers it didn't change anything. I did that because none of them implemented half-close. Now it all comes down to a timeout by the server, or a leaked connection in case that never happens.

Maybe I should put the CLOSE/0x02 better back in, even though it will break certain applications and has no effect with some WISP servers.

ProgrammerIn-wonderland:
Are you saying the other wisp servers don't support closing or don't support half closing? Sorry I think I'm misunderstanding

r58Playz:
> > I've never heard of half close, okay I see, wisp doesn't support that. Does full closing still work? I thought that commit removed full closing and that's why I was concerned
> 
> > close 0x02 closes the TCP stream entirely, no more receiving or transmitting can be done
> 
> Only WispServerCpp implements CLOSE/0x02 like that, neither epoxy-tls nor wisp-js do, in my observation. There was no CLOSE/0x02 in V86's WISP client before I added it a couple of days ago.
> 
> So in the case of WispServerCpp that commit did remove full closing, but in the case of the other two WISP servers it didn't change anything. I did that because none of them implemented half-close. Now it all comes down to a timeout by the server, or a leaked connection in case that never happens.
> 
> Maybe I should put the CLOSE/0x02 better back in, even though it will break certain applications and has no effect with some WISP servers.

@chschnell

The protocol does not specify any special behavior on the server side for specific close reasons. I don't know why WispServerCpp behaves differently here, but wisp-js and epoxy-server are following protocol and closing fully.
[Wisp V1 reference](https://github.com/MercuryWorkshop/wisp-protocol/blob/main/protocol.md?plain=1#L68) [Wisp V2 reference](https://github.com/MercuryWorkshop/wisp-protocol/blob/v2/protocol.md?plain=1#L71)
> Any CLOSE packets sent from either the server or the client must immediately close the associated stream and TCP socket. The close reason in the payload doesn't affect this behavior, but may provide extra information which is useful for debugging.


On epoxy, you can use `RUST_LOG=trace` to see that it does close streams (`stream uuid f9f5dcdb-5edc-40d2-9ba7-b7686b6cfc19 disconnected for client id "127.0.0.1:46658"`):
```
[2024-11-21T20:32:37Z TRACE epoxy_server] routed "127.0.0.1:46658": Wisp
[2024-11-21T20:32:37Z DEBUG epoxy_server::handle::wisp] new wisp client id "127.0.0.1:46658" connected with extensions [1], downgraded false
[2024-11-21T20:32:37Z TRACE epoxy_server::handle::wisp] sent ping to wisp client id "127.0.0.1:46658"
[2024-11-21T20:32:37Z TRACE epoxy_server::handle::wisp] sent ping to wisp client id "127.0.0.1:46658"
[2024-11-21T20:32:37Z DEBUG epoxy_server::handle::wisp] new stream created for client id "127.0.0.1:46658": (stream uuid f9f5dcdb-5edc-40d2-9ba7-b7686b6cfc19) ConnectPacket { stream_type: Tcp, destination_port: 8000, destination_hostname: "localhost" } ConnectPacket { stream_type: Tcp, destination_port: 8000, destination_hostname: "127.0.0.1" }
[2024-11-21T20:32:37Z DEBUG epoxy_server::handle::wisp] stream uuid f9f5dcdb-5edc-40d2-9ba7-b7686b6cfc19 disconnected for client id "127.0.0.1:46658"
[2024-11-21T20:33:07Z TRACE epoxy_server::handle::wisp] sent ping to wisp client id "127.0.0.1:46658"
[2024-11-21T20:33:37Z TRACE epoxy_server::handle::wisp] sent ping to wisp client id "127.0.0.1:46658"
[2024-11-21T20:33:59Z DEBUG epoxy_server::handle::wisp] wisp client id "127.0.0.1:46658" multiplexor result Ok(())
[2024-11-21T20:33:59Z DEBUG epoxy_server::handle::wisp] shutting down wisp client id "127.0.0.1:46658"
[2024-11-21T20:33:59Z TRACE epoxy_server::handle::wisp] waiting for tasks to close for wisp client id "127.0.0.1:46658"
[2024-11-21T20:33:59Z DEBUG epoxy_server::handle::wisp] wisp client id "127.0.0.1:46658" disconnected
```

ProgrammerIn-wonderland:
Also wisp server Cpp sends a 0x04 Close packet back after the client sends one. It is the only server to do this, and it actually isn't in the spec. The client is supposed to assume the stream is closed immediately after sending the close packet.

Just wanted to note that on this GitHub issue

r58Playz:
Another note: WispServerCpp doesn't actually close any sockets that it forwards, violating spec, so the remote server never sees the close from the wisp client.

ProgrammerIn-wonderland:
I think it's worth looking into making a protocol extension for half closing, @ading2210 @r58Playz

chschnell:
@ProgrammerIn-wonderland

> Are you saying the other wisp servers don't support closing or don't support half closing? Sorry I think I'm misunderstanding

Sorry about the confusion. I meant to say the other WISP servers don't support closing, but it turned out I was wrong about that.

> Also close is 0x04, the reason is 0x02. Just wanted to clarify because I stated it incorrectly above in my own message

Understood. By 0x02 I always meant reason 0x02 (Voluntary stream closure).

@r58Playz

> On epoxy, you can use RUST_LOG=trace to see that it does close streams (stream uuid f9f5dcdb-5edc-40d2-9ba7-b7686b6cfc19 disconnected for client id "127.0.0.1:46658")

Thanks, I enabled tracing and indeed, I see the same message. Sorry about my confusion, I was wrong when I said it wasn't supported by epoxy-tls/wisp-js, they indeed do.

chschnell:
My [latest commit](https://github.com/copy/v86/pull/1178/commits/5ec28a2631d19617a15f1feaa4183a8981269e10) is a slight change in strategy with this close issue.

I now distingush between on_shutdown() (FIN) and on_close() (RST), this is the right way to model it in fake_network.js.

I implemented both on_shutdown() and on_close() for WISP using a CLOSE frame with reason 0x02. This does not implement half-closed TCP connections (currently not supported), but this also doesn't leak idle TCP connections.

It would be trivial to change on_shutdown() in case WISP support changes in the future.

ProgrammerIn-wonderland:
Thanks! This does make retrofitting it migh easier if the server advertises support for TCP half closing.

chschnell:
Currently there are two different fake DNS methods in V86:

1. in fake_network.js: offline, static implementation that returns `192.168.87.1` for any given DNS name
2. in wisp_network.js: uses fetch() and DoH (DNS over HTTPS), DoH server defaults to `cloudflare-dns.com` unless specified in config.doh_server

Fetch-based networking uses the first method, wisp-based the second, and I'm not sure why.

Faking DNS with DoH works just fine for both fetch- and wisp-based networking, and it returns real results.

So I would like to move the DoH-code from wisp_network.js into handle_fake_dns() (fake_network.js) and make it the default fake DNS method.

Question: Is the first method without DoH still needed or can it be removed? I think it is not needed, only fetch-based networking uses it, and with fetch() this DNS method doesn't really matter (the browser performs the DNS lookup for the URL passed to fetch()).

In case I should keep the first method I would suggest to add a new option `config.dns_method` next to `config.doh_server` with possible values `static` and `doh` (default `doh`), unless someone has a better idea.

SuperMaxusa:
> Question: Is the first method without DoH still needed or can it be removed? I think it is not needed, only fetch-based networking uses it, and with fetch() this DNS method doesn't really matter (the browser performs the DNS lookup for the URL passed to fetch()).

For fetch-based networking, there is not much need for it: https://github.com/copy/v86/pull/1061#issuecomment-2184188633, and fetch with IP address works buggy, especially if HTTPS-only mode is enabled in browser.

ProgrammerIn-wonderland:
Fetch networking uses the Host: header so the IP address given is irrelevant, to save time it just returns a dummy IP.

Since wisp actually connects to the IP it has to do the real resolution

ProgrammerIn-wonderland:
Sorry just saw your command supermaxusa, my Internet is glitching

ProgrammerIn-wonderland:
I don't think it makes much sense to add Doh to fetch personally since it's not really necessary based on how it operates (unless you're running a SUPER OLD pre Host header http client). But I do think the Doh server should be configurable

chschnell:
@SuperMaxusa @ProgrammerIn-wonderland

Thanks a lot for your input!

We all agree that DoH doesn't make much sense with fetch-based networking, my point was that it also doesn't hurt because it matters so little. But it's fine, I'll leave it at that.

In my latest [commit](https://github.com/copy/v86/pull/1178/commits/cf3b38d5ba8fe361a3cb9e6020a4b779e552061a) I moved the DoH-code from WispNetworkAdapter.send() into new function `handle_fake_dns_doh()` in fake_network.js, and only WispNetworkAdapter uses DoH, just as before. So it's only a refactoring, the code behaves unchanged. Everything is in place in case someone wants to add a config option for this, I don't see the point.

The benefit of all this is that all of the code in WispNetworkAdapter.send() that was duplicated from fake_network.js is now gone which somewhat simplifies reasoning about the fake networking control flow (it made me scratch my head quite a little).

src/browser/fake_network.js
copy: You're absolutely right, I was somehow under the impression that `data` was returned from `peek` and so already had the right length.

copy:
Thanks! This looks good to merge to me now. Let me know if there's another change you'd like to get in. I will try to rebase (to get rid of the merge commits).

chschnell:
> Thanks! This looks good to merge to me now. Let me know if there's another change you'd like to get in. I will try to rebase (to get rid of the merge commits).

Yes, let's merge it.

I see two things left to do, but I want to split them off into a new PR in a couple of weeks.

1. [replace the bus](https://github.com/copy/v86/blob/f767d4d73a7abe4300fa5dad1e68f15c250fc20b/src/browser/starter.js#L313) with direct function calls
2. further improve buffering, for example to encode ethernet packets at the latest possible stage, or to use GrowableRingbuffer in other places

I will do other things in the next week or two, but I think then I'll take another go at this, if you don't mind.

basicer:
Great work! A huge improvement over my hacked up proof of concept.

chschnell:
> Great work! A huge improvement over my hacked up proof of concept.

My pleasure! And it was just a bit rough around the edges, the core was rock-solid.

copy:
Merged manually, thanks all!

copy:
@chschnell Sure

PR #1182, opened 2024-11-10
fetch-networking: fix parsing of HTTP headers

SuperMaxusa:
Currently, I have only seen this bug in older versions of Dillo, but it's possible that other HTTP clients have it too.

**Test-case**: boot from https://distro.ibiblio.org/damnsmall/current/current.iso with enabled fetch networking and then try to open any page in Dillo, you get in the browser console:
```
TypeError: Headers.set: accept-encoding:gzip is an invalid header name.
```

dillo-bug

A raw HTTP request from Dillo looks like this:

```http
GET / HTTP/1.0
Host: copy.sh
User-Agent: Dillo/0.8.5-i18n-misc
Accept-Language: en-us, ja
Accept-Encoding:gzip
Accept-Charset:utf-8,ISO8859-1
Cookie2: $Version="1"
```

According to [RFC 2616 (section 4.2)](https://datatracker.ietf.org/doc/html/rfc2616#section-4.2), HTTP client/server can use any amount of whitespaces after separator, previous code expected only one space after the colon (separator).

### DIFF
--- a/src/browser/fetch_network.js
+++ b/src/browser/fetch_network.js
@@ -83,9 +83,9 @@ async function on_data_http(data)
 
         let req_headers = new Headers();
         for(let i = 1; i < headers.length; ++i) {
-            let parts = headers[i].split(": ");
+            let parts = headers[i].split(/:(.+)/);
             let key =  parts[0].toLowerCase();
-            let value = parts[1];
+            let value = parts[1].trim();
             if( key === "host" ) target.host = value;
             else if( key.length > 1 ) req_headers.set(parts[0], value);
         }
--- a/src/browser/fetch_network.js
+++ b/src/browser/fetch_network.js
@@ -85,7 +85,7 @@ async function on_data_http(data)
         for(let i = 1; i < headers.length; ++i) {
             const header = this.net.parse_http_header(headers[i]);
             if(!header) {
-                console.warn('The request contains an invalid header: "%s"', header);
+                console.warn('The request contains an invalid header: "%s"', headers[i]);
                 this.write(new TextEncoder().encode("HTTP/1.1 400 Bad Request\r\nContent-Length: 0"));
                 return;
             }
--- a/src/browser/fetch_network.js
+++ b/src/browser/fetch_network.js
@@ -83,19 +83,13 @@ async function on_data_http(data)
 
         let req_headers = new Headers();
         for(let i = 1; i < headers.length; ++i) {
-            let parts = headers[i].match(/^([^:]*):(.*)$/);
-            let key = "";
-            let value = "";
-            if(parts !== null && parts.length === 3) {
-                key = parts[1].toLowerCase();
-                value = parts[2].trim();
-            }
-            if(!this.net.validate_header(key, value)) {
+            const header = this.net.parse_http_header(headers[i]);
+            if(!header) {
                 this.write(new TextEncoder().encode("HTTP/1.1 400 Bad Request\r\nContent-Length: 0"));
                 return;
             }
-            if( key === "host" ) target.host = value;
-            else if( key.length > 1 ) req_headers.append(parts[1], value);
+            if( header.key.toLowerCase() === "host" ) target.host = header.value;
+            else req_headers.append(header.key, header.value);
         }
 
         dbg_log("HTTP Dispatch: " + target.href, LOG_FETCH);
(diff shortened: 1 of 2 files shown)


src/browser/fetch_network.js
```diff
@@ -83,9 +83,9 @@ async function on_data_http(data)
 
         let req_headers = new Headers();
         for(let i = 1; i < headers.length; ++i) {
-            let parts = headers[i].split(": ");
+            let parts = headers[i].split(/:(.+)/);
             let key =  parts[0].toLowerCase();
-            let value = parts[1];
+            let value = parts[1].trim();
             if( key === "host" ) target.host = value;
             else if( key.length > 1 ) req_headers.set(parts[0], value);
```
copy, line 90: Unrelated to this PR, but this should probably use `req_headers.append`

src/browser/fetch_network.js
```diff
@@ -83,9 +83,9 @@ async function on_data_http(data)
 
         let req_headers = new Headers();
         for(let i = 1; i < headers.length; ++i) {
-            let parts = headers[i].split(": ");
+            let parts = headers[i].split(/:(.+)/);
```
copy, line 86: I think `.match(/^([^:]*):(.*)$/)` would be clearer

src/browser/fetch_network.js
```diff
@@ -83,9 +83,9 @@ async function on_data_http(data)
...
             let key =  parts[0].toLowerCase();
-            let value = parts[1];
+            let value = parts[1].trim();
```
copy, line 88: This should protect against invalid headers (no colon or last character is a colon).

SuperMaxusa:
Thanks for the suggestions, I think that it's better make a properly HTTP headers validation, so my apologies that I had remake this PR. Let me know if I need to create a new PR instead.

Changes:
 - added check on empty header key/value and forbidden chars
 - send Error 400 on failed check
 - added tests for each behaviour

src/browser/fetch_network.js
```diff
@@ -83,11 +83,19 @@ async function on_data_http(data)
...
+            let key = "";
+            let value = "";
+            if(parts !== null && parts.length === 3) {
+                key = parts[1].toLowerCase();
+                value = parts[2].trim();
+            }
+            if(!this.net.validate_header(key, value)) {
+                this.write(new TextEncoder().encode("HTTP/1.1 400 Bad Request\r\nContent-Length: 0"));
+                return;
+            }
```
copy, line 96: Since we're moving things into a helper function, I'd suggest:

```javascript
           const header = this.net.parse_http_header(headers[i]);
           if(!header) {
                this.write(new TextEncoder().encode("HTTP/1.1 400 Bad Request\r\nContent-Length: 0"));
                return;
            }
```

chschnell:
Another concern I see with this HTTP request parser: Only the first fragment of the request body is taken into account, that is as many bytes that fit into the last TCP headers packet right after the end of headers, up until the end of packet. If the originating client sends a request with a longer body this will break.

After (or while) reading the headers it should be checked whether there is a Content-Length field, and if there is then as many bytes should be read from the originating client before continuing to process the request. For the sake of completeness (the originating client should be free to do whatever is allowed), also chunked body encoding should be considered when reading this reuest body.

src/browser/fetch_network.js
```diff
@@ -159,30 +153,45 @@ FetchNetworkAdapter.prototype.fetch = async function(url, options)
...
 };
 
-FetchNetworkAdapter.prototype.validate_header = function(key, value)
+FetchNetworkAdapter.prototype.parse_http_header = function(header)
 {
+    if(!header.includes(":"))
+    {
+        dbg_log("Header doesn't have a separator", LOG_FETCH);
+        return;
+    }
```
copy, line 162: Isn't this already handled by the `parts === null` check below?

SuperMaxusa:
> After (or while) reading the headers it should be checked whether there is a Content-Length field, and if there is then as many bytes should be read from the originating client before continuing to process the request.

Agreed, it might be worth adding a check if `Content-Length: 0` but the client sends data.
If I correctly understand, HTTP/1.1 client also can send chunked data: https://everything.curl.dev/http/post/chunked.html, and I'm not sure that even without `Transfer-Encoding: chunked` can send any amount of data before the connection is closed, as in this case with the server response: https://github.com/copy/v86/pull/1178#issuecomment-2470997329

chschnell:
Let me rephrase what I mean, forgive me if that was already clear to you.

Think of a longer request from the originating client (by that I mean something like HTGET.EXE from mTCP). I mean "long" in relation to the TCP frame size.

For simplicitly let's say we have a TCP max. payload size of 1500 bytes, and a HTTP request of 500 bytes header data (plus separator) and a body of another 2000 bytes, so a request of 2500 bytes total. The client's TCP stack splits this HTTP request into two TCP frames:

    0      500    1000   1500   2000   2500
    :      :      :      :      :      :
    +------+---------------------------+
    |Header|           Body            |
    +------+---------------------------+
    +--------------------+-------------+
    |    TCP frame 1     | TCP frame 2 |
    +--------------------+-------------+

[`TCPConnection.on_data_http()`](https://github.com/copy/v86/blob/15372291dbab4cfa141db26f028e0807780345cf/src/browser/fetch_network.js#L59) gets called for each TCP frame received from the originating client. The [condition](https://github.com/copy/v86/blob/15372291dbab4cfa141db26f028e0807780345cf/src/browser/fetch_network.js#L64) that checks if we have received all headers becomes true with the first TCP frame. When `TCPConnection.on_data_http()` gets called again later with the second frame, what part of the code handles that? Note that [the fetch() transaction](https://github.com/copy/v86/blob/15372291dbab4cfa141db26f028e0807780345cf/src/browser/fetch_network.js#L102) has already started at this point.

SuperMaxusa:
Thank you for your clarification, probably checking Content-Length is worth it, I will investigate that in my free time.

copy:
@SuperMaxusa It's good to merge from my point of view, any objections?

SuperMaxusa:
> It's good to merge from my point of view, any objections?

Ok, I have no objections.
@chschnell Do you have any suggestions? How about an idea to create an issue from [your comment about HTTP request on multiple TCP frames](https://github.com/copy/v86/pull/1182#issuecomment-2471756879)?

copy:
Merci!

chschnell:
> > It's good to merge from my point of view, any objections?
> 
> Ok, I have no objections. @chschnell Do you have any suggestions? How about an idea to create an issue from [your comment about HTTP request on multiple TCP frames](https://github.com/copy/v86/pull/1182#issuecomment-2471756879)?

It's fine, it can wait.

chschnell:
@copy I am very sorry but I just noticed that I broke the `cors_proxy` member in FetchNetworkAdapter, I'd need to fix a single line in fetch_network.js. What should I do, is a quick PR ok?

PR #1187, opened 2024-11-26
Fixed broken FetchNetworkAdapter.cors_proxy attribute

chschnell:
Fixed wrong reference to cors_proxy in fetch_network.js


copy:
Danke!

Issue #1188, opened 2024-11-27
Allow to embed CORS proxy URL in "fetch" relay_url


chschnell:
https://github.com/copy/v86/blob/298da237c2e8ae3b605ab60a5f873884a96ed054/src/browser/starter.js#L314-L317

My suggestion would be to change this to:

```javascript
if(relay_url.startsWith("fetch")) {
    const fetch_options = {};
    if(relay_url.length > 6 && relay_url[5] === ":") {
        fetch_options.cors_proxy = relay_url.slice(6);
    }
    this.network_adapter = new FetchNetworkAdapter(this.bus, fetch_options);
}
```

Things would behave as before when passing only `fetch` to `config.net_device.relay_url`, but when passing something like:

```
fetch:https://example.com/?url=
```

then CORS proxy `https://example.com/?url=` would be used.

copy:
Cheers!

PR #1189, opened 2024-11-27
Allow embedding of CORS proxy in fetch relay_url

chschnell:
Extends the syntax of relay_url for fetch to:

    fetch[:CORS-PROXY-URL]

where CORS-PROXY-URL is a full URL pointing to the CORS proxy server, for example:

    fetch:https://example.com/?url=

Fixes issue #1188.

### DIFF
--- a/src/browser/main.js
+++ b/src/browser/main.js
@@ -1737,6 +1737,10 @@
             settings.relay_url = $("relay_url").value;
             if(!DEFAULT_NETWORKING_PROXIES.includes(settings.relay_url)) new_query_args.set("relay_url", settings.relay_url);
         }
+        if (settings.relay_url.startsWith("fetch") && settings.relay_url.length > 6 && settings.relay_url[5] === ":") {
+            settings.cors_proxy = settings.relay_url.slice(6);
+            settings.relay_url = "fetch";
+        }
         settings.disable_audio = $("disable_audio").checked || settings.disable_audio;
         if(settings.disable_audio) new_query_args.set("mute", "1");
 
@@ -1852,6 +1856,7 @@
             net_device: {
                 type: settings.net_device_type || "ne2k",
                 relay_url: settings.relay_url,
+                cors_proxy: settings.cors_proxy
             },
             autostart: true,
 
--- a/src/browser/main.js
+++ b/src/browser/main.js
@@ -1737,7 +1737,8 @@
             settings.relay_url = $("relay_url").value;
             if(!DEFAULT_NETWORKING_PROXIES.includes(settings.relay_url)) new_query_args.set("relay_url", settings.relay_url);
         }
-        if(settings.relay_url.startsWith("fetch") && settings.relay_url.length > 6 && settings.relay_url[5] === ":") {
+        if(settings.relay_url.startsWith("fetch:"))
+        {
             settings.cors_proxy = settings.relay_url.slice(6);
             settings.relay_url = "fetch";
         }
--- a/src/browser/main.js
+++ b/src/browser/main.js
@@ -1737,7 +1737,7 @@
             settings.relay_url = $("relay_url").value;
             if(!DEFAULT_NETWORKING_PROXIES.includes(settings.relay_url)) new_query_args.set("relay_url", settings.relay_url);
         }
-        if (settings.relay_url.startsWith("fetch") && settings.relay_url.length > 6 && settings.relay_url[5] === ":") {
+        if(settings.relay_url.startsWith("fetch") && settings.relay_url.length > 6 && settings.relay_url[5] === ":") {
             settings.cors_proxy = settings.relay_url.slice(6);
             settings.relay_url = "fetch";
         }
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -311,9 +311,12 @@ V86.prototype.continue_init = async function(emulator, options)
     if(relay_url)
     {
         // TODO: remove bus, use direct calls instead
-        if(relay_url === "fetch")
-        {
-            this.network_adapter = new FetchNetworkAdapter(this.bus);
+        if(relay_url.startsWith("fetch")) {
+            const fetch_options = {};
+            if(relay_url.length > 6 && relay_url[5] === ":") {
+                fetch_options.cors_proxy = relay_url.slice(6);
+            }
+            this.network_adapter = new FetchNetworkAdapter(this.bus, fetch_options);
         }
         else if(relay_url.startsWith("wisp://") || relay_url.startsWith("wisps://"))
         {
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -311,12 +311,9 @@ V86.prototype.continue_init = async function(emulator, options)
     if(relay_url)
     {
         // TODO: remove bus, use direct calls instead
-        if(relay_url.startsWith("fetch")) {
-            const fetch_options = {};
-            if(relay_url.length > 6 && relay_url[5] === ":") {
-                fetch_options.cors_proxy = relay_url.slice(6);
-            }
-            this.network_adapter = new FetchNetworkAdapter(this.bus, fetch_options);
+        if(relay_url === "fetch")
+        {
+            this.network_adapter = new FetchNetworkAdapter(this.bus, options.net_device ? options.net_device : {});
         }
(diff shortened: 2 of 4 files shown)


src/browser/starter.js
```diff
@@ -311,9 +311,12 @@ V86.prototype.continue_init = async function(emulator, options)
...
         // TODO: remove bus, use direct calls instead
-        if(relay_url === "fetch")
-        {
-            this.network_adapter = new FetchNetworkAdapter(this.bus);
+        if(relay_url.startsWith("fetch")) {
+            const fetch_options = {};
+            if(relay_url.length > 6 && relay_url[5] === ":") {
+                fetch_options.cors_proxy = relay_url.slice(6);
+            }
+            this.network_adapter = new FetchNetworkAdapter(this.bus, fetch_options);
```
copy, line 319: Thinking about this again, I think this code belongs in `browser/main.js`. The `V86` constructor is supposed to be the high-level api, while `browser/main.js` implements the more stringly-typed frontend.

chschnell: Forgive me if I'm wrong, but isn't main.js the entry point for your web page and V86 embedding? In my own embedding I call the V86 constructor directly without using anything from main.js (to my knowledge), so I wouldn't gain anything from my own PR, or would I? :)

How about this: We currently have this in starter.js:

```javascript
if(relay_url.startsWith("fetch"))
{
    const fetch_options = {};
    if(relay_url.length > 6 && relay_url[5] === ":") {
        fetch_options.cors_proxy = relay_url.slice(6);
    }
    this.network_adapter = new FetchNetworkAdapter(this.bus, fetch_options);
}
else if(relay_url.startsWith("wisp://") || relay_url.startsWith("wisps://"))
{
    this.network_adapter = new WispNetworkAdapter(relay_url, this.bus, options);
}
else
{
    this.network_adapter = new NetworkAdapter(relay_url, this.bus);
}
```

Note that `relay_url` is passed as the first argument to the constructors of `WispNetworkAdapter` and `NetworkAdapter`.

Why not do the same with `FetchNetworkAdapter` and do away with its `fetch_options` alltogether.

The rule would then be that it is the NetworkAdapter's responsibility to decode its `relay_url` according to its internal rules, this knowledge doesn't need to leak outside the respective NetworkAdapter classes.

And the code from above would simplify to (nice and clean):

```javascript
if(relay_url.startsWith("fetch"))
{
    this.network_adapter = new FetchNetworkAdapter(relay_url, this.bus);
}
else if(relay_url.startsWith("wisp://") || relay_url.startsWith("wisps://"))
{
    this.network_adapter = new WispNetworkAdapter(relay_url, this.bus, options);
}
else
{
    this.network_adapter = new NetworkAdapter(relay_url, this.bus);
}
```

copy: Your second example looks worse to me than the current version in this PR, because it splits the parsing over multiple files.

You're making a valid point though, which is that the the existing implementation already does parsing to a certain degree.

I'm mostly worried that we end up with:
```js
    net_device: {
        relay_url: "ws://.../?ipv4=off&dns=1.2.3.4&other-config=..."
    }
```

When, in my opinion, a nicer JavaScript API would look like this:

```js
    net_device: {
        type: "relay",
        relay_url: "ws://.../",
        ipv4: false,
        dns: "1.2.3.4",
    }
```

> Forgive me if I'm wrong, but isn't main.js the entry point for your web page and V86 embedding? In my own embedding I call the V86 constructor directly without using anything from main.js (to my knowledge), so I wouldn't gain anything from my own PR, or would I? :)

The V86 constructor should provide an option to specify the cors proxy.

chschnell: Ah, I see what you mean now.

So I moved the fetch URL decoding out into main.js (and also into my own embedding).

I am not entirely sure where to put `cors_proxy` though, I've put it under `net_device`:

```javascript
let v86_config = {
    net_device: {
        type: "ne2k" | "virtio",
        relay_url: "fetch" | "ws://..." | "wisp://...",
        cors_proxy: undefined | "https://..."
    },
    // ...
};
let v86 = new V86(v86_config);
```

In starter.js, I decided to pass `options.net_device` as the `options` argument to the `FetchNetworkAdapter` constructor, meaning this constructor expects an object like (all members optional):

```javascript
options = {
    id: undefined | <int>,
    router_mac: undefined | <str>,
    router_ip: undefined | <str>,
    vm_ip: undefined | <str>,
    cors_proxy: undefined | "https://...",
}
```

I am pointing this out because this is not how it is implemented for `WispNetworkAdapter`, the whole `options` object and not just its member `options.net_device` is passed to its constructor in starter.js.

I'd suggest to also pass `options.net_device` instead of `options` to the `WispNetworkAdapter` constructor.

Here's the combined set of all possible members for `options.net_device` that we would have then (for both constructors):

```javascript
let v86_config = {
    net_device: {
        type: "ne2k" | "virtio",
        relay_url: "fetch" | "ws://..." | "wisp://...",
        id: undefined | <int>,
        router_mac: undefined | <str>,
        router_ip: undefined | <str>,
        vm_ip: undefined | <str>,
        cors_proxy: undefined | "https://..."
        dns_method: undefined | "static" | "doh",
        doh_server: undefined | <str>,
    },
    // ...
};
let v86 = new V86(v86_config);
```

chschnell:
I don't understand how the latest build error relates to this PR.

FunnyCorgi:
> Feel free to correct me if I'm wrong.
@SuperMaxusa  I don't think your wrong about this. It does seem excessive now that I think about it. I'm just wondering if fetch based networks that work like the websocket do are faster(but then we're basically doing the exact same thing as the websocket).

chschnell:
@copy I'll rephrase what I wrote eralier.

In [`starter.js`](https://github.com/copy/v86/blob/298da237c2e8ae3b605ab60a5f873884a96ed054/src/browser/starter.js#L320), `WispNetworkAdapter` is instantiated like this:

```javascript
this.network_adapter = new WispNetworkAdapter(relay_url, this.bus, options);
```

Note that the `options` argument is the unmodifed object that was passed to the V86 constructor.

This is how `options` is then used in the [constructor](https://github.com/copy/v86/blob/298da237c2e8ae3b605ab60a5f873884a96ed054/src/browser/wisp_network.js#L18-L22) of `WispNetworkAdapter`:

```javascript
this.id = config.id || 0;
this.router_mac = new Uint8Array((config.router_mac || "52:54:0:1:2:3").split(":").map(function(x) { return parseInt(x, 16); }));
this.router_ip = new Uint8Array((config.router_ip || "192.168.86.1").split(".").map(function(x) { return parseInt(x, 10); }));
this.vm_ip = new Uint8Array((config.vm_ip || "192.168.86.100").split(".").map(function(x) { return parseInt(x, 10); }));
this.masquerade = config.masquerade === undefined || !!config.masquerade;
```

My only question is whether this is correct or if `WispNetworkAdapter` should be constructed in `starter.js` like this instead:

```javascript
this.network_adapter = new WispNetworkAdapter(relay_url, this.bus, options.net_device);
```

Are `id`, `router_mac` etc. supposed to be members of `options` or of `options.net_device`?

chschnell:
@SuperMaxusa Thank you for you info about rust.

In my opinion it should only be a single, optional CORS proxy in the fetch URL.

copy:
> My only question is whether this is correct or if WispNetworkAdapter should be constructed in starter.js like this instead:
> `this.network_adapter = new WispNetworkAdapter(relay_url, this.bus, options.net_device);`

Yes, `WispNetworkAdapter` should receive `options.net_device`, and `id`, `router_mac`, etc. should be on `options.net_device`.

src/browser/main.js
```diff
@@ -1737,6 +1737,10 @@
             settings.relay_url = $("relay_url").value;
             if(!DEFAULT_NETWORKING_PROXIES.includes(settings.relay_url)) new_query_args.set("relay_url", settings.relay_url);
         }
+        if(settings.relay_url.startsWith("fetch") && settings.relay_url.length > 6 && settings.relay_url[5] === ":") {
```
copy, line 1740:
```suggestion
        if(settings.relay_url.startsWith("fetch:"))
        {
```

copy:
@chschnell Good to merge?

chschnell:
> @chschnell Good to merge?

Sure!

Issue #1094, opened 2024-07-27
Is it possible to network v86 instances together inside the browser? 


Darin755:
I am looking for a way to network multiple instances together without having to do anything server side. This will be a part of a virtual networking lab for educational purposes. It might be possible to use serial to setup an connection but it would be tricky and not ideal. Ideally it would be nice to have a local network interface that can be used to make a virtual switch in the browser. You could have some sort of hook that runs a handler program when a frame is received. From there I could write a proper switch that could do routing based on mac.

copy:
Yes, see https://github.com/copy/v86/blob/master/examples/two_instances.html (haven't tested it in a while, but it should still be working)

ProgrammerIn-wonderland:
> I am looking for a way to network multiple instances together without having to do anything server side. This will be a part of a virtual networking lab for educational purposes. It might be possible to use serial to setup an connection but it would be tricky and not ideal. Ideally it would be nice to have a local network interface that can be used to make a virtual switch in the browser. You could have some sort of hook that runs a handler program when a frame is received. From there I could write a proper switch that could do routing based on mac.

I've been playing around with the new fetch based network driver and have to say, its really friendly to begin work on (I even made my own handler for TCP in a day that uses https://github.com/MercuryWorkshop/wsproxy and is much faster than the relay based networking available), you could probably play around with that + webrtc if you're looking for cross computer communication.

alternatively you could hook virtio console and serial and send messages that way, it would be a form of networking, just not through the NIC

Darin755:
> Yes, see https://github.com/copy/v86/blob/master/examples/two_instances.html (haven't tested it in a while, but it should still be working)

Is it possible to have multiple interfaces? I am looking to have a mini web based network. For now I am looking into possibly writing a switch program that matches Mac addresses.

SuperMaxusa:
> Is it possible to have multiple interfaces?

There are multiple NE2K support, if I correctly understand what you are need: https://github.com/copy/v86/pull/1061/commits/692b9ad01028b907b15b7f8176265fd5e5551342

ProgrammerIn-wonderland:
> Is it possible to have multiple interfaces? I am looking to have a mini web based network. For now I am looking into possibly writing a switch program that matches Mac addresses.

I think the switch is the best idea here, there doesn't seem to be a way to create multiple NICs in v86 in starter.js, you'd probably need to do substantial modifications

Darin755:
[...]

PR #1196, opened 2024-12-13
examples: add "Networking between browser windows/tabs using the Broadcast Channel API"

SuperMaxusa:
From #1094, including a few toy network services (DNS, HTTP, Telnet servers) for testing.

v86-broadcast-api


copy:
Muito obrigado!

SuperMaxusa:
De nada!

PR #1198, opened 2024-12-19
Stop reconnecting websocket after NetworkAdapter.destroy() was called

chschnell:
Stops NetworkAdapter from reconnecting to its Websocket proxy server after V86 was shut down.

### DIFF
--- a/src/browser/network.js
+++ b/src/browser/network.js
@@ -25,6 +25,7 @@ function NetworkAdapter(url, bus, id)
     this.reconnect_interval = 10000;
     this.last_connect_attempt = Date.now() - this.reconnect_interval;
     this.send_queue_limit = 64;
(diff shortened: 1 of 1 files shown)


chschnell:
I came across this when writing a websockproxy server.

When shutting down V86 while using NetworkAdapter, I see at the server side that the WebSocket connection gets closed by the browser.

But a few seconds later the browser creates a new WebSocket connection, even though V86 is off, and this new connection then just sits around idle.

When I restart V86 a new NetworkAdapter instance is created which creates its own WebSocket connection, but the idle connection keeps sitting there.

This PR fixes that by preventing Websocket reconnects after NetworkAdapter.destroy() was called.

copy:
Nice find, thanks!

PR #1201, opened 2024-12-29
ci: update actions from v3 to latest v4

SuperMaxusa:
See https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/ and https://github.com/actions/cache/discussions/1510


copy:
Cheers!

PR #1216, opened 2025-01-19
Add new network type "inbrowser"

chschnell:
Added a new class `InBrowserNetworkAdapter` which provides a simple [`BroadcastChannel`](https://developer.mozilla.org/de/docs/Web/API/BroadcastChannel)-based network to v86 guests.

This networking method allows multiple v86 instances within the same browser process to communicate with each other like in a physical network, but without access to any external networks (it uses the same method as demonstrated in `examples/broadcast-network.html`).

The URL scheme (for `config.net_device.relay_url`) of this new networking method is `inbrowser`.

Added it to the v86 web interface in `main.js` because it works out of the box and is a great method to get started with v86 networking.

This PR is only meant to be a proposal, everything is up for debate, if this proposal should get rejected I'd be absolutely fine with the decision, of course.

### DIFF
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -315,9 +315,9 @@ V86.prototype.continue_init = async function(emulator, options)
         {
             this.network_adapter = new FetchNetworkAdapter(this.bus, options.net_device);
         }
-        else if(relay_url === "localhub")
+        else if(relay_url === "inbrowser")
         {
-            this.network_adapter = new LocalhubNetworkAdapter(this.bus, options.net_device);
+            this.network_adapter = new InBrowserNetworkAdapter(this.bus, options.net_device);
         }
         else if(relay_url.startsWith("wisp://") || relay_url.startsWith("wisps://"))
         {
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -317,6 +317,7 @@ V86.prototype.continue_init = async function(emulator, options)
         }
         else if(relay_url === "inbrowser")
         {
+            // NOTE: experimental, will change when usage of options.net_device gets refactored in favour of emulator.bus
             this.network_adapter = new InBrowserNetworkAdapter(this.bus, options.net_device);
         }
         else if(relay_url.startsWith("wisp://") || relay_url.startsWith("wisps://"))
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -315,6 +315,10 @@ V86.prototype.continue_init = async function(emulator, options)
         {
             this.network_adapter = new FetchNetworkAdapter(this.bus, options.net_device);
         }
+        else if(relay_url === "localhub")
+        {
+            this.network_adapter = new LocalhubNetworkAdapter(this.bus, options.net_device);
+        }
         else if(relay_url.startsWith("wisp://") || relay_url.startsWith("wisps://"))
         {
             this.network_adapter = new WispNetworkAdapter(relay_url, this.bus, options.net_device);
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,8 @@ CORE_FILES=const.js config.js io.js main.js lib.js buffer.js ide.js pci.js flopp
 LIB_FILES=9p.js filesystem.js jor1k.js marshall.js
 BROWSER_FILES=screen.js keyboard.js mouse.js speaker.js serial.js \
 	      network.js starter.js worker_bus.js dummy_screen.js \
-	      fake_network.js wisp_network.js fetch_network.js print_stats.js filestorage.js
+	      localhub_network.js fake_network.js wisp_network.js fetch_network.js \
+          print_stats.js filestorage.js
 
 RUST_FILES=$(shell find src/rust/ -name '*.rs') \
 	   src/rust/gen/interpreter.rs src/rust/gen/interpreter0f.rs \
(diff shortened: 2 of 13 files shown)


src/browser/starter.js
```diff
@@ -315,6 +315,10 @@ V86.prototype.continue_init = async function(emulator, options)
         {
             this.network_adapter = new FetchNetworkAdapter(this.bus, options.net_device);
         }
+        else if(relay_url === "localhub")
+        {
+            this.network_adapter = new LocalhubNetworkAdapter(this.bus, options.net_device);
```
copy, line 320: In terms the v86 API, I'm not too happy about the open-endedness of this network configuration (e.g., what if the user wants to mix this with web workers?). I'd prefer if users of the API used the `net0-{send,receive}` API instead.

Could you leave a comment here (or above the constructor) saying that this is experiment and might change at any time?

copy (commented):
This looks useful as a feature for the website. I'm not sure about the name, how about `broadcast_channel` or `browser_tab` or similar?

SuperMaxusa:
Maybe it makes sense to rewrite `examples/broadcast-network.html` to this network type?

chschnell:
> This looks useful as a feature for the website. I'm not sure about the name, how about `broadcast_channel` or `browser_tab` or similar?

Happy to hear that!

I'd also like to find a better name for it than `localhub`, my current favorite would be `inbrowser`.

I want to change the naming only once though, I'll wait with that until I get a final word.

src/browser/starter.js
chschnell: Using `net_device` can be entirely avoided here, perhaps that's better than a comment.

LocalhubNetworkAdapter's constructor only cares about a single optional member in `net_device` which is `net_device.id`, and this defaults to `0` if not defined.

This member `net_device.id` is used in other NetworkAdapter classes to create separate networks by using distinct bus message names `net<ID>-{send,receive}`, and I thought LocalhubNetworkAdapter should follow and do the same. Should it not?

I could reduce the constructor's arguments to (pseudocode for clarity):

```
function LocalhubNetworkAdapter(BusConnector bus, int id=0)
```

and then call it in `starter.js` like this (I don't know where else to take `id` from than from `options.net_device.id`):

```
this.network_adapter = new LocalhubNetworkAdapter(this.bus, options.net_device.id);
```

Or I could remove this `id` alltogether.

I'm not sure which direction to take here, but I'd be happy to change this however you prefer.

chschnell:
> Maybe it makes sense to rewrite `examples/broadcast-network.html` to this network type?

I wasn't sure whether to touch it, but now it's clear since you're the author of `examples/broadcast-network.html`.

I changed it accordingly, see commit 2dc962c.

By the way, it would be really nice to have `dnsmasq` in the `buildroot` image (`udhcpc` is already there). Perhaps we could also put a script in there to ease static network configuration including an option to spin up a DHCP server.

SuperMaxusa:
> By the way, it would be really nice to have `dnsmasq` in the `buildroot` image (`udhcpc` is already there). Perhaps we could also put a script in there to ease static network configuration including an option to spin up a DHCP server.

Good idea, I'm also looking at the busybox' [`udhcpd`](https://boxmatrix.info/wiki/Property:udhcpd_(lincmd)) server, it's also small and might be good for simple testing.

chschnell:
> Good idea, I'm also looking at the busybox' [`udhcpd`](https://boxmatrix.info/wiki/Property:udhcpd_(lincmd)) server, it's also small and might be good for simple testing.

I didn't know about the smaller `udhcpd` in busybox, thanks for pointing it out! Its [`udhcpd.conf`](https://udhcp.busybox.net/udhcpd.conf) covers all that is needed (plus a lot more) which makes it much preferrable to `dnsmasq`.

src/browser/starter.js
copy: I think the implementation of LocalhubNetworkAdapter is fine (including how the net_device or id is passed around). No need to change anything.

I'd like to leave a comment here so that in the future I can deprecate or change `{ network_relay_url: "localhub" }` and recommend users of the v86 API to use `emulator.bus` instead.

copy:
> Maybe it makes sense to rewrite examples/broadcast-network.html to this network type?

Reasonable idea, but I prefer showing people interested in v86 the `net0-receive` / `net0-send` api. @chschnell Would you mind reverting this change?

> I'd also like to find a better name for it than localhub, my current favorite would be inbrowser.

`inbrowser` sounds good.

chschnell:
> > Maybe it makes sense to rewrite examples/broadcast-network.html to this network type?
> 
> Reasonable idea, but I prefer showing people interested in v86 the `net0-receive` / `net0-send` api. @chschnell Would you mind reverting this change?
> 
> > I'd also like to find a better name for it than localhub, my current favorite would be inbrowser.
> 
> `inbrowser` sounds good.

All done 👍

src/browser/starter.js
chschnell: Done.

copy:
Thanks!

chschnell:
> Thanks!

You're welcome.

I believe the failed test when you commited this PR stems from the asynchronous fetch() response implementation that I made a couple of weeks ago. The [assert()](https://github.com/copy/v86/blob/e2b0badfcd1f7829c7e83010834453b2db2ac79d/tests/devices/fetch_network.js#L153) in tests/devices/fetch_network.js fails because only the beginning of the response has been received at this point.

copy:
> I believe the failed test when you commited this PR stems from the asynchronous chunk-by-chunk fetch() response implementation that I made a couple of weeks ago. The [assert()](https://github.com/copy/v86/blob/e2b0badfcd1f7829c7e83010834453b2db2ac79d/tests/devices/fetch_network.js#L153) in tests/devices/fetch_network.js fails because only the beginning of the response has been received at this point.

To me it looks like `abort` is called from wasm code, so it's unrelated to the network code. It just happens to happen in the middle of printing the response.

Issue #1071, opened 2024-07-05
Linux - Synchronize NumLock, CapsLock & ScrollLock


evonox:
NumLock, CapsLock & ScrollLock work correctly when the terminal window is active. However, if I click somewhere else, change the state of these keys and then click back to the terminal window, it works in the opposite direction.

I suggest upon re-activating the terminal window to check the state of these flags and send to the terminal virtual e.g. NumLock press if it is not synchronized. To my surprise, the current API of browsers allows to check the state of NumLock / CapsLock / ScrollLock - https://www.w3schools.com/jsref/event_mouse_getmodifierstate.asp

copy:
A PR would be welcome.

evonox:
Ok, I will prepare Pull Request for this issue.

copy:
I'm not sure there will be a satisfying solution for this, since we don't know the state of these flags in the guest OS (even if we knew the initial state on boot for every OS, it could be changed by a different mechanism, e.g. by remapping the key inside of the guest).

A proper solution would be a guest agent to synchronise this, but it's probably not worth the effort. So even though this issue is valid, I will close this.

PR #1218, opened 2025-01-21
Synchronize NumLock, CapsLock & ScrollLock

vip-delete:
This PR is fixing #1071


copy:
Let's close this, see https://github.com/copy/v86/issues/1071#issuecomment-2668639004

PR #1219, opened 2025-01-23
loop unrolling in pack_memory

rndmcnlly:
To improve the state saving time for very large and sparse memories, we unroll the loop that checks for non-zero pages to work with blocks of 64 bytes per iteration.

Context: Towards implementing [playable quotes](https://adamsmith.as/papers/fdg2023_tenmile.pdf) for PC games, I want to take whole-system snapshots quite frequently (e.g. once or more per second). I found that the scan for non-zero pages was dominating the overall time spent in saving state. It looks like the code was already someone optimized for speed (by working with mem32 rather than mem8), so this enhancement, while somewhat ugly, might fit the surrounding code. In microbenching with a VM with 128MB of guest memory (and an M3 Max host processor) this lead to a ~2x speedup for the loop in question. Deeper unrolling helps slightly more on my processor, but that is probably overspecializing for my wide cache lines (128 bytes). The code here only exploits 64-byte lines.

### DIFF
--- a/src/cpu.js
+++ b/src/cpu.js
@@ -632,9 +632,18 @@ CPU.prototype.pack_memory = function()
         const view = this.mem32s.subarray(offset >> 2, offset + 0x1000 >> 2);
         let is_zero = true;
 
-        for(let i = 0; i < view.length; i++)
+        // partial loop unrolling to check 64 bytes per iteration
+        for(let i = 0; i < view.length; i += 8)
         {
-            if(view[i] !== 0)
+            if(   view[i+0]
+                | view[i+1]
+                | view[i+2]
+                | view[i+3]
+                | view[i+4]
+                | view[i+5]
+                | view[i+6]
+                | view[i+7]
+                !== 0)
             {
                 is_zero = false;
                 break;
--- a/src/cpu.js
+++ b/src/cpu.js
@@ -642,8 +642,7 @@ CPU.prototype.pack_memory = function()
                 | view[i+4]
                 | view[i+5]
                 | view[i+6]
-                | view[i+7]
-                !== 0)
+                | view[i+7])
             {
                 is_zero = false;
                 break;


src/cpu.js
```diff
@@ -632,9 +632,18 @@ CPU.prototype.pack_memory = function()
...
-            if(view[i] !== 0)
+            if(   view[i+0]
+                | view[i+1]
+                | view[i+2]
+                | view[i+3]
+                | view[i+4]
+                | view[i+5]
+                | view[i+6]
+                | view[i+7]
+                !== 0)
```
copy, line 646: `!==` has a higher precedence than `|`, so this doesn't evaluate as expected (although it doesn't cause a problem in this case).

I'd suggest removing the `!==`.

copy (commented):
Thanks, this looks good to merge, modulo a minor suggestion below.

Optionally, if you want even better performance, you could try porting this to rust. The compiler should be able to find SIMD instructions for code like this.

If you're interested, here are some pointers: 

- Add a new `#[no_mangle]` function to src/rust/cpu/memory.rs
- Rust already has access to `mem8` (see the other functions in memory.rs)
- In cpu.js, search for wasm_patch and add the new import

src/cpu.js
rndmcnlly: Good catch!

rndmcnlly:
I looked into moving this loop into rust, and it is indeed much faster (though the degree of speedup is strongly host processor dependent). Here's an example of the rust compiler's intelligence in this scenario.

A fairly readable loop to or together all of the bytes in a page:

```rust
for i in 0..num_pages {
    let nonzero = bytes[(i * 0x1000) as usize..((i + 1) * 0x1000) as usize]
        .iter()
        .fold(0, |acc, &byte| acc | byte) != 0;
    *bitmap.add((i / 8) as usize) |= (nonzero as u8) << (i % 8);
    nonzero_page_count += nonzero as u32;
}
```

Even though the loop appears to work byte-by-byte using high-level `iter` and `fold` operations, I can confirm that the compiler expresses the inner loop as four `v128.load` instructions chained unconditionally. That's processing 64 *bytes* per iteration without the person reading the code to think about SIMD at all. Not bad. (And with slightly uglier code it goes even faster.)

Communicating the resulting bitmap back to JavaScript is somewhat tedious, however. Space for the bitmap needs to be allocated, JS needs to get the base and bounds of the zone, and JS needs a way to ask for the nonzero page scan to be run.

I'm holding off on submitting this rust version until I have a better overall design. If I can find a few different bitmap related rust routines, it would make sense to follow the design used for ZStd in V86 where JS can create/free an extra context object. If `bitmap_set_from_nonzero_pages` is the only routine, however, the complexity of the design doesn't feel justified.

The slightly manually-unrolled version in this PR seems to be a good balance for performance for complexity.

rndmcnlly:
Closing because I messed up my fork while trying to clean up with a force push, will open a new PR with just the clean code.

PR #1223, opened 2025-01-25
tcp connect

basicer:
Allows connection to servers running inside the VM from javascript.

- Move fetch_networking tests from linux4 to buildroot68
- Add a test talking to a TCP server inside the VM
- Add back fetch_adapter.fetch override for fetch()
- Fix fetch mocking test
- Fix virtio_net's mac address preservation logic.


copy:
Thanks!

PR #1236, opened 2025-01-31
Change `instance of Array` to `Array.isArray`

basicer:
This fixes a problem saving state in nodejs with `nodejs-loader.mjs` and is more Idiomatic.


copy:
Cheers.

PR #1237, opened 2025-01-31
Optimize pack_memory by moving zero check to rust.

basicer:
As suggested by @copy  in #1219 move the check if a memory page is all zeros to rust.  Executes on 64bit chunks, which is likely not as good simd could do, but is still a 5x speedup in the included snapshot benchmark on my machine.

Co-authroed-by: @rndmcnlly for the original idea

In the final version the bitmap can be done in rust as suggested, however until thats worked out we might as well get this basic acceleration in as its such a huge improvement over whats there now.

### DIFF
--- a/src/rust/cpu/memory.rs
+++ b/src/rust/cpu/memory.rs
@@ -310,3 +310,15 @@ pub unsafe fn mmap_write128(addr: u32, v0: u64, v1: u64) {
         )
     }
 }
+
+#[no_mangle]
+pub unsafe fn is_memory_zeroed(addr: u32, length: u32) -> bool {
+    dbg_assert!(addr % 8 == 0);
+    dbg_assert!(length % 8 == 0);
+    for i in (addr..addr + length).step_by(8) {
+        if *(mem8.offset(i as isize) as *const i64) != 0 {
+            return false;
+        }
+    }
+    return true;
+}
--- a/tests/benchmark/snapshot.js
+++ b/tests/benchmark/snapshot.js
@@ -1,0 +1,63 @@
+#!/usr/bin/env node
+"use strict";
+
+const BENCH_COLLECT_STATS = +process.env.BENCH_COLLECT_STATS;
+
+const V86 = require(`../../build/${BENCH_COLLECT_STATS ? "libv86-debug" : "libv86"}.js`).V86;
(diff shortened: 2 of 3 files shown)


rndmcnlly:
Excellent! A few things I'd like to praise about this design:

- The design allows flexible selection of memory range, so it can be reused elsewhere to check things that aren't 4096KiB pages or aren't part of guest physical memory.
- In the case of supporting very sparse memories, someone could use this new function to scan for nonzeros at a coarse granularity (e.g. 1MiB blocks) before doing a finer granularity scan (e.g. 512B blocks) within the nonzero coarse blocks.
- The function can be monkey patched from JavaScript if someone wants to temporarily substitute a different zero checking method. This matters for my application where I'll want to temporarily replace it with one that instantly return 0 to help make a kind of lightweight snapshot (one that doesn't include any memory).

basicer:
I tried a few other configurations just to play around.  One theory was that it's faster to just or all the memory and check at the end so the worst case time which happens quite often is minimized and the compiler will more aggressively simd as well as unroll the loop.  I also tried a version where I checked ro early out on the first dword isn't 0, and then just commit to oring the rest as fast as possible.  I couldn't see much of an impact between them in the benchmark so kept this simple version.

tests/benchmark/snapshot.js
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env node
```
copy, line 1: Could you `chmod +x` this file?

src/rust/cpu/memory.rs
```diff
@@ -310,3 +310,15 @@ pub unsafe fn mmap_write128(addr: u32, v0: u64, v1: u64) {
         )
     }
 }
+
+#[no_mangle]
+pub unsafe fn is_memory_zeroed(addr: u32, length: u32) -> bool {
+    dbg_assert!(addr % 8 == 0);
+    dbg_assert!(length % 8 == 0);
+    for i in (addr..addr + length).step_by(8) {
+        if (ptr::read(mem8.offset(i as isize) as *const i64)) != 0 {
```
copy, line 319:
```suggestion
        if *(mem8.offset(i as isize) as *const i64) != 0 {
```

copy:
Thanks!

PR #1244, opened 2025-02-02
Expanding the V86 networking documentation

chschnell:
Many changes to the current networking documentation.

Added new terms *network backend* and *backend URL* as they stand for central concepts in the V86 networking architecture, and I found it a lot easier and clearer to write about the topics using them.


chschnell:
Will resubmit without any merges.

PR #1247, opened 2025-02-08
Fake Networking TCP Improvements

basicer:
- No need for handle now tha we have TCPConnection
- Add a function to probe if a port is open on the VM without finishing the three way TCP handshake.

### DIFF
--- a/src/browser/fetch_network.js
+++ b/src/browser/fetch_network.js
@@ -47,7 +47,7 @@ FetchNetworkAdapter.prototype.on_tcp_connection = function(packet, tuple)
         let conn = new TCPConnection();
         conn.state = TCP_STATE_SYN_RECEIVED;
         conn.net = this;
-        conn.on_data = on_data_http;
+        conn.on("data", on_data_http);
         conn.tuple = tuple;
         conn.accept(packet);
         this.tcp_conn[tuple] = conn;
@@ -59,6 +59,11 @@ FetchNetworkAdapter.prototype.on_tcp_connection = function(packet, tuple)
 FetchNetworkAdapter.prototype.connect = function(port)
 {
     return fake_tcp_connect(port, this);
+};
+
+FetchNetworkAdapter.prototype.tcp_probe = function(port)
+{
+    return fake_tcp_probe(port, this);
 };
 
 /**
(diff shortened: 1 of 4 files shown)


tests/devices/fetch_network.js
```diff
@@ -107,8 +107,13 @@ const tests =
         allow_failure: true,
         start: async () =>
         {
+            let open = await emulator.network_adapter.tcp_probe(80);
+            assert(!open, "Probe shows port not open");
             emulator.serial0_send("echo -n hello | socat TCP4-LISTEN:80 - && echo -e done\\\\tlisten\n");
             await wait(1000);
+            open = await emulator.network_adapter.tcp_probe(80);
+            assert(open, "Probe shows port open, but does not show as a connection");
```
copy, line 115: These assertion messages sound confusing. Should the first one be "expected port to be closed" and the second one "expected port to be open"?

(this isn't written by an llm, is it?)

src/browser/fetch_network.js
```diff
@@ -61,6 +61,11 @@ FetchNetworkAdapter.prototype.connect = function(port)
     return fake_tcp_connect(port, this);
 };
 
+FetchNetworkAdapter.prototype.tcp_probe = function(port)
```
copy, line 64: Out of curiosity, what do you use this for?

tests/devices/fetch_network.js
basicer: It will be an exciting day when the LLMs are good enough to write devices for v86, but nah this is just my confusing grammar I guess!  I copied the bottom assert to test the opposite case when it was just "Prob shows port open" to "Probe shows port not open" with out resolving that not open means closed.

src/browser/fetch_network.js
basicer: Great question!  In v86 I have some tech worked out so that the service worker can receive web requests, and then route them from the browser into an httpd server running in the VM. I like the classic look of it just being a terminal, so it uses this to probe common ports in the background.  When it finds one open it adds a split pane with an iframe.  You don't want to actually initiate the connections, as that can often confuse servers, so similar to nmap, we dont send SYN ACK so the user-space program never sees the probing connections.

copy: Cool stuff, looking forward to playing around with that.

Issue #103, opened 2016-07-26
Implement single stepping


darkf:
I'd like to construct an x86 emulator+debugger using V86, but it does not support single-stepping, which is quite necessary for it.

copy:
Here's something to play with:

```javascript
> cpu = new CPU
CPU {memory_size: 0, segment_is_null: Array[0], segment_offsets: Array[0], segment_limits: Array[0], tlb_data: Int32Array[1048576]…}
> cpu.create_memory(1024 * 1024 * 16)
undefined
> cpu.memory.mem8[0] = 0x40; // inc ax
64
> cpu.instruction_pointer
0
> cpu.reg32s[reg_eax]
0
> cpu.cycle()
undefined
> cpu.reg32s[reg_eax]
1
> cpu.instruction_pointer
1
```
Use [this script](https://github.com/copy/v86/blob/master/loader.js) to load the required files. None of this is stable now, we'll need to think how to expose internals.

We might need to write some functions similar to this (in order to make sure the names don't get mangled and we can change `reg32s` at a later point`):

```javascript
/** @export */
CPU.prototype.get_get32s = function(n)
{
    return this.reg32s[n];
};
```

almosnow:
I'm trying to follow your snippet on Node. I can't get it to work.

When doing this:

    var CPU = new require('./v86/build/libv86.js').CPU();
    CPU.mem8[0] = 0x40;
    CPU.cycle();

It throws an exception like:

    TypeError: this.memory_map_read8[(a >>> MMAP_BLOCK_BITS)] is not a function
        at CPU.mmap_read8 (libv86.js:254:947)
        at CPU.read8 (libv86.js:256:331)
        at CPU.read_imm8 (502:243)

PR #1255, opened 2025-02-13
Export `cycle_internal` for single stepping

donno2048:
Using a `cycle`s function approach was suggested in #103 as a single-stepping solution.

The exported `cycle` was removed however.

I think it could be useful to export such a function, maybe the `cycle_internal` should be having some wrapping before it can be shipped, but I think it'd beneficial overall.

### DIFF
--- a/src/cpu.js
+++ b/src/cpu.js
@@ -220,6 +220,8 @@ CPU.prototype.wasm_patch = function()
         return f;
     };
 
+    this.cycle_internal = get_import("cycle_internal");
+
     this.reset_cpu = get_import("reset_cpu");
 
     this.getiopl = get_import("getiopl");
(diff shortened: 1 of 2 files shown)


copy:
I think exposing `cycle_internal` is sub-optimal for two reasons:

- It doesn't really single step, e.g. the jitted code will usually execute more than one instruction
- It's a big function that is called often, so exporting it may negatively affect inlining decisions and/or code size

What you really want is probably something like this:

```rust
#[no_mangle]
unsafe fn single_step() {
    *previous_ip = *instruction_pointer;
    let opcode = match read_imm8() {
        Ok(x) => x,
        Err(()) => return,
    };
    run_instruction(opcode | (*is_32 as i32) << 8);
    *instruction_counter += 1;
}
```

Note that this may still execute more than one instruction in a few cases, e.g. after `STI` or `REP` prefixes (I wouldn't mind a flag to disable that, if you really need it).

donno2048:
> I think exposing `cycle_internal` is sub-optimal for two reasons:
> 
> - It doesn't really single step, e.g. the jitted code will usually execute more than one instruction

I guess, my use-case was more "clocking" over single-stepping so it didn't matter, but in general, it may not be a good decision...

donno2048:
Just to adjust the clock speed of the "cpu" to fit better the clock speed required by the program

copy:
> Just to adjust the clock speed of the "cpu" to fit better the clock speed required by the program

That sounds useful to other people, I'd suggest to upstream that feature directly rather than exposing `cycle_internal`. Let's close this PR.

PR #1261, opened 2025-02-21
fix: Invalid invocation of fetch()

felixrieseberg:
The original code in `fetch_network` binds `fetch` to `FetchNetworkAdapter`. In many modern browsers (including Chromium), binding `fetch` results in a `Illegal invocation` error.

An alternative solution to fix this would be to just call `globalThis.fetch()` or `fetch()` directly in the adapter but I didn't want to touch on any toes in case having a method on the prototype is useful for anyone.


copy:
Thanks!

> An alternative solution to fix this would be to just call globalThis.fetch() or fetch() directly in the adapter but I didn't want to touch on any toes in case having a method on the prototype is useful for anyone.

This is for mocking purposes (see tests/devices/fetch_network.js).

PR #1262, opened 2025-02-21
fix: Correctly detect Node environment

felixrieseberg:
Right now, v86 prefers `XMLHttpRequest` when available. I'm patching this in my Electron use case to prefer `fs` for performance reasons.

With this change, v86 prefers `XMLHttpRequest` when available _and_ `process.versions.node` is undefined. If desired, I'm also happy to write a simple `try/catch` function that returns a boolean if `require('fs')` works.

### DIFF
--- a/lib/jor1k.js
+++ b/lib/jor1k.js
@@ -54,7 +54,7 @@ message.Abort = function()
 // XXX: Should go through emulator interface
 var LoadBinaryResource;
 
-if(typeof XMLHttpRequest !== "undefined")
+if(typeof XMLHttpRequest !== "undefined" && !(typeof process !== "undefined" && process.versions && process.versions.node))
 {
     LoadBinaryResource = function(url, OnSuccess, OnError) {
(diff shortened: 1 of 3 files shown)


copy:
> With this change, v86 prefers XMLHttpRequest when available and process.versions.node is undefined. If desired, I'm also happy to write a simple try/catch function that returns a boolean if require('fs') works.

I think it would be nice if we could move all three occurrences into `src/lib.js` and write this only once, e.g.:

```js
let fs;
if(typeof require !== "undefined")
{
    try {
        fs = require("fs");
    }
    catch(e) {}
}

if(fs)
{
    v86util.determine_size = ... // using fs
    v86util.load_file = ...
}
else
{
    v86util.determine_size = ... // using xhr
    v86util.load_file = ...
}
```

copy:
I included this in https://github.com/copy/v86/commit/16d961e8623ca412914adbb7d088d1df0fc5eab1

Issue #1268, opened 2025-03-04
Listing bug.


oierjgioewr:
There's a weird bug that causes the category buttons to stack, for example, when you select Linux and 32 bit, instead of searching for Linux OSes that are 32 bit, it shows OSes that are 32 bit, and OSes that are Linux, which means it shows Linux and 32 Bit Windows

chschnell:
This is not a bug, it's a choice.

If you select two Families at https://copy.sh/v86/ you have selected two sets of OSes. One way to combine the two sets would be their union (which is what you see) and another their intersection (what you want).

For example, select `BSD` and `DOS`, they are small (3 and 10 OSes, respectively) and mutually exclusive. When you select both you'll get a list of 13 OSes.

oierjgioewr:
I feel like there should atleast have the architecture/storage stuff be the way I feel like should be right, because it would make sense for 32 Bit and windows to show 32 Bit windows and not 32 Bit OSes and Windows OSes which in turn shows both windows and all other 32 bit OSes

cyberplant:
I never used this feature, always opened directly on the OSs, but I do agree with @oierjgioewr that maybe makes more sense to be able to filter things (intersection) than to make the union. 

For example, you want to test 16-bit opensource systems, and when you select both options you will get a list with 16-bit and 32-bit apps, and opensource and not opensource, so it's not very useful. 

Do you see any use case for the current method?

chschnell:
Well, I only wanted to provide additional information.

However, the change is quite simple, to intersect the selected OS families function [`update_filters()`](https://github.com/copy/v86/blob/8561f1abdf4051b190f0b208077083d95b41bd73/src/browser/main.js#L1543-L1582) in `src/browser/main.js` needs to be changed to:

```javascript
function update_filters()
{
    const filter = {};
    for(const element of filter_elements)
    {
        filter[element.id.replace(/filter_/, "")] = element.checked;
    }

    const show_all = !Object.values(filter).includes(true);
    for(const os of os_info)
    {
        const show = show_all ||
            (!filter["graphical"] || os.graphical) &&
            (!filter["text"] || !os.graphical) &&
            (!filter["linux"] || os.family === "Linux") &&
            (!filter["bsd"] || os.family === "BSD") &&
            (!filter["windows"] || os.family === "Windows") &&
            (!filter["unix"] || os.family === "Unix") &&
            (!filter["dos"] || os.family === "DOS") &&
[...]

PR #1272, opened 2025-03-08
Fine-tune OS filter behaviour in web UI

chschnell:
Currently, the different guest OS filters are logically combined in function `update_filters()` by `OR`-ing them together.

This patch splits filters into the same categories that are already defined in the UI (`Family`, `UI`, `Medium`, ...). Filters within a category are logically combined using `OR`, and categories are then combined using `AND`.

Note that the web UI DOM is left unchanged by this patch.

Fixes issue #1268.

### DIFF
--- a/src/browser/main.js
+++ b/src/browser/main.js
@@ -1534,7 +1534,7 @@
             };
         });
 
-        const filter_categories = [
+        const known_filter = [
             [   // Family:
                 { id: "linux", condition: os => os.family === "Linux" },
                 { id: "bsd", condition: os => os.family === "BSD" },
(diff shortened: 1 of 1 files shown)


src/browser/main.js
```diff
@@ -1534,50 +1534,81 @@
...
+            ],
+        ];
+
+        for(const filter_category of filter_categories)
         {
-            element.onchange = update_filters;
+            for(const filter of filter_category)
+            {
+                const element = document.getElementById(`filter_${filter.label}`);
+                dbg_assert(element);
```
copy, line 1585: This breaks debug.html (this code should not assume that filters or the OS table are present on the page)

chschnell: Fixed it, sorry about that.

copy:
Cheers!

PR #1280, opened 2025-03-15
Small fixes

basicer:
- Use the response contructor in the fetch unit tests instead of faking it
- Allow setting size of multiple virtio-consoles independently


copy:
Thanks!

PR #1281, opened 2025-03-18
es6

basicer:


debug.html
```diff
@@ -4,59 +4,15 @@
...
+<script>
+    var DEBUG = true;
+</script>
 <script src="build/capstone-x86.min.js"></script>
-<script src="build/libwabt.js"></script>
+<script src="build/libwabt.cjs"></script>
 
+<script type="module" src="./src/browser/main.js"></script>
+
+</script>
```
SuperMaxusa, line 15: unneeded closing tag

lib/jor1k.js
```diff
@@ -92,7 +97,7 @@ else
     LoadBinaryResource = function(url, OnSuccess, OnError)
     {
         //console.log(url);
-        require("fs")["readFile"](url, function(err, data)
+        import("node:" + "fs").then(fs => fs["readFile"](url, function(err, data)
```
SuperMaxusa, line 100: Do we need to concatenate "node:" and module name for `import()` every time?

basicer: Yeah, otherwise closure compalins it cant find the module.  Making it an expression like this seems to suppress that.

copy:
Excellent stuff, thanks!

chschnell:
@basicer: Many thanks for this!

My v86 embedding was already module-based and switching over to `libv86.mjs` worked like a charme!

Issue #1300, opened 2025-04-06
PS/2 soft reset


chschnell:
When executing the REBOOT console command under FreeDOS 1.4 the PS/2 keyboard becomes unresponsive after rebooting.

You can reproduce it using the [XCom profile](https://copy.sh/v86/?profile=xcom) (which is FreeDOS-based) or with the attached HDA image file **[fdos14-64m.zip](https://github.com/user-attachments/files/19620816/fdos14-64m.zip)** which contains a minimal FreeDOS 1.4 installation (no audio or network). The [KolibriOS profile](https://copy.sh/v86/?profile=kolibrios) actually freezes after a soft reset when it initializes the PS/2 mouse.

In all cases, a PS/2 device is in trouble after a reboot was triggered from within the guest OS.

I believe the reason for this to be an invalid runtime state of the `PS2` object after [`reboot_internal()`](https://github.com/copy/v86/blob/ff56eed36ba8363d2b9bc72e453313e2d1b7ce42/src/cpu.js#L673) has been invoked.

So I added a method `PS2.reset()` which simply resets all scalar object members back to the values assigned in the [PS2 constructor](https://github.com/copy/v86/blob/ff56eed36ba8363d2b9bc72e453313e2d1b7ce42/src/ps2.js#L11), and then I added a call to `this.devices.ps2.reset()` in [`reboot_internal()`](https://github.com/copy/v86/blob/ff56eed36ba8363d2b9bc72e453313e2d1b7ce42/src/cpu.js#L673) along the same pattern in which the other devices `virtio_9p`, `virtio_console` and `virtio_net` are handled there.

This fixed it.

It seems that FreeDOS expects the PS/2 device to be reset after triggering a soft reset (which suggests that other emulators do so). Obviously though, most other guest OSes do not have that problem.

I don't think that this patch could have any negative effects on the other OSes, but before creating a PR I just want to make sure that I didn't miss anything here.

copy:
Sounds good, feel free to send a PR.

chschnell:
All right, it's checked in. I did not want to duplicate all those assignments so I moved them out of the constructor, hope that's ok.

This also greatly simplifies using the FreeDOS 1.4 installer which triggers a reboot during installation, and before it needed some trickery to reenter that 2nd phase of the installation process with the keybord intact.

PR #1301, opened 2025-04-06
Reset PS/2 runtime state when soft-resetting CPU

chschnell:
Added PS/2 to devices which are reset in CPU.reboot_internal().

- Moved assignment of all member variables from constructor to new method PS2.reset().
- Added call to PS2.reset() in PS2 constructor and in CPU.reboot_internal().

Fixes issue #1300.


copy:
Thanks!

PR #1302, opened 2025-04-08
fetch: HTTP redirect to `localhost` for guest

SuperMaxusa:
This PR adds redirect to the host's `localhost` from the guest for fetch networking. 
To use it, go to `<port>.external` in your guest OS (note that this also requires CORS headers on the server).

Example: Kolibri WebView with opened `1234.external` that redirects to `localhost:1234`

kolibri-fetch

Also added a test and benchmark with HTTP local server.

~~P. S. If you have an idea how to make an embed for it (see https://github.com/copy/v86/commit/e0e80779a8300523036533ea979a296058db4716), let me know.~~


chschnell:
Ah, you're introducing a special DNS name for HTTP access to v86's localhost. Nice!

I installed your branch and played around with it, a bit of feedback:

I think that the DNS name pattern:

    <port>.v86local.http

is specific enough to drop the explicit `net_device.local_http` flag to activate this new feature. In my opinion this could always be active (`http` is no valid TLD, so it's fine), the backend URL would then simply remain `fetch`.

I want to suggest an even simpler DNS name pattern:

    v86host[:<port>]

with <port> defaulting to 80. The name `v86host` is leaned from `localhost`.

Next to the need for a CORS-enabled HTTP server it may be worth noting that this supports only HTTP requests, no HTTPS, but here's a workaround: If you use the v86 host's IP address (in the URL that you enter in the guest) it will make a fetch-based HTTPS request to the v86 host (the HTTPS server must of course listen on that IP, not just on `localhost` for this to work).

SuperMaxusa:
Thanks for your interest!

> I think that the DNS name pattern:  `<port>.v86local.http`  is specific enough to drop the explicit `net_device.local_http` flag to activate this new feature.

Mostly yes. To be honest, I wanted to make this option toggleable for security reasons, i. e. avoiding situations where VM has access to your local HTTP ports "with no reasons", like a https://security.stackexchange.com/questions/232345/ebay-web-site-tries-to-connect-to-wss-localhostxxxxx-is-this-legit-or-they. On the other hand, we have CORS in browsers and (maybe) this security restrictions are not needed.

> I want to suggest an even simpler DNS name pattern: `v86host[:<port>]` with <port> defaulting to 80. The name v86host is leaned from localhost.

If I understand correctly, this will be omits the HTTP gateway, if port is not 80 in the current code:

https://github.com/copy/v86/blob/ed0ed07a18b6688507606f6b87bf506067ccd6a1/src/browser/fetch_network.js#L58-L64

To fix that, we will need to create a gateway with a different IP and add that IP to the DNS handler that a bit of difficult.


Feel free to correct me if I'm wrong.

SuperMaxusa:
> Or is there a risk that HTTP clients (in the guests) perhaps reject these domain names?

Hmm, technically TLDs with only numbers are not allowed (https://stackoverflow.com/a/53875771), because we can't have the first character as a non-alphabetic, so we can use something like `v86host.p<port>`, for example:  `v86host.p8080`.

According to [RFC 2606](https://www.rfc-editor.org/rfc/rfc2606.html#section-2), the `.localhost` TLD is reserved. Maybe we can use `<port>.localhost` (e.g. `8080.localhost`)? At least some servers, such as Caddy, uses this TLD and it works with browsers and other HTTP clients.

chschnell:
> Maybe we can use `<port>.localhost` (e.g. `8080.localhost`)?

That looks fine!

tests/benchmark/fetch-download.js
```diff
@@ -0,0 +1,74 @@
...
+    serial_text += chr;
+
+    if(!booted && serial_text.endsWith("~% "))
+    {
+        booted = true;
+
+        emulator.serial0_send(`udhcpc;curl -o /dev/null -w '<%{speed_download}>%{exitcode}' ${SERVER_PORT}.v86local.http/bench\n`);
+    }
+
+    if(serial_text.endsWith(">0"))
```
copy, line 58: This should detect non-zero exit codes and call process.exit

tests/benchmark/fetch-download.js
```diff
@@ -0,0 +1,74 @@
...
+
+    if(LOG_SERIAL) process.stdout.write(chr);
+
+    serial_text += chr;
+
+    if(!booted && serial_text.endsWith("~% "))
+    {
+        booted = true;
+
+        emulator.serial0_send(`udhcpc;curl -o /dev/null -w '<%{speed_download}>%{exitcode}' ${SERVER_PORT}.v86local.http/bench\n`);
```
copy, line 55: Maybe add `-s`?

tests/benchmark/fetch-download.js
```diff
@@ -0,0 +1,74 @@
...
+        {
+            console.error("Can't parse console log");
+            process.exit(1);
+        }
+
+        console.log();
+        console.log("Average download speed: %s kB/s", (speed / 1024).toFixed(2));
+        process.exit(0);
+    }
+});
```
copy, line 74: This bench currently fails for me:

```
Booting now, please stand by
mount: mounting host9p on /mnt failed: No such file or directory

Enable networking using 'udhcpc'
Files send via emulator appear in /mnt/
~% udhcpc;curl -o /dev/null -w '<%{speed_download}>%{exitcode}' 8686.v86local.ht
tp/bench
udhcpc: started, v1.36.1
udhcpc: broadcasting discover
udhcpc: broadcasting select for 192.168.86.100, server 192.168.86.1
udhcpc: lease of 192.168.86.100 obtained from 192.168.86.1, lease time 134217728
deleting routers
adding dns 192.168.86.1
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0Fetch Failed: http://8686.v86local.http/bench
TypeError: fetch failed
100   202  100   202    0     0   2595      0 --:--:-- --:--:-- --:--:--  2805
<2595>0
Average download speed: 2.53 kB/s
```

(could you also have a quick look why curl returns 0 even though the header should be 502?)

tests/benchmark/fetch-download.js
```diff
@@ -0,0 +1,74 @@
...
+import { Worker } from "node:worker_threads";
+
+const __dirname = url.fileURLToPath(new URL(".", import.meta.url));
+
+const TEST_RELEASE_BUILD = +process.env.TEST_RELEASE_BUILD;
+const USE_VIRTIO = !!process.env.USE_VIRTIO;
+const SERVER_PORT = parseInt(process.env.SERVER_PORT, 10) || 8686;
+const BENCHFILE_SIZE = (parseInt(process.env.BENCHFILE_SIZE_MB, 10) || 32) * 1024 * 1024;
+
+const { V86 } = await import(`../../build/${TEST_RELEASE_BUILD ? "libv86" : "libv86-debug"}.mjs`);
```
copy, line 14:
```suggestion
const { V86 } = await import(`../../build/libv86.mjs`);
```

(I think debug mode should be off for benchmarks to reduce the noise)

src/browser/fetch_network.js
```diff
@@ -113,13 +115,41 @@ async function on_data_http(data)
...
+            } else {
+                console.warn('Unknown port for localhost: "%s"', target.href);
+                const body = new TextEncoder().encode(`Unknown port for localhost: ${target.href}`);
+                const resp_headers = new Headers({
+                    "content-type": "text/plain",
+                    "content-length": body.length.toString(10),
+                    "connection": "close"
+                });
+                this.writev([new TextEncoder().encode(this.net.form_response_head(400, "Bad Request", resp_headers)), body]);
+                this.close();
```
copy, line 148: Could be moved into a function (also used in some other places):


```javascript
                const body = new TextEncoder().encode(`Unknown port for localhost: ${target.href}`);
                const resp_headers = new Headers({
                    "content-type": "text/plain",
                    "content-length": body.length.toString(10),
                    "connection": "close"
                });
                this.respond_and_close(400, "Bad request", resp_headers, body);
```

SuperMaxusa: I would make a method to respond with a message/error:
```js
this.respond_message_and_close(400, "Bad request", `Unknown port for localhost: ${target.href}`);
```

tests/benchmark/fetch-download.js
SuperMaxusa: > (could you also have a quick look why curl returns 0 even though the header should be 502?)

I added [`--fail`](https://curl.se/docs/manpage.html#-f), so curl should react on "Fetch failed".

> This bench currently fails for me

Hmm, have you tried using a different HTTP port? Most likely, some app uses this port (`8686`) and `fetch()` cannot connect to the test server.

copy:
> Gently pinging @copy: which option looks better to you: <port>.localhost or <port>.v86local.http?

How about `<port>.v86.internal`?

It would also make sense to have a "magic" ip address (like qemu's 10.0.2.2). That might be useful with wisp networking or with user mode networking in the future.

SuperMaxusa:
> How about `<port>.v86.internal`?

Sounds good

> It would also make sense to have a "magic" ip address (like qemu's 10.0.2.2). That might be useful with wisp networking or with user mode networking in the future.

I agree, but not sure about wisp.
If I correctly remember, all TCP connections go through relay, so we need an HTTP handler from `fetch` and make the same thing that I wrote [above](https://github.com/copy/v86/pull/1302#issuecomment-2787324352).

copy:
> If I correctly remember, all TCP connections go through relay, so we need an HTTP handler from fetch and make the same thing that I wrote https://github.com/copy/v86/pull/1302#issuecomment-2787324352.

At the moment, yes. But we already have tcp/ip packet parsing code, so we could redirect those connections. But that's not really related to this PR.

tests/benchmark/fetch-download.js
SuperMaxusa: Done, I fixed the server, now it uses a random available port (or statically by `SERVER_PORT=1234`) for test and benchmark.

tests/devices/fetch_network.js
```diff
@@ -2,17 +2,20 @@
...
 
 const USE_VIRTIO = !!process.env.USE_VIRTIO;
-
 const TEST_RELEASE_BUILD = +process.env.TEST_RELEASE_BUILD;
-const { V86 } = await import(TEST_RELEASE_BUILD ? "../../build/libv86.mjs" : "../../src/main.js");
 
+const { V86 } = await import(TEST_RELEASE_BUILD ? "../../build/libv86.mjs" : "../../src/main.js");
 const SHOW_LOGS = false;
 
+var SERVER_PORT = parseInt(process.env.SERVER_PORT, 10) || 0;
```
copy, line 16: Is there still a need to specify this using an env variable? If not, I'd suggest removing it and simplifying the remaining code (that is, the worker always sends a random port to the caller).

tests/benchmark/fetch-download.js
```diff
@@ -0,0 +1,93 @@
...
+    net_device: {
+        relay_url: "fetch",
+        type: USE_VIRTIO ? "virtio" : "ne2k",
+        local_http: true
+    }
+});
+
+emulator.bus.register("emulator-ready", function()
+{
+    server = new Worker(__dirname + "../devices/fetch_testserver.js", { workerData: { port: SERVER_PORT, benchsize: BENCHFILE_SIZE } });
```
copy, line 34: I think this code would be overall more readable if you inline the worker code in here (by creating a function and converting to string, if possible: https://github.com/copy/v86/blob/a3840e32b30d7005f2e7533a557b769333127425/src/main.js#L135)

Then simplify the worker (inline get_handler, only handle `/bench`).

Same for the test case.

tests/devices/fetch_network.js
SuperMaxusa: It's for debugging. When I wrote that server worker, I wanted to make similar test for browser (like a [manual/gc.html](https://github.com/copy/v86/blob/master/tests/manual/gc.html)) but abandoned this idea later.

tests/benchmark/fetch-download.js
SuperMaxusa: Worker threads only supports script files, so converting code to string is not suitable here.

```
> the_worker = () => console.log("hello")

> url = URL.createObjectURL(new Blob(["(" + the_worker.toString() + ")()"], { type: "text/javascript" }))

> const { Worker } = await import("node:worker_threads")

> new Worker(url)
Uncaught:
TypeError [ERR_WORKER_PATH]: The worker script or module filename must be an absolute path or a relative path starting with './' or '../'. Received "blob:nodedata:<...>"
    at new Worker (node:internal/worker:187:15) {
  code: 'ERR_WORKER_PATH'
}
```

SuperMaxusa:
Revisited some ideas:
 - `<port>.v86local.http` was replaced by `<port>.external` (`external` sounds more logical from VM perspective than `internal`)
 - thinking about removing the `local_http` flag: the risks of "unethical access to localhost" are quite small, and it also requires an extra switch in the UI

docs/networking.md
```diff
@@ -62,6 +62,7 @@ Backends `fetch` and `wisp` support a couple of special settings in `config.net_
 | **dns_method** | str  | DNS method to use, either `static` or `doh`. `static`: use built-in DNS server, `doh`: use [DNS-over-HTTPS](https://en.wikipedia.org/wiki/DNS_over_HTTPS) (DoH). Defaults to `static` for `fetch` and to `doh` for `wisp` backend. |
 | **doh_server** | str  | Host name or IP address (and optional port number) of the DoH server if `dns_method` is `doh`. The value is expanded to the URL `https://DOH_SERVER/dns-query`. Default: `cloudflare-dns.com`. |
 | **cors_proxy** | str  | CORS proxy server URL, do not use a proxy if undefined. Default: undefined (`fetch` backend only). |
+| **local_http** | bool | Allow HTTP access from the guest to the host's `localhost` using the URL `http://<port>.v86local.http` (e.g. `1234.external` -> `localhost:1234`). Default: `False` (`fetch` backend only). |
```
copy, line 65: This still mentions v86local.http

tests/devices/fetch_network.js
```diff
@@ -202,19 +250,61 @@ const tests =
...
+    server.on("message", function(msg) {
+        SERVER_PORT = msg;
+        console.log("Server started on port " + SERVER_PORT);
+    });
+}
+else
+{
+    // placeholder for main thread handlers
+    emulator = { bus: {} };
+    emulator.add_listener = emulator.bus.register = () => null;
```
copy, line 281: Wouldn't it be simpler to move the calls to emulator.add_listener etc. into the if branch above and delete this code?

copy:
> * `<port>.v86local.http` was replaced by `<port>.external` (`external` sounds more logical from VM perspective than `internal`)
> 
> * thinking about removing the `local_http` flag: the risks of "unethical access to localhost" are quite small, and it also requires an extra switch in the UI

Sounds good. The code looks good now (two minor remarks above), please unmark as draft so I can merge later.

copy:
Merci!

Issue #133, opened 2017-02-26
Any plans to support 64 bit?


DaniellPlasmmer:

copy:
It would be nice to have, but it's a huge project and I won't find enough time any time soon to work on it.

I'll leave this open, if someone wants to work on this I can help you get started.

DaniellPlasmmer:
Before implement it, is better make v86 to support Windows XP.

BenNottelling:
I'd like to try this out, but I know almost nothing about how v86 works :)

czenzel:
@copy I read you are experimenting with Web Assembly on issue #170? Do you have a sneak peek we can take a look at? I am ready to experiment!

copy:
> @copy I read you are experimenting with Web Assembly on issue #170? Do you have a sneak peek we can take a look at? I am ready to experiment!

Not yet.

Darker:
What are the things that would need to be changed to make this work?

Broyojo:
Lots of things, the processor is only an i386 processor.

Broyojo:
It would be cool if we could change it to an amd64 cpu

Broyojo:
You would have to change the architexture and the clock speed to run even
some 64 bit operatin systems

Broyojo:
architecture

sr229:
Reminder v86 emulates only x86 instruction sets, I don't see why would we emulate x86_64 since its more modern and techincally you can just grab a whole PC today

yuri2peter:
WE NEED X64 !!!

tlrobinson:
This would be awesome, if only to get [TempleOS](https://en.wikipedia.org/wiki/TempleOS) running.

samdenty:
https://github.com/ktock/container2wasm

androidtest3817:
.

androidtest3817:
> _

androidtest3817:
> WE NEED X64 !!!

How you minimized it

tennox:
@androidtest3817 Please stop spamming unnecessary & unrelated comments. 
Many people are following this thread to stay up to date. 🙏

agilgur5:
[Box64](https://github.com/ptitSeb/box64) might have some useful code as an example, but it's implemented quite differently and with different feature sets than v86 as far as I can tell.

Also `copy`, the author, wrote up some tips on implementing this in https://github.com/copy/v86/issues/648#issuecomment-1119820520

Acrotap:
[...]

PR #1303, opened 2025-04-09
Add SculptOS

xelxebar:


copy:
Does SculptOS work in v86?

SuperMaxusa:
Can you clarify which version you tested?

> Does SculptOS work in v86?

From my tests, all versions since 2018 (https://genode.org/download/sculpt-archive) are built for x86_64. 
[Genode OS Framework](https://www.qemu-advent-calendar.org/2016/download/day11.tar.xz) from QEMU Advent Calendar 2016 also doesn't work (maybe incompatibility with non-q35 chipsets).

xelxebar:
In my original test, I thought it booted, but I cannot reproduce now. I must have gotten some wires crossed.

Upstream's codebase does have explicit references to x86 support, but I can't seem to find a public build.

Apologies for the mixup!

copy:
Let's close this then. Probably needs https://github.com/copy/v86/issues/133

PR #1307, opened 2025-04-13
VGA BIOS input

SuperMaxusa:


src/browser/main.js
```diff
@@ -1888,7 +1894,7 @@ import { log_data } from "../log.js";
                 settings.initrd = { buffer: initrd };
             }
 
-            const title = multiboot?.name || hda?.name || cdrom?.name || hdb?.name || fda?.name || bios?.name;
+            const title = multiboot?.name || hda?.name || cdrom?.name || hdb?.name || fda?.name || bios?.name || vga_bios?.name;
```
copy, line 1897: This code selects the first boot medium as the title. Since the vgabios is not bootable on its own, I don't think it should be here.

src/browser/main.js
```diff
@@ -10,6 +10,7 @@ import { log_data } from "../log.js";
 
 
     const ON_LOCALHOST = !location.hostname.endsWith("copy.sh");
+    const BIOSPATH = "bios/";
```
copy, line 13: I prefer having this constant closer to the code where its used.

copy:
Thanks!

PR #1340, opened 2025-06-03
Fix serial not handling end of line properly

ConfusedPenguin:
Fixed a weird issue with the serial terminal not handling end of line characters correctly. For example, on ToaruOS, the serial output is displayed sporadically and the stream does not proceed to the next line. See image:

toaruOSbefore

Adding the convertEol option to the xterm initialization results in this output: 

toaruOSafter

Based on my limited testing on FreeDOS, ToaruOS with serial debug on and my own simple kernel, this does not affect the other functionalities of the serial monitor, but it's possible.
Let me know if it can be done better or if it affects any other functionality.
Cheers.

### DIFF
--- a/src/browser/serial.js
+++ b/src/browser/serial.js
@@ -230,6 +230,7 @@ export function SerialAdapterXtermJS(element, bus)
 
     var term = this.term = new window["Terminal"]({
         "logLevel": "off",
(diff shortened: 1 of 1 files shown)


copy:
Thanks!

PR #1345, opened 2025-06-07
Improve IDE Controller (rebased + fix for old state images)

copy:
This is the same as #1325, with three extra changes:

- eb9f20dd7dcd30ca7f83d06c491d3b47efae61e4     make ide changes compatible with old state images 
- 57feb57 minor cleanup
- rebased to remove merge commit

@chschnell Could you do a quick review of this? An alternative would have been to layout the ide controller the same as previously (CD-ROM primary if there is no hard disk), but I slightly prefer having the CD-ROM always as the secondary.

### DIFF
--- a/debug.html
+++ b/debug.html
@@ -212,6 +212,7 @@
         <input type="button" value="Get second hard disk image" id="get_hdb_image">
         <input type="button" value="Get CD-ROM image" id="get_cdrom_image">
         <input type="button" value="Insert floppy image" id="change_fda_image">
+        <input type="button" value="Insert CD image" id="change_cdrom_image">
         <input type="button" value="Save State" id="save_state">
         <input type="button" value="Load State" id="load_state"> <input type="file" style="display: none" id="load_state_input">
         <input type="button" value="Memory Dump" id="memory_dump">
--- a/index.html
+++ b/index.html
@@ -241,6 +241,7 @@
         <input type="button" value="Get second hard disk image" id="get_hdb_image">
         <input type="button" value="Get CD-ROM image" id="get_cdrom_image">
         <input type="button" value="Insert floppy image" id="change_fda_image">
+        <input type="button" value="Insert CD image" id="change_cdrom_image">
         <input type="button" value="Save State" id="save_state">
         <input type="button" value="Load State" id="load_state"> <input type="file" style="display: none" id="load_state_input">
         <input type="button" value="Memory Dump" id="memory_dump">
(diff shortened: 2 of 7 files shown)


chschnell:
> @chschnell Could you do a quick review of this? An alternative would have been to layout the ide controller the same as previously (CD-ROM primary if there is no hard disk), but I slightly prefer having the CD-ROM always as the secondary.

Much agree to everything in the second commit (cleanup).

But I don't fully understand the first. [This condition in cpu.js](https://github.com/copy/v86/blob/d9088888e477663f0bf627a3e262d00af2849109/src/cpu.js#L541) in your commit can never be true, see [further down in cpu.js](https://github.com/copy/v86/blob/d9088888e477663f0bf627a3e262d00af2849109/src/cpu.js#L1146-L1154) where we create the IDEController. In line 1152, we statically assign the cdrom device to `this.devices.ide.secondary` and use it then in 1154. Only `this.devices.ide.primary` can be undefined in case there is no `hda` image.

EDIT: I missed your change even further down below in `set_state()`. I don't know enough about the set/get_state() meachanics and what cases from the past must be handled in get_state(), so if it works for you then it is correct. The new state field for the entire IDEController settles it for the future.

copy:
> But I don't fully understand the first. [This condition in cpu.js](https://github.com/copy/v86/blob/d9088888e477663f0bf627a3e262d00af2849109/src/cpu.js#L541) in your commit can never be true, see [further down in cpu.js](https://github.com/copy/v86/blob/d9088888e477663f0bf627a3e262d00af2849109/src/cpu.js#L1146-L1154) where we create the IDEController. In line 1152, we statically assign the cdrom device to `this.devices.ide.secondary` and use it then in 1154. Only `this.devices.ide.primary` can be undefined in case there is no `hda` image.
> 
> EDIT: I missed your change even further down below in `set_state()`. I don't know enough about the set/get_state() meachanics and what cases from the past must be handled in get_state(), so if it works for you then it is correct. The new state field for the entire IDEController settles it for the future.

Indeed, this code is to handle the `set_state(old_state_image) -> get_state` case.

I will write some more testcases in the future, but my manual testing and the other tests we have so far have all been successful.

Merged, thanks a lot!

chschnell:
> I will write some more testcases in the future, but my manual testing and the other tests we have so far have all been successful.
> 
> Merged, thanks a lot!

Awesome, and you're welcome!

Issue #1349, opened 2025-06-13
Async CDROM functionality broken when booting from HDD


ericmackrodt:
- Version of v86: latest-22-g0eec2965
- Browser/OS: Chrome and Zen (Firefox) on MacOS 15.3.2 (M1 Mac)

Hello,

First I want to congratulate @chschnell, @JoeOsborn and @copy for the incredible work getting the PR that allows booting with both the CD and HDD drives. It's an amazing advancement for the emulator.

The emulator does boot from the HDD with the CD in the drive and the CD image is accessible from within the OS, if the cd is not set to async.

Here's my current config that is doing that, the hard drive image fresh from qemu:

```
 const initOpts: V86Options = {
      wasm_path: "/v86.wasm",
      memory_size: 64 * 1024 * 1024,
      vga_memory_size: 32 * 1024 * 1024,
      screen_container: emulatorContainer,
      preserve_mac_from_state_image: true,
      hda: { url: "/hdd.img", async: true },
      cdrom: { url: "/98install.iso" }, // cd is not async (also, same image used on qemu to install Windows 98 SE)
      bios: {
        url: "/bios/seabios.bin",
      },
      vga_bios: {
        url: "/bios/vgabios.bin",
      },
      autostart: true,
    };
```

Keep in mind that, when I boot the machine fresh from qemu, it re-detects the hardware/devices and reboots before I can go to My Computer to check the cd rom state.
But that works perfectly.

Unfortunately I found a couple of issues when it comes to booting from the HDD while having a cd image configured as async.

## Scenario 1: Booting the machine with async cdrom image.

If I do the exact same as above, but with the cdrom set to `async: true`:

```
 const initOpts: V86Options = {
      wasm_path: "/v86.wasm",
      memory_size: 64 * 1024 * 1024,
      vga_memory_size: 32 * 1024 * 1024,
      screen_container: emulatorContainer,
      preserve_mac_from_state_image: true,
      hda: { url: "/hdd.img", async: true },
      cdrom: { url: "/98install.iso", async: true }, // cd is async
      bios: {
        url: "/bios/seabios.bin",
      },
      vga_bios: {
        url: "/bios/vgabios.bin",
      },
      autostart: true,
    };
```

The emulator starts and, in the networking tab from the browser's dev tools, I can see the chunks of the ISO are being downloaded, but after the drivers are re-detected and the machine rebooted, I see the label of the cd drive is just `Cd` and if I try to browse it's files, it's blank:

[...]

PR #1351, opened 2025-06-16
Fix async CD-ROM with DMA (experimental)

chschnell:
See issue #1349.

### DIFF
--- a/src/ide.js
+++ b/src/ide.js
@@ -1431,6 +1431,11 @@ IDEInterface.prototype.atapi_handle = function()
             if(this.features_reg & 1)
             {
                 this.atapi_read_dma(this.data);
+                if(!(this.status_reg & ATA_SR_DRQ))
+                {
+                    // ATAPI READ response is delayed, do not raise IRQ (yet)
+                    return;
+                }
             }
             else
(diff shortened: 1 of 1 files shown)


chschnell:
Unfortunately, I get occasional errors under Debian with this. I didn't capture the errors, but I think I know what's going on here.

I think the ATA code in ide.js is fine, the ATAPI code less so, but the interface between ATA and ATAPI is a real mess. The problem I faced there is that I was not able to spot proper documentation about this interface in any of the specifications I went through. I understand how to do it correctly, but when I tried to implement it (using qemu as a reference) it would work with the newer guest OSes but broke the older DOS guests. So there's something I missed there.

Now, async ATAPI hits right into that ATA/ATAPI interface problem, it's already bad in the synchronous case, but the delays that the async I/O adds here makes all this even worse.

I'll get back to this, but I can't say when (this is not trivial). Until then, maybe we should close this PR and discourage use of async ATAPI, it's not stable.

copy:
> Until then, maybe we should close this PR and discourage use of async ATAPI, it's not stable.

Okay, let's do that.

PR #1354, opened 2025-06-19
Typescript definitions

copy:


chschnell:
A small detail regarding mouse capture: Using the [`unadjustedmovement`](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestPointerLock#unadjustedmovement) option of [requestPointerLock()](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestPointerLock) causes raw mouse pointer data to be captured (as opposed to accelerated mouse data). Synthetic mouse acceleration should be left to the guest OS and this might avoid doing it twice.

copy:
@chschnell Thanks, added

PR #1361, opened 2025-07-03
Add TLS support to fetch based networking.

basicer:


src/browser/fetch_network.js
```diff
@@ -36,6 +35,14 @@ export function FetchNetworkAdapter(bus, config)
     this.eth_encoder_buf = create_eth_encoder_buf();
     this.fetch = (...args) => fetch(...args);
 
+    eval(`import("../../build/mitm.mjs")`).then(factory => factory.default()).then(obj => {
+        this.tls = new obj["MITM"]();
+        this.tls["generateECCPrivateKey"]();
+    }).catch(e => {
+        console.log(e);
+        dbg_log("No TLS library detected.");
```
SuperMaxusa, line 43: Missing `LOG_FETCH`

copy:
I don't think I want to take on the large mitm.mjs dependency in the default build of v86. I'd accept the following:

- comment out the test (remove mitm.mjs from the makefile)
- fetch_network.js searches for window.TLS or similar and uses that if present
- document how a user of v86's api could enable tls support for fetch-based networking by importing mitm.js (and adding it to window if mitm.js doesn't already do that)
- alternatively pass tls as an object to V86 (instead of going through window)
- add documentation on how to use this

copy:
Feel free to reopen if you'd like to revisit this idea.

basicer:
I'm still cooking on this, stay tuned!

PR #1372, opened 2025-07-22
[PWA] add pwa support but no local storage yet

trcrsired:
Add PWA support but not local storage support (so the vm state won't be saved) but at least it is now installable on the phone


copy:
Thanks, I wrote a manifest.json some time ago (https://github.com/copy/v86/commit/8e4a29eee60576602df6a803f19c2d2c1e8e58cb), so let's use that instead.

PR #1386, opened 2025-08-17
add support for custom 9p request handler

### LINKED ISSUE #1385: Generalize 9p support
The virtio 9P powered `filesystem` has been great, but I'd like to submit a PR to generalize it at the 9P level. In [Wanix][1] we [patch Virtio9p][2] to hand off the 9P request and reply buffers to Wanix, which just pipes them into an actual 9P server connection. Obviously, I'd like to avoid maintaining a downstream patch, but also after putting together [v86-system][3], it would be nice for it to also have the option to point to a 9P file server for root without needing networking.

I'm willing to put this together since I've already done it, but I'm curious the best way to configure it. Originally I was thinking maybe we could specify the the class to use and default to the current Virtio9p. But maybe it would be better to create a hook in the same place as the Wanix patch and let you specify a request handler for Virtio9p since that's all that matters and you'd always want the Virtio part that's already in Virtio9p. 

[1]: https://github.com/tractordev/wanix
[2]: https://github.com/tractordev/wanix/blob/main/external/v86/patch/9p.js#L278
[3]: https://github.com/progrium/v86-system

progrium:
This is an initial pass at allowing a custom 9P handler function to be provided that would be used by Virtio9p instead of the HTTP backed filesystem. It's based on a working patch used by [Wanix][1]. It compiles but hasn't been thoroughly tested yet. I'd like feedback on implementation/API and how much testing would be appropriate. I can provide an example/test that's based on a Wasm 9P file server written in Go if that would be good. 

It also includes a pre-made WebSocket proxy adapter for using non-browser 9P servers.

This proposes two new fields on V86Option's `filesystem` object:
```typescript
export interface V86Options {
    // ...
    
    filesystem?: {
        // ...
        
        /**
         * A function that will be called for each 9p request.
         * If specified, this will back Virtio9p instead of a filesystem.
         * Use this to connect Virtio9p to a custom 9p server.
         */
        handle9p?: (reqbuf: Uint8Array, reply: (replybuf: Uint8Array) => void) => void;

        /**
         * A URL to a websocket proxy for 9p.
         * If specified, this will back Virtio9p instead of a filesystem.
         * Use this to connect to a custom 9p server over websocket.
         */
        proxy_url?: string;
    };
}
```

This would close #1385 

[1]: https://github.com/tractordev/wanix/blob/main/external/v86/patch/9p.js#L278

### DIFF
--- a/docs/filesystem.md
+++ b/docs/filesystem.md
@@ -1,9 +1,9 @@
 A 9p filesystem is supported by v86, using a virtio transport. There are several
-ways it can be set up. 
+ways it can be set up.
 
 ### Guest mount
 
-In all cases, the filesystem is mounted in the guest system using the `9p` 
+In all cases, the filesystem is mounted in the guest system using the `9p`
 filesystem type and the `host9p` device tag. Typically you want to be specific
 with the version and transport options:
 
@@ -17,7 +17,7 @@ Here are kernel arguments you can use to boot directly off the 9p filesystem:
 rw root=host9p rootfstype=9p rootflags=trans=virtio,version=9p2000.L
 ```
 
-The `aname` option can be used to pick the directory from 9p to mount. The `rw` 
+The `aname` option can be used to pick the directory from 9p to mount. The `rw`
 argument makes this a read-write root filesystem.
 
 
@@ -25,7 +25,7 @@ argument makes this a read-write root filesystem.
 
 This is the standard way to setup the 9p filesystem. It loads files over
 HTTP on-demand into an in-memory filesystem in JS. This allows files to be
-exchanged with the guest OS. See `create_file` and `read_file` in 
+exchanged with the guest OS. See `create_file` and `read_file` in
 [`starter.js`](https://github.com/copy/v86/blob/master/src/browser/starter.js).
 
 This mode is enabled by passing the following options to `V86`:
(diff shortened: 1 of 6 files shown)


progrium:
I split it out into its own `Virtio9pHandler` class as recommended. However, I also added a `Virtio9pProxy` class for directly connecting to a 9P server over WebSocket based on the `NetworkAdapter` code. This comes with a new `filesystem.proxy_url` configuration field. Let me know if naming is ok, I was using convention of `relay_url` but might be weird with `baseurl` in the same namespace.

I actually used this build in the 0.4 release of [v86-system](https://github.com/copy/v86/discussions/1384) to allow a `virtfs` flag (again similar in design to qemu), letting you boot a v86 Linux system with just a kernel and a rootfs directory if running a 9P WebSocket server:
```bash
v86-system -kernel vmlinuz -virtfs proxy,ws://localhost:7654 -append "console=ttyS0 rw root=host9p rootfstype=9p rootflags=trans=virtio,version=9p2000.L,aname=rootfs"
```

You may be wondering, why not just have the proxy. Well, Wanix still implements a 9P server in-browser, so for anything like that you still want the handler hook. Besides allowing the functionality in `v86-system`, having a pre-made WS proxy adapter as well seems like it would be more accessible and useful to a wider audience.

Because there are now 3 implementations, I also refactored the core Virtio setup code into an unexported class they all use. Hopefully this is acceptable.

SuperMaxusa:
> having a pre-made WS proxy adapter seems like it would be more accessible and useful to a wider audience.

By the way, do you have a guide on how to set up a server for `filesystem.proxy_url`?
I think it would be great to have this in [`docs/filesystem.md`](https://github.com/copy/v86/blob/master/docs/filesystem.md).

progrium:
Great suggestion. I'll push docs on it next.

copy:
Thanks!

PR #1396, opened 2025-08-31
ide: ATAPI command `READ(12)`

SuperMaxusa:
required for Mojo OS

Not sure if this command is deprecated in final ATAPI revisions because I only found mention of it in drafts. Also found in the [Bochs](https://github.com/bochs-emu/Bochs/commit/afc06f029686ecacd9de3decce5a4f5c5507cfce) and [MINIX 3](https://github.com/Stichting-MINIX-Research-Foundation/minix/blob/4db99f4012570a577414fe2a43697b2f239b699e/minix/drivers/storage/ahci/ahci.h#L131) source codes.

### DIFF
--- a/src/ide.js
+++ b/src/ide.js
@@ -35,6 +35,9 @@ import { BusConnector } from "./bus.js";
 // - [BMI-1]
 //   Programming Interface for Bus Master IDE Controller, Revision 1.0, 5/16/94
 //   https://storage.googleapis.com/google-code-archive-downloads/v2/code.google.com/u9proj/idems100.pdf
+// - [SFF-8020]
+//   ATA Packet Interface for CD-ROMs (Rev. 1.2, Feb. 12, 1994)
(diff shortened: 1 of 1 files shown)


src/ide.js
```diff
@@ -1640,7 +1646,7 @@ IDEInterface.prototype.atapi_read = function(cmd)
 {
     // Note: Big Endian
     var lba = cmd[2] << 24 | cmd[3] << 16 | cmd[4] << 8 | cmd[5];
-    var count = cmd[7] << 8 | cmd[8];
+    var count = cmd[0] === ATAPI_CMD_READ_12 ? (cmd[6] << 24 | cmd[7] << 16 | cmd[8] << 8 | cmd[9]) : (cmd[7] << 8 | cmd[8]);
```
copy, line 1649: To be safe, use `count >>>= 0` to cast to an unsigned int (also below)

copy:
Thanks!

PR #1423, opened 2025-09-27
add loopback to UART

russor:
A lot of OSes use uart loopback to detect or confirm a uart is present and that it is working.


copy:
Thanks!

PR #1426, opened 2025-10-04
fix iso file extension detection

SuperMaxusa:
See https://mimetype.io/application/x-iso9660-image and https://en.wikipedia.org/wiki/ISO_9660#Disc_images


copy:
Thanks!

PR #1428, opened 2025-10-06
feat: Add WebSocket serial adapter and UI integration

sysoleg:
This commit introduces a new `SerialAdapterWebSocket` that allows forwarding of serial port communication over a WebSocket connection.

### DIFF
--- a/src/browser/main.js
+++ b/src/browser/main.js
@@ -2111,7 +2111,7 @@ function start_emulation(profile, query_args)
     }
     settings.disable_audio = $("disable_audio").checked || settings.disable_audio;
     if(settings.disable_audio) new_query_args.set("mute", "1");
-    settings.serial_websocket_url = $("serial_websocket_url").value;
+    settings.serial_websocket_url = $("serial_websocket_url").value || settings.serial_websocket_url;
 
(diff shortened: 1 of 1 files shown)


src/browser/starter.js
```diff
@@ -17,6 +17,7 @@ import { ScreenAdapter } from "./screen.js";
 import { DummyScreenAdapter } from "./dummy_screen.js";
 import { SerialAdapter, SerialAdapterXtermJS } from "./serial.js";
 import { InBrowserNetworkAdapter } from "./inbrowser_network.js";
+import { SerialAdapterWebSocket } from "./serial_websocket.js";
```
SuperMaxusa, line 20: Add this file to Makefile:
https://github.com/copy/v86/blob/9735a0eed83a426b3b14cf98f9043a5a9241b142/Makefile#L88-L91
...or merge with [`src/browser/serial.js`](https://github.com/copy/v86/blob/master/src/browser/serial.js)

copy:
What is this used for?

I don't mind having this in the code, but unless there's particularly interesting public endpoints, I don't think this should occupy space in the UI.

copy:
Feel free to reopen if you can explain the use-case better.

PR #1430, opened 2025-10-07
docs: update `windows-9x.md`

SuperMaxusa:
Added instructions for fixing `VFBACKUP Could Not Load VFD.VXD` and installing VMDisp9x.


copy:
Thanks!

Issue #1427, opened 2025-10-06
1 MB hard drive (both a and b)


johntim-winlinuxfan:
Version: ccb5e1ce (Sep 20, 2025 11:09)
Browser/OS Chrome Version 126.0.6478.222 (Official Build) (64-bit), ChromeOS

i have no words whats happening so basically i tried doing a 200 megabyte hard drive (note : this bug happened on this update before it was fine) and it keeps saying its 1 megabyte now it had only 1 directory and it was dos 7.10

<img width="602" height="232" alt="Image" src="https://github.com/user-attachments/assets/8b04a476-ca2f-4aa6-bafe-b873215c6ed6" />

johntim-winlinuxfan:
(note 2 : poweriso was taking ages so i just made an smaller partition also i need more than 220 mb so i made an img file with 256mb only because storage problem so thats why i use the built in empty disk (BEFORE the bug))

FunnyCorgi:
I'm having the same problem: setting any size for any drive makes any guest read it as 1mb and the image export also reads at 1mb.
I think the culprit is how the hard disk size is set: if you look in the url the byte size of the disk is incorrect:
100 MB is 104857600 bytes
V86 writes into the url 1048576 bytes(1mb) which is less than the target no matter what you put in(unless you put in less than one megabyte).
FIRST EDIT:
I think something in the following where it sets the size doesn't work correctly:
```javascript
const hda_empty_size = +$("hda_empty_size")?.value;
        if(hda_empty_size)
        {
            const size = Math.min(1, Math.max(MAX_ARRAY_BUFFER_SIZE_MB, hda_empty_size)) * 1024 * 1024;
            settings.hda = { buffer: new ArrayBuffer(size) };
            new_query_args.set("hda.empty", String(size));
        }
```
Though this code seems like it should work just fine. Since it is able to correctly access the size value as far as I can tell, the only reason I would see this failing is if:
```javascript
Math.max(MAX_ARRAY_BUFFER_SIZE_MB, hda_empty_size)
```
returned a value greater than one(which it will for user provided sizes) and then `math.min` would take the value of the 1mb minimum then multiply to get the result size in bytes.


**EDIT**: I realized that it's a little hard to understand what I wrote here: basically the math.min and math.max appear to be evaluated in the wrong order and the evaluation currently goes as follows:
[...]

PR #1433, opened 2025-10-12
Fix HDA and HDB always being set to 1 MB and fix HDB to buffer to be created to the proper value

FunnyCorgi:
This is towards #1427. This PR fixes an issue where the math performed on the target size of either HDA or HDB would always return 1 MB in size.

### DIFF
--- a/src/browser/main.js
+++ b/src/browser/main.js
@@ -2170,7 +2170,7 @@ function start_emulation(profile, query_args)
         if(hdb_empty_size)
         {
             const size = Math.max(1, Math.min(MAX_ARRAY_BUFFER_SIZE_MB, hdb_empty_size)) * 1024 * 1024;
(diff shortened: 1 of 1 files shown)


FunnyCorgi:
After testing the docker build in a Codespace I can confirm that the webpage properly sets `hda.empty` in the query args and exporting the image from the built webpage with the change does indeed download a 100 MB image when and sets it to the proper bytes value in the url. I didn't test `hdb.empty` but I'd assume it would run just the same. If the need arises I can but the logic for both was the same before and is the same now so there shouldn't be any issues.

FunnyCorgi:
Just noticed an issue with HDB using the raw MB value that the user provides instead of the clamped value in bytes for that same MB value when creating the buffer. ~~I'll go ahead and fix that.~~ EDIT: This has been fixed in the following commit.

copy:
Thanks!

Issue #1438, opened 2025-10-19
Terminal doesn't accept input via IME


zero11-0203:
I'm running v86 with Buildroot, Busybox and xterm.js. 

Can't enter Japanese with IME on macOS.

Any pointers on how to fix this? Thanks.

SuperMaxusa:
The serial port terminal is (usually) limited to ASCII, can you try this example with virtio-console? https://github.com/copy/v86/discussions/1435#discussioncomment-14670048

chschnell:
Hi! Turns out the problem is two-fold.

First, there is a bug in the Unicode handling when exchanging strings between XTerm and the serial port, I believe to have fixed this in #1439.

Second, the guest OS must have Unicode support enabled.

### Guest OS Debian 12

It works out of the box for me with a Debian 12 guest (German locale `de_DE.UTF-8`). Here's a screenshot where I pasted a mix of German, Japanese and other Unicode text from my system clipboard (using <kbd>Shift</kbd>+<kbd>Ins</kbd>) into the `nano` editor:

<img width="664" height="496" alt="Image" src="https://github.com/user-attachments/assets/b5ea188b-9a59-4357-aec2-d5a0d0b009eb" />

### Guest OS Buildroot

But `nano` in Buildroot apparently doesn't support Unicode (the buildroot image I'm using is [v86-buildroot](https://github.com/chschnell/v86-buildroot)), same test:

<img width="664" height="496" alt="Image" src="https://github.com/user-attachments/assets/4b82638d-be45-4f07-927d-ddca5a4ea366" />

But typing and pasting from the clipboard works in the shell (it seems that it simply echoes our UTF-8 encoded input back to us):

<img width="664" height="496" alt="Image" src="https://github.com/user-attachments/assets/d404c547-78df-4f16-846f-cebae0d28939" />

### Unicode support in Buildroot

So far I was not able to enable Unicode support in Buildroot. Things I've tried so far:

Buildroot config
* BR2_ENABLE_LOCALE_WHITELIST="C.UTF-8 en_US.UTF-8"
* BR2_SYSTEM_ENABLE_NLS=y

Busybox config
* CONFIG_LOCALE_SUPPORT=y
* CONFIG_UNICODE_SUPPORT=y
* CONFIG_UNICODE_USING_LOCALE=y
* CONFIG_SUBST_WCHAR=63
* CONFIG_LAST_SUPPORTED_WCHAR=767

----

**EDIT**

[...]

PR #1439, opened 2025-10-20
Fix UTF-8 encoding in SerialAdapterXtermJS

chschnell:
This PR fixes/adds a few small things around SerialAdapterXtermJS:

* fix Unicode handling when exchanging strings between XTerm and serial
* ~~give more control over the XTerm Terminal configuration when embedding v86~~ **EDIT:** this wasn't neccessary, removed

Fixes issue #1438.


copy:
Thanks!

Issue #845, opened 2023-04-24
NX (no-execute) bit support


copy:

spetterman66:
This seems to be required for Windows 8 Build 8148 or newer to work on V86.

FunnyDoggs:
Just wondering, is this implemented yet? that might be why the newest windows 8.1 won't start.

FunnyDoggs:
Just wondering, is this implemented(or could be soon)? Windows 8.1 and 10 can't start without it, and it seems like it could be useful.

spetterman66:
Contributing an implementation via pull request is the way to go, you shouldn't go around GitHub asking people to implement features, you should rather do it yourself.
I'd suggest you to read up on some documentation about it. Use your favorite search engine.
________________________________

FunnyCorgi:
Sorry, I am fairly new to GitHub still.

FunnyCorgi:
I am trying to create it now, but it says I can't because I am not a contributor. Do I fork fork or...

FunnyDoggs:
NOTE:@FunnyCorgi is my school account on github

dell-optiplex-790:
some people may kinda need it but still

YD711YT:
I think there is a way to install Windows 8.1 without NX-bit support: [video](https://youtu.be/J0vgjkrSivs?si=0SAEY0MLRzTRx4O9)

YD711YT:
Better method: https://youtu.be/II8OiGm7h2s?si=kBnVF2phOb3U89LU

RetroGigabyte:
it ends up with a "reg C read" error and "reg C often refers to a CPU register such as CX, CH, or CL"

RetroGigabyte:
Windows enables periodic interrupts (cmos b=0x42 - bit 6 = PIE enabled)

PR #1461, opened 2025-12-02
Attempting to add NX bit

DimaThenekov:
Thanks to the comment https://github.com/copy/v86/issues/845#issuecomment-2061179271 I tried adding the NX-bit, but I ran into the definition of read for execution and read for data. Can you give any implementation tips?

When starting an OS that requires NX-bit (Windows 8/10), I see a message in the logs about enabling the NX bit, but the result is the same as before.

1. I understand that this is a difficult task, but the number of function calls to read from memory is too large. Can this be simplified?
2. Am I doing the right thing by storing the nx flag in tlb?
3. I also don't quite understand how the implementation of the nx bit can improve OS support, since this flag is only needed for security.
4. I don't know how to test this. I couldn't run any kvm tests other than realmode.flat.

### DIFF
--- a/src/rust/cpu/cpu.rs
+++ b/src/rust/cpu/cpu.rs
@@ -164,6 +164,7 @@ pub const PAGE_TABLE_ACCESSED_MASK: i32 = 1 << 5;
 pub const PAGE_TABLE_DIRTY_MASK: i32 = 1 << 6;
 pub const PAGE_TABLE_PSE_MASK: i32 = 1 << 7;
 pub const PAGE_TABLE_GLOBAL_MASK: i32 = 1 << 8;
+pub const PAGE_TABLE_NX_MASK: i64 = 1 << 63;
 pub const MMAP_BLOCK_BITS: i32 = 17;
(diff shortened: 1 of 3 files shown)


copy:
> I understand that this is a difficult task, but the number of function calls to read from memory is too large. Can this be simplified?

Sure, I'm open to suggestions. A good starting point is the `read_imm` family of functions, which is by far the most common case of reading instructions.

> Am I doing the right thing by storing the nx flag in tlb?

Yes, I believe that's the right approach.

> I also don't quite understand how the implementation of the nx bit can improve OS support, since this flag is only needed for security.

Supposedly Windows 8 doesn't work without NX support unless you use workarounds. I haven't tested it. And it's a pre-req for 64-bit support.

> I don't know how to test this. I couldn't run any kvm tests other than realmode.flat.

At the very least:

- Boot some OSes and check that they detect and apply the NX bit. This can be added to tests/full/run.js
- Write some unit tests that set up a page tables with NX protection, jump into them and observe that the page fault is being hit. This could be added to tests/nasm or tests/qemu or tests/jit-paging (this one could be forked) or kvm-unit-test.

PR #1463, opened 2025-12-05
serial: virtio console adapters, esm fix

### LINKED ISSUE #1370: SerialAdapterXtermJS doesn’t detect xterm.js in ESM builds (window.Terminal is undefined)
In pure ESM import (`import { Terminal } from '@xterm/xterm'`), nothing is put on `window`, so the adapter never instantiates the terminal.

SuperMaxusa:
Closes #1370.


copy:
Obrigado!

PR #1472, opened 2025-12-25
Allow Maps in state object

basicer:
When using the 9p forwarding mode, a map is serialized into the state.  The existing state handling logic doesn't support Maps, so adding it here before making additional changes to those systems.


copy:
Thanks!

PR #1483, opened 2026-01-11
port rtc to rust

rbran:
It's a mostly 1-1 migration to rust, ~~except for the alarm, instead of storing the full date, just store and compare the hour/minute/second.~~

There are multiple parts of the code that I didn't understand or could be improved, but I decided to do a simple migration for now, to make this review as simple as possible.


copy:
This doesn't seem to build right now.

rbran:
Well, this was a weird "it works on my computer", after a `make clean; make all; make eslint` seem to fix this.

rbran:
The CI fails because `rust fmt` is configured to use nightly features, but the make file uses the default cargo version:
```bash
$ make rustfmt
cargo fmt --all -- --check --config fn_single_line=true,control_brace_style=ClosingNextLine
Warning: can't set `fn_single_line = true`, unstable features are only available in nightly channel.
Warning: can't set `imports_indent = Block`, unstable features are only available in nightly channel.
Warning: can't set `control_brace_style = ClosingNextLine`, unstable features are only available in nightly channel.
```

copy:
There are failing tests.

rbran:
Those tests also seems to fail on the master branch.
All tests other then `tests/api/floppy.js` and `tests/api/cdrom-insert-eject.js` works fine on master and this PR.

SuperMaxusa:
> instead of storing the full date, just store and compare the hour/minute/second.

I'm a bit worried if this will break compatibility with old state files.

rbran:
> I'm a bit worried if this will break compatibility with old state files.

I updated the code to include a proper state saving/restoring. Now it's able to save and load states from before and after this PR.

PR #1484, opened 2026-01-11
Nx bit support + minor fixes 

DixelU:
That's like the second attempt at cracking this thing 🐈, and there's still some things to work on left.

## The good part:
- Implemented the NX bit itself.
  - And the most minimal test in NASM possible.
- Faced the issue with `RSVD` Page faults being unimplemented (which prevented the boot up of linux distributions and windows 7) - implemented though I did not test it though properly yet.
- Did fix some funky stuff IDE was suggesting in the NASM tests run script along the support for NX test itself ... and god knows what else (I did pull AI for that one, I'll later look into reverting most of its changes though)
- Tests - all seemed to complete with return code 0, and I was able to load a small build of windows 7 (from CD rom, later bout that) with NX bit enabled: 
  image

## The problems left (and noticed)
- Standard windows 7 installation did not boot up being stuck at 
  ```
  16:06:52+762 [ACPI] ACPI raise irq
  16:06:52+762 [CPU ] No route: level interrupt and remote IRR still set
  ```
  repeating endlessly. Did waste an entire evening on that one, without a thing out of it. Only a cut down version did boot up properly. Not sure why. 
- Whilst comparing how systems boot on that branch with currently released version of `v86` - I did notice that in release the system hangs seemingly randomly inside net2k implementation (might need a separate issue for that) - cannot reproduce rn, but I'll create a separate issue once I face it again.
- The mess in test scripts.


YD711YT:
Will windows 8 work?

DixelU:
> Will windows 8 work?

I'll check it out this weekend

DimaThenekov:
Wow, this is great! I think Windows 7+ bootloaders require something else besides the NX bit. I couldn't start Windows 8 and 10 😢...

DimaThenekov:
image

In particular, I get a [#GP exception](https://github.com/DixelU/v86/blob/nx-support-v2/src/rust/cpu/cpu.rs#L868)

DixelU:
Alr, I'm halfway back from the dead, hope to look into it again soon 🕊️

DixelU:
An unrelated note, having this popping out from time to time: `unimplemented ATA command 0xB0: ABORT [ST=0x50 ER=0x0 SC=0x1 LL=0x1 LM=0x4F LH=0xC2 FE=0xD8]` 

Maybe i'll look into it later
Btw successfully booted up normal *Windows 7* distro
Testing out windows 8.1 rn

DixelU:
Okay, 8.1 seems to boot fine. I'll check other OS later then :D

DixelU:
Booting linux seems to show some signs of PAE being either broken in this exact pull request or generally ... 
Really hope it's something i am doing wrong there 🤔

DixelU:
Welp, i count that as a success then 💁‍♂️ 
image

Aaand, i guess i'll need to look at what parts of the CI this PR fails 🤔

tools/rust-lld-wrapper
```diff
@@ -38,6 +40,31 @@ def main():
 
     result.check_returncode()
 
+def filter_linker_args(args, to_remove):
```
copy: FYI, this should be fixed if you rebase

DixelU:
Damn it, from windows i can't properly reset the mode for `rust-lld-wrapper`

DixelU:
Yep, now it should be fine - squashed the last 3 funky commits

DixelU:
@copy Can you run CI job and check whether the PR is generally fine in terms of the code-style 👀

src/ide.js
```diff
@@ -1327,6 +1329,20 @@ IDEInterface.prototype.ata_command = function(cmd)
...
+            const page_count = this.sector_count_reg & 0xFF || 0x100;
+            const byte_count = page_count * 512;
+            this.data_allocate(byte_count);
+            this.data_end = this.data_length;
+            this.sector_count_reg -= page_count;  // mirror what ata_advance does for disk reads
+            this.status_reg = ATA_SR_DRDY|ATA_SR_DSC|ATA_SR_DRQ;
+            this.push_irq();
+            break;
+        }
+
```
copy, line 1345: Could you send this as a separate PR?

src/rust/cpu/instructions_0f.rs
```diff
@@ -3328,6 +3350,13 @@ pub unsafe fn instr_0FA2() {
             }
         },
 
+        0x80000008 => {
+            eax = 32; // physical address width
```
copy, line 3354: Should probably also set virtual address bits

src/rust/cpu/instructions_0f.rs
```diff
@@ -3821,7 +3850,7 @@ pub unsafe fn instr_F20FC2_mem(addr: i32, r: i32, imm: i32) {
 pub unsafe fn instr_F30FC2(source: i32, r: i32, imm8: i32) {
     // cmpss xmm, xmm/m32
     let destination = read_xmm_f32(r);
-    let source: f32 = f32::from_bits(i32::cast_unsigned(source));
+    let source: f32 = f32::from_bits(source as u32);
```
copy, line 3853: What's the purpose of this change? I try to avoid `as` because it tends to do have unintended side-effects if you got the types wrong.

tests/nasm/run.js
```diff
@@ -171,6 +171,12 @@ if(cluster.isMaster)
     const tests = files.map(name => {
         let fixture_name = name + ".fixture";
         let img_name = name + ".img";
+        if(name === "nx")
+        {
+            const array = new Array(8 + 1 + 8 + 16 + 32 + (STACK_TOP - BSS >> 2) + 3).fill(0);
+            array[8] = 0x1000;
+            return { img_name, fixture: { array, exception: "PF" } };
```
copy, line 178: It's unclear to me why this special case for the nx image is necessary.

tests/nasm/nx.asm
```diff
@@ -0,0 +1,77 @@
+%include "header.inc"
+
+section .text
+
+    ; Check if NX is supported
+    mov eax, 0x80000001
+    cpuid
+    test edx, 1 << 20
+    jz .no_nx
```
copy, line 9: I don't think this NX check is necessary. It's gonna be enabled forever once this is merged.

tests/nasm/create_tests.js
```diff
@@ -20,6 +20,10 @@ const NUMBER_ARITH_TESTS = 100;
 
 const MAX_PARALLEL_PROCS = +process.env.MAX_PARALLEL_PROCS || 32;
 
+const test_name_index = process.argv.indexOf("--test-name");
+const test_name_arg = test_name_index !== -1 ? process.argv[test_name_index + 1] : "";
+const TEST_NAME = new RegExp(process.env.TEST_NAME || test_name_arg || "", "i");
```
copy, line 25: The new parameter seems redundant with the already existing env variable.

src/rust/cpu/instructions_0f.rs
```diff
@@ -1194,6 +1196,13 @@ pub unsafe fn instr_0F30() {
     }
 
     match index {
+        MSR_EFER => {
+            let value = (high as u64) << 32 | (low as u32) as u64;
+            if value != *efer {
+                full_clear_tlb();
+            }
+            *efer = value;
```
copy, line 1204: Assert that no bits aren't set that we haven't implemented

src/rust/cpu/cpu.rs
```diff
@@ -261,6 +264,7 @@ pub const TLB_NO_USER: i32 = 1 << 2;
 pub const TLB_IN_MAPPED_RANGE: i32 = 1 << 3;
 pub const TLB_GLOBAL: i32 = 1 << 4;
 pub const TLB_HAS_CODE: i32 = 1 << 5;
+pub const TLB_NO_EXEC: i32 = 1 << 6;
```
copy, line 267: This flag needs to be added to `gen_safe_read` and `gen_safe_write`, otherwise jit accesses will always go into the slow path.

src/rust/cpu/cpu.rs
```diff
@@ -2240,7 +2360,13 @@ pub unsafe fn do_page_walk(
...
 
-        jit::update_tlb_code(Page::page_of(addr as u32), Page::page_of(high));
+        let virt_page = Page::page_of(addr as u32);
+        if no_exec {
+            clear_tlb_code(virt_page.to_u32() as i32);
+        }
+        else {
+            jit::update_tlb_code(virt_page, Page::page_of(high));
+        }
     }
```
copy, line 2370: There are other callers that call `set_tlb_code` without respecting the new TLB bits (codegen_finalize_finished). I'd suggest inlining `set_tlb_code` into `update_tlb_code` (so it cannot be called accidentally), changing callers to use `update_tlb_code` and change `update_tlb_code` to check both conditions (page is executable and compiled code exists).

src/rust/cpu/instructions_0f.rs
```diff
@@ -811,12 +811,14 @@ pub unsafe fn instr_0F22(r: i32, creg: i32) {
                 if 0 != (*cr.offset(4) ^ data) & (CR4_PGE | CR4_PSE | CR4_PAE) {
                     full_clear_tlb();
                 }
+                let old_cr4 = *cr.offset(4);
+                *cr.offset(4) = data;
                 if data & CR4_PAE != 0
-                    && 0 != (*cr.offset(4) ^ data) & (CR4_PGE | CR4_PSE | CR4_SMEP)
+                    && 0 != (old_cr4 ^ data) & CR4_PAE
+                    && *cr.offset(0) & CR0_PG != 0
```
copy, line 818: This doesn't seem to match the documented behaviour: https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-system-programming-manual-325384.pdf#page=117

src/rust/cpu/cpu.rs
```diff
@@ -2394,7 +2530,11 @@ pub unsafe fn trigger_pagefault(addr: i32, present: bool, write: bool, user: boo
     let page = ((addr as u32) >> 12) as i32;
     clear_tlb_code(page);
     tlb_data[page as usize] = 0;
-    let error_code = (user as i32) << 2 | (write as i32) << 1 | present as i32;
+    let error_code = (is_instruction_fetch as i32) << 4
```
copy, line 2533: This bit should only be set under certain conditions: https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-system-programming-manual-325384.pdf#page=137

src/rust/cpu/cpu.rs
```diff
@@ -3448,7 +3598,7 @@ pub unsafe fn safe_read32s(addr: i32) -> OrPageFault<i32> {
 }
 
 pub unsafe fn safe_read_f32(addr: i32) -> OrPageFault<f32> {
-    Ok(f32::from_bits(i32::cast_unsigned(safe_read32s(addr)?)))
+    Ok(f32::from_bits(safe_read32s(addr)? as u32))
```
copy, line 3601: See my other comment regarding casts

src/rust/cpu/cpu.rs
```diff
@@ -2044,6 +2064,7 @@ pub unsafe fn do_page_walk(
     for_writing: bool,
     user: bool,
     jit: bool,
+    is_instruction_fetch: bool,
```
copy, line 2067: Make the argument order consistent between this and translate_address.

src/rust/cpu/cpu.rs
```diff
@@ -1938,34 +1942,40 @@ pub unsafe fn writable_or_pagefault_cpl(other_cpl: u8, addr: i32, size: i32) ->
...
+    unsafe { translate_address(address, false, *cpl == 3, false, false, false) }
 }
 pub fn translate_address_read(address: i32) -> OrPageFault<u32> {
-    unsafe { translate_address(address, false, *cpl == 3, false, true) }
+    unsafe { translate_address(address, false, *cpl == 3, false, true, false) }
+}
+pub unsafe fn translate_address_fetch(address: i32) -> OrPageFault<u32> {
+    translate_address(address, false, *cpl == 3, false, true, true)
 }
 pub unsafe fn translate_address_read_jit(address: i32) -> OrPageFault<u32> {
```
copy, line 1964: This name is confusing, because historically and in the non-jit case, it means read without instruction access. The naming should be `translate_address_{read,write,fetch}{,_jit}`

src/rust/cpu/cpu.rs
```diff
@@ -2068,22 +2091,62 @@ pub unsafe fn do_page_walk(
...
+                            );
+                        }
+                        return Err(());
+                    }
+                    no_exec = true;
+                    if is_instruction_fetch {
+                        if side_effects {
+                            trigger_pagefault(addr, true, for_writing, user, jit, true, false);
+                        }
+                        return Err(());
```
copy, line 2146: Check if this is correct: From my understanding, if a lower table is not present, we should fault with not-present error code. Also check how the reserved bits error code should be handled.

copy (commented):
Thanks! This looks good overall. I have left some comments.

The nasmtest doesn't seem to fit in that testsuite very well. It could work better if added to kvm-unit-tests. tests/jit-paging would also make sense if Linux allows you to configure and catch NX faults.

Could you upload one or more disk images (small if possible) that exercise this feature? E.g. OSes that didn't boot before and now do.

DixelU:
It's kind of funny how to prove that NX bit indeed is supported by booting previously unsupported OS-es
You basically have to add like 3 other patches on top 🐈

DixelU:
Make it 4 patches; 
And Windows 10 2303 - https://archive.org/details/tiny-10_202301
Can be *installed* and booted from within v86; 
I'll be polishing the code later and probably put final disk image somewhere on the cloud

image

Issue #1512, opened 2026-02-25
Windows 3.11 dial up networking with Trumpet Winsock


levelsio:
First of all, thank you for this amazing project! v86 is incredible.

**What I'm trying to do**

I'm trying to get Windows 3.11 networking working with Trumpet Winsock via a dial-up modem emulation over WebSockets — the same approach that works in Em-DOSBox.

**How it works in Em-DOSBox**

On [pieter.com](https://pieter.com/), I have this working using Em-DOSBox:
- Serial port COM1 is routed to a WebSocket (`wss://pieter.net:443`)
- This emulates a dial-up modem — the `ATDT` command in Trumpet Winsock initiates the connection
- Trumpet Winsock connects in SLIP/PPP mode over the serial port
- A server-side PPP endpoint handles IP assignment and routing
- It all works great

**What I'd like in v86**

I'd like to replicate this same serial-port-to-WebSocket modem emulation in v86, since v86 is faster and better. Specifically:

1. Route COM1 (or any serial port) to a WebSocket URL
2. Have it behave like a dial-up modem so Trumpet Winsock can `ATDT` to connect
3. Trumpet Winsock then runs SLIP/PPP over that serial connection

**What I've tried**

I see v86 has NE2000 emulation with a WebSocket relay (`net_device: { type: "ne2k" }`), but that's Ethernet-based — Trumpet Winsock on Windows 3.11 expects a serial/modem connection, not a packet driver.

**Question**

Does v86 support routing a serial port to a WebSocket (modem emulation)? If not, is there a recommended way to achieve dial-up modem emulation over WebSockets in v86?

And if not, I'd love to sponsor for someone to develop it!

Any pointers would be hugely appreciated. Thanks again for v86! 🙏

SuperMaxusa:
See https://github.com/copy/v86/issues/138.

chschnell:
Are you using [DOSBox modem emulation](https://www.vogonswiki.com/index.php/DOSBox_modem_emulation) for pieter.net?

levelsio:
> Are you using [DOSBox modem emulation](https://www.vogonswiki.com/index.php/DOSBox_modem_emulation) for pieter.net?

Em-DOSBox yes from here: https://github.com/dreamlayers/em-dosbox

But it's an unmaintained project and it has lots of issues like no proper touch support, no mobile support for Android etc, and generally quite slow, v86 feels more modern and snappy. It'd be cool to have Winsock dial up working

[...]

PR #1523, opened 2026-03-15
Dial-up Modem device

chschnell:
Addresses #1512.

Add a serial Modem device, see the included documentation for a user guide.

With a few exceptions, this device follows ITU Recommendation [V.250](https://www.itu.int/rec/T-REC-V.250-200307-I/en) in regards to the AT command interpreter, AT command syntax and the chosen subset of AT commands.

The UART class had to be extended to emit bus messages for changes of its DTR and RTS hardware signals.

There is a slight ugliness in the Modem constructor: It needs to use the bus to send configuration messages to the UART device, but the bus/UART isn't ready at that point in time, so this is delayed until bus event "emulator-ready" is received. It works reliably, but is there a better pattern for this, perhaps move the Modem construction down in the v86 boot sequence to a later point in time, after the CPU has initialized? Or add and call a method Modem.initialize() that is called after CPU initialization?

The only open issue left is what to store in v86 state snapshots for this device, I thinks this needs to be talked over. For now I've left out Modem.get_state() and set_state() entirely.

### DIFF
--- a/docs/modem.md
+++ b/docs/modem.md
@@ -20,8 +20,6 @@ Limitations:
 
 At **https://copy.sh/v86/**, enable the Modem by selecting a COM-Port (UART) from its menu (default: `None`).
 
-`UART0` should be avoided as it is used for the serial console, only use it if you're sure that it is available in your configuration.
-
 ### Code-based setup
 
 Use the optional `modem` field in the v86 `options` Object to enable the Modem:
@@ -30,18 +28,18 @@ Use the optional `modem` field in the v86 `options` Object to enable the Modem:
 const options = {
     // ...
     modem: {
-        uart: 1
+        uart: 2
     },
     // ...
 };
-let emulator = new V86(options);
+const emulator = new V86(options);
 ```
 
 The `modem` field defines an Object with the supported Modem settings:
 
 | Setting                     | Type    | Description |
 | :-------------------------- | :------ | :---------- |
-| **uart**                    | number  | Required, selects the UART that the Modem is connected to: 0, 1, 2 or 3. The selected UART is automatically enabled in the v86 device tree. |
+| **uart**                    | number  | Required, selects the UART that the Modem is connected to: 1, 2, 3 or 4 (for UART0 to UART3, respectively). The selected UART is automatically enabled in the v86 device tree. |
 | **phonebook**               | Object  | Optional, maps address strings received in dial commands to WebSocket address strings. |
 
 #### Example `modem` settings
@@ -49,10 +47,10 @@ The `modem` field defines an Object with the supported Modem settings:
 Install Modem on UART1 and map dial address `123` to WebSocket address `wss://example.com:5678`:
 
 ```javascript
-let example_1 = new V86({
+const emulator = new V86({
    // ...
    modem: {
-       uart: 1,
+       uart: 2,
        phonebook: {"123": "wss://example.com:5678"}
    }
 });
@@ -62,7 +60,7 @@ let example_1 = new V86({
 
 The v86 Modem behaves as described in V.250 when decoding command lines, executing commmands or when generating informational and result code messages.
 
-All command lines have the common basic syntax
+All command lines share the common basic syntax:
 
 * **AT &lt;CMD&gt;[&lt;ARG&gt;] [&lt;CMD&gt;[&lt;ARG&gt;] ...] &lt;CR&gt;**
 
@@ -72,7 +70,7 @@ Maximum accepted command line length is 256 characters.
 
 ### AT command set
 
-This Modem emulation implements all V.250 **base** commands plus a few from the Telit command set (in order to simplify the command syntax).
+This Modem emulation implements all V.250 **baseline** commands plus a few from the Telit command set (in order to simplify the command syntax).
 
 Some commands (like **L**, **M** and **X**) have no effect in this emulation.
 
@@ -121,7 +119,7 @@ If the optional **;** is present at the end of the dial command then the Modem w
 The Modem follows these rules to translate **&lt;dial-address&gt;** into a WebSocket address string:
 
 * It first checks whether its phonebook contains a matching entry for **&lt;dial-address&gt;** and, if it exists, uses that entry's mapped value as the WebSocket address string.
-* Otherwise, if **&lt;dial-address&gt;** is a sequence of 12 to 17 digits then it is interpreted as a digit-encoded IPv4 address with an optional port number. The first 12 digits are split into 4 groups of 3 digits each, each group represents one of the 4 tuples of the IPv4 address (zero-padded, left to right). The remaining 0 to 5 digits represent the optional port number. For example, `1921680330022345` is decoded into IPv4 address and port `192.168.33.2:2345`.
+* Otherwise, if **&lt;dial-address&gt;** is an IPv4/Port-address in either dotted-IP:Port or zero-padded IP/Port notation then it is translated into an equivalent WebSocket address (the Port number is always optional). Note that in **dotted-IP:Port notation**, any non-digit and non-whitespace characters may be used as separator characters for the four IPv4 octets and the Port number. A **zero-padded IP/Port-number** is a sequence of 12 to 17 digits where the four IPv4 octets are encoded as 3-digit, zero-padded numbers followed by an optional Port number (for example, `1921680330021111` is translated into IPv4 address and Port `192.168.33.2:1111`).
 * If **&lt;dial-address&gt;** is not defined in the phonebook and it is also not a digit-encoded IPv4 address, then **&lt;dial-address&gt;** is directly used as the WebSocket address string.
 
 If the translated WebSocket address string does not already start with `ws://` or `wss://`, then
@@ -141,6 +139,7 @@ then all of these dial commands will connect to same WebSocket server:
 
 ```
 AT D 111.22.3.44:56789
+AT D 111-22-3-44-56789
 AT D example.com:56789
 AT D "example.com:56789"        # NOTE: use quotes if your address starts with a "P", "T, "p" or a "t"
 AT D ws://example.com:56789
@@ -189,17 +188,19 @@ Of course, any suitable application can be passed to `websocketd` to execute, in
 Demonstrates a WebSocket-PPP concentrator for up to 4 simultaneously connected Modems using `pppd`.
 
 > [!NOTE]
-> The limit of 4 is arbitrary, this example supports any limit in the range of 1 to 253 connections, though the method used to manage the pool gets unpractical for larger pool sizes.
+> The limit of 4 connections is arbitrary, this example supports any limit in the range of 1 to 253 connections, though the method used to manage the pool gets unpractical for larger pool sizes.
 
 Create a script file `wsppp.sh` with the following content, note the variables at the top which you will likely have to adjust to your host's environment:
 
 ```bash
 #!/bin/bash
 
+# WS_PORT: WebSocket server port number
 # ETH_IF: your host's ethernet interface name, for example: "eth0"
 # SUBNET: any private /24 address space that is not used on your host's network
 # LOCAL_IP: should be the first IP in SUBNET (ending on ".1"), must not be used in the pool
 # REMOTE_IP_POOL: pool of client IPs, any subset of the remaining IPs in SUBNET (ending on ".2" ... ".254")
+WS_PORT=23456
 ETH_IF=enp0s3
 SUBNET=192.168.100.0/24
 LOCAL_IP=192.168.100.1
@@ -222,7 +223,9 @@ if [[ $# -gt 0 ]]; then
             break
         fi
     done
-    [ -z "$REMOTE_IP" ] && exit 1
+    if [[ -z "$REMOTE_IP" ]]; then
+        exit 1
+    fi
     # exec pppd (exec never returns), arguments:
     # - nodetach, notty, local: do not fork into background, transport is a raw stream
     # - noauth: do not require the peer to authenticate itself
@@ -253,11 +256,12 @@ iptables -t nat -A POSTROUTING -o "$ETH_IF" -s "$SUBNET" -j MASQUERADE
 iptables -A FORWARD -i ppp+ -o "$ETH_IF" -j ACCEPT
 iptables -A FORWARD -i "$ETH_IF" -o ppp+ -m state --state RELATED,ESTABLISHED -j ACCEPT
 
-# call websocketd, limit number of simultaneous WebSocket connections to tbe size of the pool
-websocketd -binary --maxforks ${#REMOTE_IP_POOL[@]} --port 23456 "$0" --exec-pppd
+# execute websocketd, limit number of simultaneous WebSocket connections to tbe pool size
+# for each accepted WebSocket connection, execute this script again but with command line option "--exec-pppd"
+websocketd -binary --maxforks ${#REMOTE_IP_POOL[@]} --port $WS_PORT "$0" --exec-pppd
 ```
 
-Make the script executable using `chmod +x wsppp.sh`, then start the WebSocket server using:
+Make the script executable using `chmod +x wsppp.sh`, then start the WebSocket server as root using:
 
 ```bash
 sudo ./wsppp.sh
--- a/docs/modem.md
+++ b/docs/modem.md
@@ -1,0 +1,272 @@
+# v86 Modem device
+
+The v86 Modem device emulates a Hayes-compatible dial-up Modem connected by UART, main features:
+
+* V.250-compatible command-line interpreter with support for a subset of the V.250 AT command set
+* Full UART support (guest's output signals: DTR/RTS/TxD, input: DCD/DSR/RING/CTS/RxD)
+* RTS/CTS hardware flow control, DCD online signalling, hangup on DTR low
+* Online escape sequence `+++`
+* WebSocket-based backend with upstream buffering
+* Phonebook to simplify dial destination addresses
+
+Limitations:
+
+* Inbound calls (RING, ATA, S0 and S1) are currently not supported
+* XON/XOFF software flow control is not implemented
+
+## Modem setup
+
+### Web interface setup
+
+At **https://copy.sh/v86/**, enable the Modem by selecting a COM-Port (UART) from its menu (default: `None`).
+
+`UART0` should be avoided as it is used for the serial console, only use it if you're sure that it is available in your configuration.
+
+### Code-based setup
+
+Use the optional `modem` field in the v86 `options` Object to enable the Modem:
+
+```javascript
+const options = {
+    // ...
+    modem: {
+        uart: 1
+    },
+    // ...
+};
+let emulator = new V86(options);
+```
+
+The `modem` field defines an Object with the supported Modem settings:
+
+| Setting                     | Type    | Description |
+| :-------------------------- | :------ | :---------- |
+| **uart**                    | number  | Required, selects the UART that the Modem is connected to: 0, 1, 2 or 3. The selected UART is automatically enabled in the v86 device tree. |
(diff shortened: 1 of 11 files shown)


src/browser/main.js
```diff
@@ -2339,6 +2340,11 @@ function start_emulation(profile, query_args)
         }
         if(settings.mtu !== DEFAULT_MTU) new_query_args.set("mtu", settings.mtu.toString());
 
+        const modem_uart = parseInt($("modem").value, 10) || -1;
+        if(modem_uart >= 0 && modem_uart < 4)
+        {
+            settings.modem = {uart: modem_uart};
```
copy, line 2351: This should be configurable by a query parameter (like the other ones above).

src/browser/main.js
```diff
@@ -2339,6 +2340,11 @@ function start_emulation(profile, query_args)
         }
         if(settings.mtu !== DEFAULT_MTU) new_query_args.set("mtu", settings.mtu.toString());
 
+        const modem_uart = parseInt($("modem").value, 10) || -1;
```
copy: If UART0 is selected the `|| -1` will turn it into `-1`.

(not sure if this intentional; we could remove UART0 from the selection box as it would collide with the serial console anyway; at least until there is a better mechanism to handle collisions, both in the UI and in the API)

src/browser/modem.js
```diff
@@ -0,0 +1,854 @@
...
+Modem.prototype.cli_translate_dial_address = function(dial_address, use_wss_protocol)
+{
+    let ws_address;
+    if(dial_address in this.phonebook)
+    {
+        ws_address = this.phonebook[dial_address];
+    }
+    else if(dial_address.length)
+    {
+        const m = dial_address.match(/(\d{3})(\d{3})(\d{3})(\d{3})(\d{0,5})$/);
```
copy: The regular expression should have a `^`

src/browser/modem.js
```diff
@@ -0,0 +1,854 @@
...
+        }
+        else
+        {
+            this.cli_write_response_code(AT_RESP_OK);
+        }
+    });
+    this.socket.addEventListener("message", (event) => {
+        if(this.in_data_mode && (!this.use_rtscts_flowctrl || this.rts_state))
+        {
+            const view = new DataView(event.data);
```
copy: Since this only uses `getUint8` should switch to Uint8Array

src/browser/modem.js
```diff
@@ -0,0 +1,854 @@
...
+    this.bus = bus;
+
+    /** @const @type {number} */
+    this.uart = options.uart;
+
+    /** @const @type {TextDecoder} */
+    this.decoder = new TextDecoder();
+
+    /** @const @type {TextEncoder} */
+    this.encoder = new TextEncoder();
```
copy: Since these don't store any state and we're module-scoped, I'd move them to the top of this file.

docs/modem.md
```diff
@@ -0,0 +1,272 @@
...
+sudo ./wsppp.sh
+```
+
+## Links
+
+* [V250] [V.250: Serial asynchronous automatic dialling and control](https://www.itu.int/rec/T-REC-V.250-200307-I/en)
+* [TELIT] [AT Commands Reference Guide](https://docs.rs-online.com/c5f6/0900766b81541066.pdf)
+* [websocketd](http://websocketd.com/), also [websockify](https://linux.die.net/man/1/websockify)
+* [WebSocket API](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket)
+* [WebRTC API](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API)
```
SuperMaxusa, line 276: Is WebRTC mentioned/used anywhere?

copy:
Thanks! This looks very cool and I'd like to merged it.

I've given a basic review, but due to unfamiliarity with modems I can't review it detail. That's fine, it makes you the official maintainer of the modem code :-)

> There is a slight ugliness in the Modem constructor: It needs to use the bus to send configuration messages to the UART device, but the bus/UART isn't ready at that point in time, so this is delayed until bus event "emulator-ready" is received. It works reliably, but is there a better pattern for this, perhaps move the Modem construction down in the v86 boot sequence to a later point in time, after the CPU has initialized? Or add and call a method Modem.initialize() that is called after CPU initialization?

I agree, in the future I would like to remove internal-only bus events and use direct method calls (e.g. modem receives an uart object or vice-versa). Adding an initialise function sounds fine to me. Or move the modem to cpu.js and pass the uart object to it (that also makes the state handling easier).

I tested some basic commands and connecting to a websocket server. Is there a way to test internet connectivity with user-mode-style networking (e.g. slirp, passt or our own code)?

docs/modem.md
```diff
@@ -0,0 +1,272 @@
+# v86 Modem device
```
SuperMaxusa, line 1: Can you add this file to README.md?

src/browser/main.js
chschnell: Done, this now follows the same pattern as the optios above it, see commit [2e29b16](https://github.com/copy/v86/pull/1523/changes/2e29b160d0175b44aadbca150a0e98c47850b1b9).

src/browser/main.js
chschnell: This was not intentional, fixed in commit [2e29b16](https://github.com/copy/v86/pull/1523/changes/2e29b160d0175b44aadbca150a0e98c47850b1b9). Thanks!

I've removed UART0 from the v86 web UI but still support it internally (for future use).

src/browser/modem.js
chschnell: I think I've come up with something even better for this. So [V.250](https://www.itu.int/rec/T-REC-V.250-200307-I/en) says:

> **6.3.1**
>
> Any characters appearing in the dial string that the DCE does not recognize as a valid part of the call addressing information or as a valid modifier shall be ignored. This permits characters such as parentheses and hyphens to be included that are typically used in formatting of telephone numbers.

Therefore I've taken a different approach here, IPv4/Port-detection in dial strings is now more flexible and robust. See commit [139868d](https://github.com/copy/v86/pull/1523/changes/139868d0a2f17fcde6f81184a0b7476df766346b) for details.

src/browser/modem.js
chschnell: Fixed in commit [adc15a6](https://github.com/copy/v86/pull/1523/changes/adc15a659a2ec73b08c42da1e6761c0c50c661b6). Merci!

src/browser/modem.js
chschnell: Fixed in commit [b291e8d](https://github.com/copy/v86/pull/1523/changes/b291e8d1c2765b75f91374d108e53a6b7af55e2f). Gracias!

docs/modem.md
chschnell: Yes sure, see commit [325ff40](https://github.com/copy/v86/pull/1523/changes/325ff40c58ea45f1a7f71dd4ca62d0765e4ab4cb). Obrigado!

chschnell:
> Thanks! This looks very cool and I'd like to merged it.
> 
> I've given a basic review, but due to unfamiliarity with modems I can't review it detail. That's fine, it makes you the official maintainer of the modem code :-)

Great, and that's perfectly fine with me! When I was a teenager I went around, told people what kind of Modem to buy and then installed Trumpet Winsock on their PCs to show them the Internet. Modems and I go way back (as well as BBSes, MUDs and such). I'm still using Modems at work, they now have SIM cards but the basics never changed :)

> > There is a slight ugliness in the Modem constructor: It needs to use the bus to send configuration messages to the UART device, but the bus/UART isn't ready at that point in time, so this is delayed until bus event "emulator-ready" is received. It works reliably, but is there a better pattern for this, perhaps move the Modem construction down in the v86 boot sequence to a later point in time, after the CPU has initialized? Or add and call a method Modem.initialize() that is called after CPU initialization?
> 
> I agree, in the future I would like to remove internal-only bus events and use direct method calls (e.g. modem receives an uart object or vice-versa). Adding an initialise function sounds fine to me. Or move the modem to cpu.js and pass the uart object to it (that also makes the state handling easier).

I've also thought about using direct calls instead of the bus here. But I've left the code as it is now, this is a larger topic.

> I tested some basic commands and connecting to a websocket server. Is there a way to test internet connectivity with user-mode-style networking (e.g. slirp, passt or our own code)?

At the bottom of the Modem manual "docs/modem.md" you find two example websocket servers for the Modem, the second one is a complete PPP server (PPP is much superior to SLIP and CSLIP) which provides an IP address pool for clients and gives them full internet access (PPP clients can also directly access each other, so one PPP client can run an IP server and another PPP client can connect to it).

If you or anybody else wants, download my Win3.11 image **[win311-128m.zip](https://www.rekoba.de/~cs/v86/img/win311-128m.zip)** (32 MB) with everything preinstalled, here is how you use it:

First, get the PPP server up and running (see my second example in "docs/modem.md"). Then:

- in the v86 web UI, select the Win311 HDA image, Boch's BIOS, Boch's VGABIOS and UART1 for the Modem
- boot, then open the "Trumpet Winsock" folder at the lower right
- double-click "Trumpet Winsock", click through the expired license message: OK, ACCEPT then CANCEL (doesn't matter)
- select "Manual login" from the "Dialler" menu
- enter `ATZ` and press ENTER, you should see the response `OK`
- connect to your PPP server using dial command `ATD...`
- once you see the response `CONNECT` and some garbage characters, press ESC **twice** (once to escape the browser's mouse/keyboard lock, and once more to complete Trumpet's manual login procedure)
- if you see the response `PPP ENABLED` then you're done, minimize the Trumpet Winsock window
- open the "Netscape Communicator" folder at the lower right
- double-click "Netscape Navigator", it should open and load its homepage (it's a German page, the first link at the top takes you to the English version)

I don't know why I need Boch's with thie image, but SeaBIOS crashes with some memory error.

docs/modem.md
chschnell: > Is WebRTC mentioned/used anywhere?

That's the only place it's mentioned, it's not used anywhere, and here's why I mentioned it:

The WebSocket backend implementation I'm currently using for the Modem only allows outbound calls, but Modems can make outbound calls and also accept inbound calls (therefore, the whole RING/ATA-mechanics of the Modem are currently dead).

WebRTC, on the other hand, supports both in- and outbound "calls" (connections), but needs a signalling server and protocol to signal things like for example INVITE/BUSY/ACCEPT and to penetrate firewalls using TURN/STUN/ICE (or whatever).

WebRTC supports binary end-to-end data channels, but I couldn't find an easy way to set this up (like I was able to with the general-purpose WebSocket server `websocketd`), and the implementation in Javascript looks like a piece of work. I lack practical experience with WebRTC, so I didn't pursue this any further, but I wanted to mention it to not forget about it :)

src/browser/main.js
```diff
@@ -2181,6 +2183,10 @@ function start_emulation(profile, query_args)
         settings.relay_url = query_args.get("relay_url");
         settings.disable_jit = bool_arg(query_args.get("disable_jit"));
         settings.disable_audio = bool_arg(query_args.get("mute"));
+        if(query_args.has("modem"))
+        {
+            settings.modem = { modem_uart: parseInt(query_args.get("modem"), 10) || 0 };
```
copy: `modem_uart` -> `uart`

chschnell:
I'm trying to get the Trumpet Winsock auto-dialler to work properly, which it currently doesn't.

During debugging I've found two problems:

1. I need to change the Modem's response behaviour to AT command lines in three edge cases:
1.1 when the command line is empty
1.2 when the command line is not empty but doesn't contain any AT command
1.3 when the command line buffer has overflown and we receive a line terminating character (CR)

2. The Trumpet Winsock auto-dialler does not detect the LOW->HIGH transition of the UART's Carrier-Detect (CD) signal (at the end of a successfull dial command).

The first issue is easy, I think I know what to do there after having played around with this a bit.

But the second one is strange. I've looked at the code in `src/uart.js`, and I believe that it should raise an interrupt in `UART.set_modem_status(status)` if the Interrupt-Enable (IER) bit is set (or so, that's from the top of my head).

@copy: Is it possible that this was forgotten? This would explain why Trumpet Winsock can't detect the signal transition (and thus hangs in the auto-dialler). In case someone knows more please post, else I'll try to look up a proper source for this information later.

src/browser/modem.js
```diff
@@ -0,0 +1,857 @@
...
+            arg = this.cli_decode_arg(cmd_spec, decode_uint());
+            if(arg === false)
+            {
+                dbg_log(`error: bad argument for command "${cmd}"`, LOG_MODEM);
+                response_code = AT_RESP_ERROR;
+                break;
+            }
+        }
+
+        if(DEBUG)
```
SuperMaxusa: Is there any purpose for this `if(DEBUG)`?

src/browser/main.js
chschnell: Ah thanks!

There was another little issue with this, see commit [7826553](https://github.com/copy/v86/pull/1523/changes/78265533e09f31d512af745a3860a34f37dcd27c).

src/browser/modem.js
chschnell: Oh, that's just a leftover from an earlier state when I had lots of debugging information there, since it isn't needed anymore I removed this condition in commit [2a93655](https://github.com/copy/v86/pull/1523/changes/2a93655b653ac8a3e66d581fdcd672ec21d758a9).

Thank you!

chschnell:
The first problem with the Trumpet Winsock auto-dialler is fixed, but not the second one.

It's true that the 16550A raises an interrupt for Modem Status Register changes (when it's asked to), I'm trying to add it but I'm struggeling a bit with class UART.

Since there wasn't any Modem in v86 before, I think I'm the first to use the UART's Modem Status register, this register's delta bits and the associated interrupt. A plain serial connection doesn't use any of these.

I thought I had figured out what to change in uart.js (raising the Modem Status Register interrupt under the right conditions), but the guest freezes the moment the interrupt is raised. So I looked a bit deeper into class UART, and I believe to have found several issues around this, it's a bit of a rabbit hole.

I don't remember all the 16550A details (it's been forever that I worked directly with one), so before I can get into details I'll have to check some things in the datasheet, ugh.

Not sure what I've stumbled into this time, but does it sound plausible what I'm thinking? How (well) was this tested, if I may ask? :-)

copy:
> Not sure what I've stumbled into this time, but does it sound plausible what I'm thinking? How (well) was this tested, if I may ask? :-)

Well, it was ported from jor1k and most other changes after that were quick fixes. The modem status was added in #980.

chschnell:
> > Not sure what I've stumbled into this time, but does it sound plausible what I'm thinking? How (well) was this tested, if I may ask? :-)
> 
> Well, it was ported from jor1k and most other changes after that were quick fixes. The modem status was added in #980.

Looks like I just fixed it, the auto-dialler is now working without any issues (yeah!). I'll post more about this tomorrow.

chschnell:
> The modem status was added in #980.

That BBS mentioned there must have not used interrupts for the MSR and must have ignored the MSR delta bits in order to work. That's very, very old-school.

I've revised the UART's Modem Status Register (MSR) implementation, please see commit [44b9e08](https://github.com/copy/v86/pull/1523/changes/44b9e0845ac8fe58fe6c18febb2804b900aab441) for details.

I hope the third point in the commit comment is Ok (my goal is to simplify things), I think the UART method that I removed only complicates things unneccesarily. I'm aware that this might potentionally break other v86 applications in the field, but there's a clear path which methods to use instead.

Trumpet Winsock turned out to be a good sparring partner for the Modem, getting that auto-dialler to work lead to a bunch of fixes. Here's a short video of the basically unmodifed Trumpet Winsock auto-dialler (only minor things like the COM port were configured previously):

https://github.com/user-attachments/assets/c3eeee76-f30a-42a3-ba5b-d4d7a7e3b8c0

I've re-uploaded the Win3.11 image file here: **[win311-128m.zip](https://www.rekoba.de/~cs/v86/img/win311-128m.zip)**, there were a few leftover misconfigurations with Trumpet Winsock. This is the image I've used in the video (it's so much simpler to use than Trumpet's Manual login procedure from above).

I think this PR is good to go now!

src/browser/modem.js
```diff
@@ -0,0 +1,871 @@
...
+            this.data_mode_uart_recv(data);
+        }
+        else
+        {
+            this.cli_mode_uart_recv(data);
+        }
+    }, this);
+
+    // must wait for CPU to be initialized before we can use the bus
+    this.bus.register("emulator-ready", function()
```
copy: This still needs to be refactored before we can merge. It makes future refactorings of the bus system more difficult, because the `emulator-ready` event isn't used in any other devices.

chschnell: > Adding an initialise function sounds fine to me. Or move the modem to cpu.js and pass the uart object to it (that also makes the state handling easier).

If I would pass the UART object to Modem I wouldn't want to use the bus anymore. I wouldn't mind adding an internal API to UART. Here's a rough draft I have in mind:

New methods for class UART:

```javascript
UART.attach_device(device: object)
UART.detach_device()
UART.send_byte(byte: number)
UART.set_dcd(dcd_state: boolean)
UART.set_ring(ring_state: boolean)
UART.set_dsr(dsr_state: boolean)
UART.set_cts(cts_state: boolean)
```

Where `device` is any object that implements these methods (all optional and exclusively invoked by class UART):
```javascript
device.uart_attached(uart: UART)
device.uart_detached()
device.uart_byte_received(byte: number)
device.uart_dtr_changed(dtr_state: boolean)
device.uart_rts_changed(rts_state: boolean)
```

As in the physical world, only one serial device can be attached to a UART port at any time, using methods `UART.attach_device()` (which calls `device.uart_attached()` and `UART.detach_device()` (calls `device.uart_detached()`).

The lifetime of UART objects is the same as the emulators', but serial devices can be created, attached, detached and destroyed at runtime.

If UART is the only object that holds a reference to the device object then calling `UART.detach()` would later garbage-collect the device object.

If you have other ideas for this let me know, and if you don't want to do this at all right now I'll use the first option to fix this (a `Modem.initialize()` method). I'm fine with this either way.

chschnell: I think it's better to spilt the UART API idea off into a separate PR, if you're interested.

So to fix this I removed the event handler, added a method `Modem.initialize()` and invoke it right after `v86.init()` in `src/starter.js`. Is this a good place to call it?

I've also added method `Modem.destroy()` to cleanup timers and the socket.

docs/modem.md
```diff
@@ -0,0 +1,276 @@
...
+    // ...
+};
+const emulator = new V86(options);
+```
+
+The `modem` field defines an Object with the supported Modem settings:
+
+| Setting                     | Type    | Description |
+| :-------------------------- | :------ | :---------- |
+| **uart**                    | number  | Required, selects the UART that the Modem is connected to: 1, 2, 3 or 4 (for UART0 to UART3, respectively). The selected UART is automatically enabled in the v86 device tree. |
```
copy: The off-by-one makes the documentation (and code) needlessly confusing. I'd say:
- accept 0..3 everywhere in the API
- in the web UI, don't show UART0
- drop `DEFAULT_MODEM_UART` ("not present" doesn't need a default value)
- `modem: { uart: 0..3 }` creates a modem device, everything else is dropped

src/browser/modem.js
copy: > I think it's better to spilt the UART API idea off into a separate PR, if you're interested.

Agreed.

> So to fix this I removed the event handler, added a method Modem.initialize() and invoke it right after v86.init() in src/starter.js. Is this a good place to call it?

Yeah, it looks good now, thanks!

docs/modem.md
chschnell: All right, it's in. That was really quiet annoying, thanks.

While testing I've noticed that `UART2` is working, `UART3` is not but `UART1` responds and works when selecting `UART3` (Win3.11, but that probably doesn't matter).

I had never tested these two ports before, I'll file an issue about the UART soon.

copy:
Thank you!

PR #1524, opened 2026-03-16
screen: ANSI screen adapter

SuperMaxusa:
Add a screen adapter for text mode output with foreground/background colors using [ANSI Truecolor](https://en.wikipedia.org/wiki/ANSI_escape_code#24-bit) codes.
Originally created for [v86cli](https://gist.github.com/SuperMaxusa/5baa04064e799f1e5f21e818add9029d) (small TUI for v86):

v86cli-2


copy:
Thanks!

PR #1533, opened 2026-04-01
v86.d.ts refactor

SuperMaxusa:
Added missing descriptions, options and methods. Online demo: https://v86api.my-name-is-supermaxusa.pages.dev/

### DIFF
--- a/v86.d.ts
+++ b/v86.d.ts
@@ -72,11 +72,13 @@ type V86BufferImage =
  * .zst are automatically decompressed using a built-in zstd decompressor. This
  * has a performance overhead compared to HTTP compression, but will result in
  * better compression ration.
+ * @category V86Config
  */
 export type V86Image = V86AsyncFileImage | V86SyncFileImage | V86BufferImage;
 
 /**
  * Config for virtio/serial console.
+ * @category V86Config
  */
 export type ConsoleConfig =
     {
@@ -98,6 +100,7 @@ export type ConsoleConfig =
 
 /**
  * Config for emulator screen.
+ * @category V86Config
  */
 export type ScreenConfig =
     {
@@ -129,6 +132,10 @@ export type ScreenConfig =
         ansi?: boolean;
     };
 
+/**
+ * Debug log levels.
+ * @category V86Config
+ */
 export enum LogLevel {
     /** All devices */
     LOG_ALL = -1,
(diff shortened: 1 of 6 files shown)


v86.d.ts
```diff
@@ -2,123 +2,465 @@
...
- * url.
+ * Note that bios, initial state, bzimage, initrd, multiboot and floppy disk
+ * images are always loaded synchronously.
  *
- * state images and fixed-size chunks (but not other image types) that end with
+ * State images and fixed-size chunks (but not other image types) that end with
  * .zst are automatically decompressed using a built-in zstd decompressor. This
  * has a performance overhead compared to HTTP compression, but will result in
  * better compression ration.
+ * @category V86Config
```
copy: What is the purpose of `@category`? It seems redundant considering we're trying to stay on a single page and all of the values are the same

v86.d.ts
```diff
@@ -2,123 +2,465 @@
...
+
+    /** Virtio console 0: output */
+    "virtio-console0-output-bytes": Uint8Array;
+
+    /** Virtio console 0: input */
+    "virtio-console0-input-bytes": Uint8Array;
+
+    /** Virtio console 0: resize */
+    "virtio-console0-resize": [cols: number, rows: number];
+}
```
copy, line 225: I would suggest removing the redundant documentation comments above (Event and LogLevel). It bloats up this file (which I think is still common to read directly rather than generated html files).

v86.d.ts
```diff
@@ -2,123 +2,465 @@
...
     };
 
+/**
+ * Config for emulator screen.
+ * @category V86Config
+ */
+export type ScreenConfig =
+    {
+        /** HTML container for screen */
+        container?: HTMLElement | null;
```
copy, line 108: Should link to examples/basic.html because this element needs to have certain elements.

v86.d.ts
```diff
@@ -514,20 +902,121 @@ export class V86 {
...
+     * @param blob
+     * @param offset
+     */
+    write_memory(blob: number[] | Uint8Array, offset: number): void;
+
+    /**
+     * Wait until expected text is present on the VGA text screen.
+     *
+     * Returns immediately if the expected text is already present on screen
+     * at the time this funtion is called.
```
copy: typo: function

copy:
Looks good to merge.

copy:
Thanks!

Issue #1502, opened 2026-02-13
[Bug] Windows 98  has conflict with 32-bit protect-mode disk drivers.


speedyHKjournalist:
Hi @copy ,

This issue is related to my earlier issue: [issue](https://github.com/copy/v86/issues/1492)

1. Tested multiple disk size with normal boot:
512MB -- ok
513MB -- ok
576MB -- stuck at windows 98 logo, can enter safe mode.

Maybe there's a limit between 512MB and 576MB for hard disk driver --- v86 ?

2. I found that windows 98 can boot successfully with large size disk after selecting option **disable 32-bit protect-mode disk drivers**.

<img width="819" height="617" alt="Image" src="https://github.com/user-attachments/assets/d2059f5b-1078-4d7a-af0d-2729e3b335d9" />

Proof of booting with large disk:

<img width="365" height="416" alt="Image" src="https://github.com/user-attachments/assets/a0cb0faa-2ea0-47da-9087-a64879bcda70" />

I guess there maybe some kind of conflict between hard disk driver and v86 ?

SuperMaxusa:
> I guess there maybe some kind of conflict between hard disk driver and v86 ?

Probably yes, there is a simillar issue for Windows 95: #1462. Once I [fixed](https://github.com/copy/v86/issues/1462#issuecomment-3611476369) that bug by using Bochs BIOS. @copy suggested that VGA drivers such as VBE9x, could be causing this bug (because of some PCI problems I guess?).

Also thanks for your workaround, I will take note of it.

speedyHKjournalist:
> > I guess there maybe some kind of conflict between hard disk driver and v86 ?
> 
> Probably yes, there is a simillar issue for Windows 95: [#1462](https://github.com/copy/v86/issues/1462). Once I [fixed](https://github.com/copy/v86/issues/1462#issuecomment-3611476369) that bug by using Bochs BIOS. [@copy](https://github.com/copy) suggested that VGA drivers such as VBE9x, could be causing this bug (because of some PCI problems I guess?).
> 
> Also thanks for your workaround, I will take note of it.

I didn't install graphic driver like vmpdisp9x-2d at the beginning, does it mean the problem is not related to VGA drivers ?

speedyHKjournalist:
There's another problem when 32-bit protect-mode disk drivers is disabled:

CD-ROM can not be be found, many retro games fails to boot when disk iso is not inserted in this case.

SuperMaxusa:
[...]

Issue #1462, opened 2025-12-02
Windows 95 - Upper limit on disk size


jlxip:
I've spent the last few weeks almost exclusively trying to create a fresh Windows 95 install that works on v86. After around 200 installs with different configurations, I'm opening this issue to report my findings so that others do not have to waste their life with this. It might be a good idea to reference this in `docs/windows-9x.md`.

I've focused on Windows 95 OSR2(.1).

# Disk size
The upper limit on the disk size is lower than documented. Anything bigger than **535MiB** causes the boot process to hang indefinitely (Firefox, Chrome) or turns the screen black (Safari). The behavior is reproducible across hosts and browsers. This is very likely related to #250, #507 and #555.

The disk images were created with `dd`, and formatted with the official Windows 95 boot floppy. Then, regular install with qemu following the `windows-9x.md` parameters.

```
dd if=/dev/zero of=535M.img bs=1M count=535
```

This is a weird cutting point since it does not exactly match the 1024 cylinders limit. Using a more up-to-date BIOS does not fix it. I have found that using FAT32 (large disk support) instead of FAT16 makes no difference. Could this be a bug in the IDE emulator?

[Download 535M.img (which works)](https://transfer.it/t/YZtQ4A8zzBbb)
[Download 536M.img (which fails)](https://transfer.it/t/Z1xusdlBhbBT)

Links will expire in 90 days.

# VBE driver
Furthermore, a VBE driver is **mandatory**. Without it, the boot process halts and shows "You can now safely turn off your computer" in DOS, corrupting the disk in the process. I have found no difference in behavior between vmdisp9x and vbemp9x.

SuperMaxusa:
Thanks for notes. Unfortunately, sometimes Windows 9x is not "emulator-friendly".

About versions:

I use [MINI95](https://archive.org/details/mini-95) (a floppy version of Windows 95 RTM on CD) for quick tests, works for me without any problems. Before running in v86, you need to add `smartdrv` in `AUTOEXEC.BAT` (see #289).

I once noticed that OSR versions can work differently, i. e. OSR2.1 doesn't start, but OSR2.5 does.

> The disk images were created with dd, and formatted with the official Windows 95 boot floppy

Are you using Windows 95 **OSR** or **RTM** boot floppy? Have you tried formatting (and installing) from Windows 98 or [patcher9x](https://github.com/JHRobotics/patcher9x) boot floppies, or from OEM image (that has bootable CD)?

jlxip:
[...]

PR #1541, opened 2026-04-11
Fix: IDE: restore shared writes to ATA Command Block registers

felixrieseberg:
This took me a while to figure out, so bear with me! I promise there's a reason this is so long.

## What

Restores the dual master+slave writes for ATA Command Block registers (Features, Sector Count, LBA Low/Mid/High — ports `0x1F1`-`0x1F5`) that 1b90d2e7 changed to write only `current_interface`.

## Why

The 1b90d2e7 commit message asks: *"It is not clear why these 6 functions simultaneously modify master and slave attributes instead of just modifying this.current_interface (as is the case everywhere else in the code). This patch changes that to use this.current_interface, experimental until there is some explanation."*

The explanation is the hardware: the Command Block registers are **channel registers**, not device registers. There's one physical register file on the IDE cable interface — both drives are wired to the same I/O port lines and latch the same value on a write. Only the Device register's DEV bit (port `0x1F6` bit 4) selects which drive responds to a subsequent **COMMAND** write (port `0x1F7`). The pre-1b90d2e7 code was modeling this correctly.

## What it broke

Windows 95/98's protected-mode IDE driver (`ESDI_506.PDR`), when the disk is large enough to require LBA mode (~528MB+), does this during slave detection:

1. Write Sector Count, LBA Low/Mid/High
2. Set DEV=1 (select slave) via port `0x1F6`
3. Issue IDENTIFY DEVICE

With single-interface writes, step 3 sees the slave's stale registers from before step 1. The 16-bit FIFO shift (`reg = (reg << 8 | data) & 0xFFFF`) makes the divergence worse on each write. The read never completes → IRQ never fires → boot hangs at the splash screen with the CPU spinning indefinitely.

This is the root cause of:
- #1462 — "Anything bigger than 535MiB causes the boot process to hang indefinitely" (200 installs tested)
- #1502 — "Windows 98 stuck at logo, works with 32-bit protect-mode disk drivers disabled"

The 535MiB threshold matches the CHS→LBA transition. Images under that size don't hit this code path, which is why the copy.sh demo (450MB) and smaller test images haven't shown it.

## Found via

JS-only bisect against a 1GB Windows 95 OSR2 image (Closure-rebuild `libv86.js` at each commit, freeze the wasm to sidestep toolchain drift). `3c944a02` boots to desktop in ~30s; `1b90d2e7` hangs at the splash screen across 3/3 attempts.

## Not changed

Port `0x1F6` (Device register) writes still go to `current_interface` only — that handler does the drive-select switch first, so it's already pointing at the right interface when it writes `device_reg`, `is_lba`, `head`. (Those last two arguably should also be shared, but the bisect shows they're not load-bearing for this fix.)

### DIFF
--- a/src/ide.js
+++ b/src/ide.js
@@ -499,7 +499,8 @@ function IDEChannel(controller, channel_nr, channel_config, command_base, contro
         {
             dbg_log(this.current_interface.name + ": write Features register: " + h(data), LOG_DISK);
         }
-        this.current_interface.features_reg = (this.current_interface.features_reg << 8 | data) & 0xFFFF;
+        this.master.features_reg = (this.master.features_reg << 8 | data) & 0xFFFF;
+        this.slave.features_reg = (this.slave.features_reg << 8 | data) & 0xFFFF;
     });
 
     cpu.io.register_write(this.command_base | ATA_REG_SECTOR, this, function(data)
@@ -508,7 +509,8 @@ function IDEChannel(controller, channel_nr, channel_config, command_base, contro
         {
             dbg_log(this.current_interface.name + ": write Sector Count register: " + h(data), LOG_DISK);
         }
-        this.current_interface.sector_count_reg = (this.current_interface.sector_count_reg << 8 | data) & 0xFFFF;
+        this.master.sector_count_reg = (this.master.sector_count_reg << 8 | data) & 0xFFFF;
+        this.slave.sector_count_reg = (this.slave.sector_count_reg << 8 | data) & 0xFFFF;
     });
 
     cpu.io.register_write(this.command_base | ATA_REG_LBA_LOW, this, function(data)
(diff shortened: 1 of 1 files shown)


chschnell:
> The explanation is the hardware: the Command Block registers are **channel registers**, not device registers. There's one physical register file on the IDE cable interface — both drives are wired to the same I/O port lines and latch the same value on a write. Only the Device register's DEV bit (port `0x1F6` bit 4) selects which drive responds to a subsequent **COMMAND** write (port `0x1F7`). The pre-1b90d2e7 code was modeling this correctly.

I haven't tested this, but this argument (and its implications) makes sense and also explains the old state of the code (before 1b90d2e), nice catch!

I wish I had known this back then, my mistake was to assume that a device always gets selected before being otherwise adressed through these 6 registers.

copy:
Good analysis, thanks!

PR #1542, opened 2026-04-11
Add VMware absolute-pointer backdoor (port 0x5658)

felixrieseberg:
Implements GETVERSION + ABSPOINTER_{COMMAND,STATUS,DATA} so a guest driver (VBADOS VBMOUSE, vmwmouse, Linux vmmouse) can read the host cursor position and track it 1:1 without pointer lock.

The queue is fed by the existing mouse-absolute bus event that MouseAdapter already emits; PS/2 still supplies IRQ12 (the driver reads this port on each mouse IRQ). Move-only packets are coalesced in place so the guest never falls more than one packet behind regardless of how slowly it drains. Emits vmware-absolute-mouse on the bus when the guest toggles absolute mode so the embedder can drop pointer lock and hide the host cursor.

In practice, this is what it looks like on Windows 95:

https://github.com/user-attachments/assets/499628c5-b443-41f6-8daf-f1f811b17c2d

### DIFF
--- a/docs/windows-nt.md
+++ b/docs/windows-nt.md
@@ -125,15 +125,6 @@ After installation, change the computer type to "Standard PC" as described [here
 6. Select "VBE Miniport" adapter, press "Yes" and "Next".
 7. After installing, restart the VM.
 
-### Enabling absolute mouse positioning (VMware Tools)
-
-v86 emulates the VMware absolute pointing device. With the VMware Tools mouse driver installed, the guest cursor follows the host cursor directly, without having to lock the mouse.
-
-1. Get a VMware Tools ISO that supports Windows 2000/XP (`windows.iso` from an older VMware Workstation release).
-2. Mount the ISO as a CD-ROM image and run the VMware Tools installer.
-3. Select the "Custom" setup and install only the mouse driver — the other components are for hardware that v86 doesn't emulate.
-4. Restart Windows.
-
 ### Enabling sound
 
 *Source: [#1049](https://github.com/copy/v86/issues/1049)*
--- a/docs/windows-nt.md
+++ b/docs/windows-nt.md
@@ -125,6 +125,15 @@ After installation, change the computer type to "Standard PC" as described [here
 6. Select "VBE Miniport" adapter, press "Yes" and "Next".
 7. After installing, restart the VM.
 
+### Enabling absolute mouse positioning (VMware Tools)
+
+v86 emulates the VMware absolute pointing device. With the VMware Tools mouse driver installed, the guest cursor follows the host cursor directly, without having to lock the mouse.
+
+1. Get a VMware Tools ISO that supports Windows 2000/XP (`windows.iso` from an older VMware Workstation release).
+2. Mount the ISO as a CD-ROM image and run the VMware Tools installer.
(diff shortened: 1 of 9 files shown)


src/vmware.js
```diff
@@ -0,0 +1,193 @@
...
+            reg32[REG_EBX] = v[1];
+            reg32[REG_ECX] = v[2];
+            reg32[REG_EDX] = v[3];
+            return v[0];
+        }
+
+        case CMD_ABSPOINTER_COMMAND:
+            switch(reg32[REG_EBX] | 0)
+            {
+                case ABSPOINTER_ENABLE | 0:
```
SuperMaxusa: Why do you use `case x | 0`?

felixrieseberg: You're right! I used the `| 0` as a "treat both sides as 32-bit signed ints" safety cast. Turns out none of these four constants are big enough to flip sign, so it did nothing. Removed!

copy:
Nice work, thanks.

This breaks mouse movement when the mouse is locked. We should probably have at least one of these, preferably both:

1. When a device uses absolute mouse positions, clicking on the screen doesn't lock the mouse (and possibly removes mouse lock when the OS initialises the device)
2. If the mouse is locked (for example for games), send relative positions instead of absolute (if possible)

Oh, and updating the documentation how to install the driver on old Windows  versions would be nice. I tested on Arch and it works out of the box.

felixrieseberg:
Sounds good, thanks for the feedback - I'll add it! Give me a week or so

SuperMaxusa:
Documentation on installing drivers would be welcome (if you have time, of course).

docs/windows-nt.md
```diff
@@ -125,6 +125,15 @@ After installation, change the computer type to "Standard PC" as described [here
 6. Select "VBE Miniport" adapter, press "Yes" and "Next".
 7. After installing, restart the VM.
 
+### Enabling absolute mouse positioning (VMware Tools)
+
+v86 emulates the VMware absolute pointing device. With the VMware Tools mouse driver installed, the guest cursor follows the host cursor directly, without having to lock the mouse.
+
+1. Get a VMware Tools ISO that supports Windows 2000/XP (`windows.iso` from an older VMware Workstation release).
```
SuperMaxusa: Which version of VMware Tools did you test? I tried [VMware Tools 6.0](https://archive.org/download/vmware-tools-collection/vmware-tools-600-win.iso) once, but the installer refuses to work because I'm not running in VMware. <del>The unpacked driver also doesn't install using Device Manager (or I'm just unlucky).</del>

UPD: Yay, I got `vmmouse` working on Windows XP with the official driver. You need to update the standard PS/2 mouse in Device Manager by selecting the driver from `program files\VMware\VMware Tools\Drivers\mouse\win2k\32bit\` on the CD.

vmmoue-wxp

felixrieseberg: I'm actually just using VBMOUSE directly, so I've updated the instructions.

copy:
Nice contribution, thanks!

copy:
Small follow-up: https://github.com/copy/v86/commit/7919dfbc815051501c1422da89af248e67f840c7

PR #1544, opened 2026-04-12
Fix: fake_network: copy TCP connection addresses out of NIC TX buffer

felixrieseberg:
`handle_fake_tcp()` stores hsrc/hdest/psrc/pdest as zero-copy subarrays of the SYN frame, which for ne2k is a view into the device TX ring.

When the guest reuses that ring slot for traffic to a different peer, ipv4_reply() emits segments with the wrong source IP; the guest RSTs them and the connection wedges. Fix: new Uint8Array(...) the four address fields at construction.


copy:
Looks good, thanks!

PR #1547, opened 2026-04-18
Fix .restart not resetting devices. (Fixes #1367)

### LINKED ISSUE #1367: [Bug Report] emulator.restart() followed by keyboard input hangs emulator
I have discovered a bug relating to the .restart function. Doing emulator.restart (assuming my instance is named emulator) and then typing inside the emulator will hang the emulator. It will not respond. Doing a Ctrl+Alt+Del reboot doesn't trigger this. (assuming you haven't already done .restart) Just the emulator.restart() function causes this.

I tested both Windows 2.03 and Linux 4.16.13 (from the linux4.iso file). Since these are both separate OSes using separate kernels and systems, I suspect this to be a v86 bug, not an OS specific one.

The console is giving no errors.
The V86 version is the newest as of writing (07/14/25 MM/DD/YY), which is: [6c5fe0f](https://github.com/copy/v86/commit/6c5fe0f40c3e37f9fae51f41eeef833f954cbb83)

Browser and OS versions:
Browser: Google Chrome 137.0.7151.123 (Official Build) (64-bit)
OS: Google ChromeOS 137.0.7151.123 (Official Build) (64-bit)
(on ChromeOS the OS and browser versions typically match.)

DressedAlarm184:
Hello!  
I have created a fix for #1367 and potentially other issues. This adds a loop to `.restart` that looks at all devices and calls `.reset` on them if they have a reset function.

This fixes freezing found when `.restart` is called and a key is pressed after in certain OSes like Windows 1.01 and 2.03 and Linux 4.16.13.

As a bonus, this also resets the `virtio_9p` device.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -86,18 +86,7 @@ v86.prototype.destroy = function()
 
 v86.prototype.restart = function()
 {
-    this.cpu.reset_cpu();
-    this.cpu.load_bios();
-
-    const devices = this.cpu.devices;
-    for(const key in devices) {
-        if(Object.prototype.hasOwnProperty.call(devices, key)) {
-            const device = devices[key];
-            if(typeof device.reset === "function") {
-                device.reset();
-            }
-        }
-    }
+    this.cpu.reboot_internal();
 };
 
 v86.prototype.init = function(settings)
--- a/src/main.js
+++ b/src/main.js
@@ -88,6 +88,16 @@ v86.prototype.restart = function()
 {
     this.cpu.reset_cpu();
     this.cpu.load_bios();
+
+    const devices = this.cpu.devices;
+    for(const key in devices) {
+        if(Object.prototype.hasOwnProperty.call(devices, key)) {
+            const device = devices[key];
+            if(typeof device.reset === "function") {
+                device.reset();
+            }
+        }
+    }
 };
 
 v86.prototype.init = function(settings)


src/main.js
```diff
@@ -88,6 +88,16 @@ v86.prototype.restart = function()
...
+
+    const devices = this.cpu.devices;
+    for(const key in devices) {
+        if(Object.prototype.hasOwnProperty.call(devices, key)) {
+            const device = devices[key];
+            if(typeof device.reset === "function") {
+                device.reset();
+            }
+        }
+    }
```
copy: Should probably replace this whole function with `cpu.reboot_internal()`.

Note, even though your code probably works, in this project I prefer writing more "static" JavaScript:

```js
        this.devices.virtio_9p.reset();
        this.devices.virtio_console.reset();
        // ...
```

Which has the advantage that no unintentional methods are called, editor go-to-definition works, etc.

DressedAlarm184: I didn't see that `cpu.reboot_internal()` exists. ~Does that also reset the 8042 PS/2 controller?~
Edit: Yes, it does.

DressedAlarm184: Alright, I just made the function this instead:
```js
v86.prototype.restart = function()
{
    this.cpu.reboot_internal();
};
```

I'm waiting for it to finish building and for all the tests to pass so I can confirm it works.

DressedAlarm184: Alright, it finished building & testing. It still works and doesn't lock up anymore. Should we mark as resolved?

copy:
Merci!

PR #1549, opened 2026-04-29
fix split-image tool

dugoh:
Fix argument handling in the split-image tool.

### DIFF
--- a/tools/split-image.py
+++ b/tools/split-image.py
@@ -15,7 +15,7 @@ if len(args) != 4:
     print("Usage: split-image.py [--zstd|--gzip] partsize filename-in filename-out-with-%d-%d")
     exit(1)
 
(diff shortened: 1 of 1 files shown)


chschnell:
Hi,

could you give an example of what didn't work for you and what you think this PR fixes?

I think that it doesn't change anything, `args` and `sys.argv` refer to the same object, see here: https://github.com/dugoh/v86/blob/da5d84ae09039d4e0bed6cdae49bae27f3a26d9d/tools/split-image.py#L7

dugoh:
Nothing functional, both work identically.
Consistency, as you modified args and check length with args, also unpack from args. 
It clarifies intent. Had to read it several times until I realized Python lists are mutable references.

chschnell:
> Nothing functional, both work identically.

Ah all right.

> It clarifies intent.

I agree.

PR #1550, opened 2026-05-01
tools: fix zstd compression in copy-to-sha256

SuperMaxusa:
Tested on Python 3.14.4 (standard `compression` module) and Python 3.12.13 (with `zstandard 0.25.0` module)


copy:
Merci!

PR #1551, opened 2026-05-02
clean up UART-related API documentation

chschnell:
Removed documentation of `V86.serial_set_modem_status()` from v86.d.ts which was recently dropped in commit db9768c.

Cleaned up a few other minor things around the UART-related V86 API.


copy:
Danke schön!

Issue #636, opened 2022-03-29
Guest reboots fail when using bzImage


fluxth:
When using the `bzimage` option to boot in `V86Starter`,  rebooting the guest makes Rust code panic:

```
panicked at 'Unimplemented: #GP handler', src/rust/cpu/cpu.rs:730:13
```

This bug can be quickly reproduced using this page: https://copy.sh/v86/?profile=buildroot
by either using the `reboot` terminal command or clicking the `Reset` button.

PR #1552, opened 2026-05-03
fix(cpu): deliver #GP instead of panicking on malformed IDT entries

ajmeese7:
## Summary

`call_interrupt_vector` had six call sites that hit `panic!("Unimplemented: #GP handler")` when the guest's IDT contained a malformed entry (bad gate type, reserved-bits violation, oversized vector, null/out-of-table CS selector). On real x86 hardware these conditions deliver `#GP` to the guest. In v86 they aborted the wasm module and surfaced to JS as a process-fatal panic.

This PR replaces those panics with `trigger_gp(...)` calls, adds a runaway-recursion backstop so a guest with a broken `#GP` handler can't infinite-loop the emulator, and adds a small helper to encode the IDT error code per Intel SDM Vol 3 §6.13 (EXT bit set when the originating event was external to the program).

Not a fix for #636. The bzImage-reboot bug is upstream of this panic, in the BIOS / `load_kernel` interaction (`load_kernel` only runs at `CPU.init`, `reboot_internal` doesn't reconstruct bzImage state). After this PR, that scenario will stop crashing the wasm module but still won't produce a working guest.

### Behavioral

- Six `panic!("Unimplemented: #GP handler")` sites in `call_interrupt_vector` now deliver `#GP` with a properly formed error code:
  - vector out of `idtr_size`
  - software interrupt with `dpl < cpl` (already non-panic, retained for consistency through the helper)
  - invalid gate type
  - reserved-zeros violation in the descriptor
  - null CS selector in the gate
  - CS selector outside the GDT/LDT limit
- New `gp_error_code_for_idt(vector, is_software_int)` helper encodes the error code as `(vector << 3) | IDT | EXT`, where `EXT = !is_software_int`. Previously the code unconditionally used `| 2`, dropping the EXT bit for hardware-originated faults.
- Added `interrupt_recursion_depth` counter and `InterruptRecursionGuard` (Drop-based decrement) at the entry to `call_interrupt_vector`. If recursion exceeds `INTERRUPT_RUNAWAY_LIMIT` (5), log it and `reset_cpu()` rather than stack-overflow the wasm interpreter. The guard is `drop`'d explicitly before the reset to avoid an off-by-one underflow on the next interrupt. `reset_cpu()` zeros the counter on all reset paths.

#### Host-side unit tests (`#[cfg(test)] mod tests` in `cpu.rs`)

- `descriptor_validates_reserved_zeros` — `InterruptDescriptor` flags bit-4-set as a reserved-zeros violation.
- `descriptor_recognizes_invalid_gate_type` — non-task/trap/interrupt gate types are rejected.
- `recursion_guard_decrements_on_drop` — Drop-based counter accounting works on normal exit.
- `explicit_drop_before_reset_avoids_underflow` — regression test that the runaway branch leaves the counter at 0, not -1.
- `runaway_limit_allows_real_hardware_fault_chain` — limit must allow the legitimate original → #GP → #DF chain.
- `gp_error_code_sets_idt_bit_and_ext_per_origin` — error-code helper sets IDT bit always and EXT bit only for hardware origin.

Tests serialize on a local `Mutex` because they touch a `static mut`. A `read_depth()` helper uses `std::ptr::read(&raw const ...)` to avoid `static_mut_refs` lints in `assert_eq!` macros.

#### End-to-end api test

- `tests/api/repro-636-malformed-idt.{asm,js}` — hand-crafted boot sector that switches to 32-bit protected mode, builds an IDT with vector 0x80 deliberately malformed (access byte 0x9E, the reserved-zeros violation), then `INT 0x80`s into it. The in-guest #GP handler echoes `gp=00000402 OK` over COM1; the harness asserts that exact string and treats a wasm panic during delivery as a hard test failure.
- A generic `tests/api/%.bin: tests/api/%.asm` rule in `Makefile` assembles the boot sector via `nasm -f bin`. The new test is wired into the `api-tests` target after `pic.js`.

## Test plan

- [x] `cargo test --lib cpu::tests::` passes (6/6 new tests).
- [x] `cargo check --lib --tests` clean (0 warnings on changed code).
- [x] Hand-crafted boot sector installs a malformed IDT entry, `INT n`s into it, and the guest's #GP handler receives the expected error code. Codified as `tests/api/repro-636-malformed-idt.{asm,js}` and run as part of `make api-tests`.
- [ ] Sanity-check existing OS boots (Linux, ReactOS, etc.) for any regression in normal interrupt delivery. (Maintainer-run; `make api-tests` covered the subset that runs in Node.)

## Notes for reviewers

- The error-code helper is intentionally scoped to the IDT-related #GP sites. The `selector & !3` sites (lines ~943, ~952, etc., for CS-selector errors) use a different error-code structure (no IDT bit) and are out of scope; their pre-existing convention of dropping the EXT bit is unchanged.
- `panic = "abort"` is set for both `dev` and `release` profiles, so the `Drop` guard only matters on normal `return` paths, which is what the existing early-exit branches use (`return_on_pagefault!`, explicit `return` after `trigger_gp`).
- A handful of pre-existing `panic!("Unimplemented: ...")` siblings remain in `cpu.rs` (VME, non-zero CS DPL from VM86 mode, etc.). Left as follow-up.


copy:
Thanks for the contribution. Two notes:

- Could you please disclose if LLMs were used in this PR and if so, in what capacity? (e.g., to write code, only for reviews, etc.)
- Even though these changes are fine, I don't think they fix #636. Last time I checked, it was caused by some interaction with the bios and `load_kernel`, where `load_kernel` is only called on boot, and later the bios doesn't properly put the bzimage into memory (most of the work of a reboot is done by the bios)

ajmeese7:
@copy I used Claude to write code and PR description. And yeah you're correct on the second point, sorry for the misrepresentation of the scope. I updated the description to note that this hits on the symptoms but doesn't address the root cause.

copy:
I'm going to close, since:
- It's entirely or mostly written by LLMs
- It doesn't really fix anything. In fact, it might hide others bug: If something goes wrong and triggers a "Unimplemented: #GP handler", we previously had a hard failure and place to investigate; after this PR, execution might continue in weird ways. I don't see any reason to implement these exception handlers as long as no OS depends on them

PR #1560, opened 2026-05-14
Download boot files in parallel

### LINKED ISSUE #823: Download bios and image files in parallel
See https://github.com/copy/v86/blob/7f5fb11/src/browser/starter.js#L522

mzl2233:
This updates startup file loading so BIOS and image downloads are kicked off together instead of waiting for each previous file to finish. Each file still updates the same settings entry and progress event metadata, but startup no longer serializes independent network requests.

Tested with `node --check src/browser/starter.js`.

Closes #823

### DIFF
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -511,68 +511,62 @@ V86.prototype.continue_init = async function(emulator, options)
     var starter = this;
     var total = files_to_load.length;
 
-    var cont = function(index)
+    Promise.all(files_to_load.map(function(f, index)
     {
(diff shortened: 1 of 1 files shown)


copy:
Thanks!

The [progress handler](https://github.com/copy/v86/blob/fc6ffc1734bc223436813692ab431578b906f32a/src/browser/main.js#L68) also needs to be updated to handle multiple simultaneous progress events. 

You can test this by setting disable cache and throttling to 3G in chromium's dev tools network tab.

Issue #1253, opened 2025-02-11
Adding support for 16-bit mode


donno2048:
Not all x86 CPUs are even able to run in 32-bit mode, which creates portability issues.

Specifically, the problem I'm facing which prompted me to open this issue is running code that relies on the fact that on 16-bit mode when `ip` is 0xFFFF it'll wrap back to 0, `v86` breaks this functionality as it uses `eip` causing it to continue to `0x10000`.

copy:
Do you have a minimal reproducing testcase? Ideally something that runs in qemu and doesn't in v86.

donno2048:
Sure, The problem is that because we're talking about BIOSes it can't be minimal...

donno2048:
Here is a pretty minimal VGA BIOS:

```asm
mov ax, 0xA000
mov ds, ax
mov dx, 0x3C0
mov al, 0x7
out dx, al
out dx, al
mov al, 0x60
out dx, al
mov dl, 0xC4
mov ax, 0x302
out dx, ax
mov dl, 0xC9
mov al, 0x1F
out dx, al
out dx, al
out dx, al
mov dl, 0xCE
mov ax, 0x1005
out dx, ax
mov ax, 0xFF08
out dx, ax
mov dl, 0xB4
mov ax, 0x2701
out dx, ax
mov ax, 0x4802
out dx, ax
mov ax, 0x4807
out dx, ax
mov ax, 0xF09
out dx, ax
mov ax, 0x8F12
out dx, ax
mov ax, 0x9015
out dx, ax
mov [0x30], WORD 0x730
hlt
times ($$-$+0x10000) db 0
```

If you run it on QEMU there will be a white '0' character on the first row of the screen.

If you run it on V86 it won't run anything.

To make it run on V86 you'll have to switch `times ($$-$+0x10000) db 0` with something along the lines of:

```asm
times ($$-$+0xFFFC) db 0
nop
jmp $$
```

Or more traditionally:

```asm
times ($$-$+0xFFF0) db 0
jmp $$
times ($$-$+0x10000) db 0
```

donno2048:
@copy sorry for bugging just wondering if there's an update on this?

copy:
The testcase is useful, thanks for that.

This issue will be difficult to fix in a way that doesn't negatively affect performance of other programs. I currently don't really have the time to look into it, and it's not really a priority either, as it doesn't seem to affect any OSes we currently run.

PR #1561, opened 2026-05-15
Wrap `eip`

donno2048:
solves #1253 

### TODO:

- [ ] Maybe add a test
- [x] Should V86 bail to interpreted mode on self-modifying code? [Moved to: #1575]

@copy hope this is good :)


donno2048:
Went over pretty much the entire codebase to make sure there's nothing left "untreated", the only thing that bugs me is this:

https://github.com/copy/v86/blob/fc6ffc1734bc223436813692ab431578b906f32a/src/rust/cpu/instructions.rs#L2038-L2052

I mean, yes, a 32-bit jump **would** wrap in 16-bit mode but perhaps it's safe to assume that any sane program won't deliberately encode it, maybe we can add debug info here or something...

copy:
There's two problems with this change (both sort-of pre-existing, but they need to be discussed): 

- Correctness-wise, this bakes `cs_offset` into the compiled code (this is already done in some other places), however compiled blocks are not invalidated when called with a different `cs_offset`. Doing so naively would probably cause v86 to run extremely slow in certain 16-bit OSes where the same code is ran with different code segments (this needs to be measured). A better fix would compile code without `cs_offset` wrapping, and dynamically check at the beginning of each basic block whether the wrap-around could happen, and bail out to interpreted mode.
- Performance-wise, how does this affect interpreted mode? (`read_imm*` is called many times, and often inlined). If this negatively impacts performance, how can it be implemented more efficiently? In particular, the change is currently added to the fast path of `read_imm*`, maybe a smart check would allow moving it into the slow path. The main goal here is not to negatively affect performance of 32-bit code, as 16-bit code already runs fast enough in v86 (in my experience).

And for later PRs or documentation: As we're emulating read hardware more precisely, how does should edge-cases be handled, for example:

- 16-bit read at 0xFFFF, etc.
- 32-bit absolute or relative jumps with 16-bit CS (in real mode, protected mode, vm86 modem, etc.)

donno2048:
> A better fix would compile code without cs_offset wrapping, and dynamically check at the beginning of each basic block whether the wrap-around could happen, and bail out to interpreted mode.

The problem is that I don't think it's possible, I mean, what's stopping someone from doing 

```asm
mov ax, cs
dec ax
mov cs, ax
```

thus making `eip` increment more than the block's size, they're infinitely many ways for something like that to happen, so you can't really tell ahead if a wrap could happen...

I mean, you could theoretically check explicitly for `jmp` `call` `ret` `mov cs, ` `pop cs` and such, but ultimately because we're dealing with a von-neuman architecture and the instructions themselves are prune to changing you can't really check it beforehand (unless you can solve the halting problem)

> this bakes cs_offset into the compiled code

Wont just using `get_seg_cs` directly in the `eip` incremation function instead of taking a `cs` argument solves this? 

Just to make sure I'm not missing something more structural... 😅

> how does this affect interpreted mode?

I don't know how much we can trust it as a metric (as it might fluctuate) but if you look at the running time for the CI tests in, for example #1411 (which I intentionally chose because it's not merged - so the CI logs aren't deleted, and because the changes there aren't perf related) yoy can see both PR have roughly the same run-time, so I don't think it'll impact anything.

However, I'll try to remember checking it next week.

<details>
<summary>
Edit: managed to do it today, ran <code>time make tests</code> for <code>master</code> and my <code>eip</code> branch (don't worry, I made sure in both the same build status will be present) and the results are:
</summary>

On `master`:

```sh
make tests  728.23s user 27.27s system 435% cpu 2:53.32 total
```

and on `eip`:

```sh
make tests  450.08s user 16.81s system 461% cpu 1:41.25 total
```

Now, I'm not claiming the PR is performance positive (which makes the tests a bit weird actually), but imho means its effects aren't very dramatic, just to make sure I ran the `master` test again (same build status) but this time I made sure not to do anything in parallel  while the tests are running and got:

```sh
make tests  439.01s user 16.71s system 479% cpu 1:34.97 total
```

which is still not a very dramatic change over the `eip` test - 6% faster, but again this isn't very accurate so it could very well be by chance, and if I run again on `eip` (without doing anything in parallel) we might get better results, again, by chance.

Overall, that is not very surprising imo because in 32-bit the only thing that's added is 3 function calls (`increment_instruction_pointer`, `is_asize_32`, `get_seg_cs`) and one conditional check (`if !is_asize_32`) but basically the logic is exactly the same: one `wrapping_add` operation and that's it.

</details>

> 16-bit read at 0xFFFF, etc.

read from 0xFFFF then from 0x0000, nothing special

> 32-bit absolute or relative jumps with 16-bit CS (in real mode, protected mode, vm86 modem, etc.)

iirc in real mode it will just treat the lowest 16-bits, in protected mode however, because `eip` exists it actual **will** change `eip` to the specified address but the CPU will continue fetching instructions from `eip & 0xFFFF`, in practice this operation is basically the same as in real mode unless you do something like

```asm
%assign x 8 ; for example
call(32) far _pop + x * 0x10000
_pop: pop ax
pop ax
```

copy:
> thus making eip increment more than the block's size, they're infinitely many ways for something like that to happen, so you can't really tell ahead if a wrap could happen...

All instructions that can change cs (far jumps/calls/returns, interrupts) already bail out to interpreted mode

> I mean, you could theoretically check explicitly for jmp call ret mov cs, pop cs and such, but ultimately because we're dealing with a von-neuman architecture and the instructions themselves are prune to changing you can't really check it beforehand (unless you can solve the halting problem)

v86 already detects self-modifying code. It currently only logs, and it's not really used by OSes we run, but it could be implemented properly:

https://github.com/copy/v86/blob/e37189a4ad6ce4138e7168508f07553d0d3b6b3f/src/rust/cpu/cpu.rs#L3583

> Wont just using get_seg_cs directly in the eip incremation function instead of taking a cs argument solves this?

The code we're talking about ([cpu_context.rs](https://github.com/copy/v86/pull/1561/files#diff-e854ab98d9091d876fcc9472ee663550dd27cd559af0da08f14df831eb9031a1)) reads the instructions that are going to be compiled once, and then code is generated for them. Later changes to the CS offset can't affect the already generated code.

donno2048:
> All instructions that can change cs (far jumps/calls/returns, interrupts) already bail out to interpreted mode

What about

```asm
pop cs
mov cs, ax ; for example
loadall
```
?

EDIT: Actually, it doesn't matter, `pop cs` is not supported on i386 (and iirc since the 186, and I don't think V86 is looking to specifically emulate the 8086), `loadall` is not supported (I think) on V86 and as for `mov cs, r16`:
https://github.com/copy/v86/blob/e37189a4ad6ce4138e7168508f07553d0d3b6b3f/gen/x86_table.js#L114


> v86 already detects self-modifying code

It only detects self-modifying in the same page, should ideally mark any written block and if it's being jumped into (or just "entered") bail to interpreted mode.

self-modifying code is not the only problem, you also need to deal with misaligned jumps , etc

Can solve this by adding something like `instruction_start_markers` to the `JitContext` and mark each instruction start, then if a jump address is not aligned to a start marker bail to interpreted mode...

Edit: Actually, if jumps and calls bail to interpreted mode we don't need to handle it, just need to handle self-modifying the block or entering a modified block (without jumping).

But okay, when I'll have time I'll make a naive block escaping mechanism 


P.S. There's no need to check for 32-bit jumps in 16-bit:

https://github.com/copy/v86/blob/e37189a4ad6ce4138e7168508f07553d0d3b6b3f/src/rust/cpu/cpu.rs#L3086-L3089
https://github.com/copy/v86/blob/e37189a4ad6ce4138e7168508f07553d0d3b6b3f/src/rust/cpu/instructions.rs#L571-L582

donno2048:
Since the self-modifying code problem is separate from the `eip` wrapping problem I'll open a separate issue for it.

Edit: #1575

donno2048:
@copy can you review?

donno2048 (commented):
I wonder why I didn't do that in the first place 8fd41fd

donno2048:
@copy

donno2048:
@copy

PR #1562, opened 2026-05-16
ne2k: forces mac address translation even when machine loaded from state

internet-exploder:
hey man great software, thank you. I've noticed when I'm loading win2k from state, then detected mac addr outside of the vm is different from what it really is inside the VM.

I was checking like this:

emu.v86.cpu.devices.net.mac
emu.v86.cpu.devices.net.mac_address_in_state

they were both different from what was inside win2k vm loaded from a state.

This PR forces any mac to be translated.

### DIFF
--- a/src/ne2k.js
+++ b/src/ne2k.js
@@ -428,9 +428,21 @@ export function Ne2k(cpu, bus, preserve_mac_from_state_image, mac_address_transl
                 dump_packet(data, "send");
             }
 
-            if(this.mac_address_in_state)
+            if(this.mac_address_translation)
             {
                 data = new Uint8Array(data); // make a copy
+
+                if(!this.mac_address_in_state)
+                {
+                    this.mac_address_in_state = new Uint8Array(6);
+                }
+
+                // Forcefully learn the guest mac from the ethernet source field (bytes 6-11)
+                for(var i = 0; i < 6; i++)
+                {
+                    this.mac_address_in_state[i] = data[6 + i];
+                }
+
                 translate_mac_address(data, this.mac_address_in_state, this.mac);
             }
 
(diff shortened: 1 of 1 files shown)


copy:
> they were both different from what was inside win2k vm loaded from a state.

I cannot reproduce this, can you explain how to reproduce it and what is being fixed by this?

internet-exploder:
1. regular win2k image (I installed using qemu, then made it "Standard PC" type)
2. ne2k with mac addr translation enabled when creating vm with v86
3. save state of the machine (snapshot thing)
4. boot machine from state -> check mac addr that being set up for interface outside of the vm, check mac addr inside vm (it would be the same from previous boot, because now we are loading from state), but the "outside" interface mac would be different this time.

Actually the way I caught this: I made custom relay server which responds to DHCP requests and logs new mac addresses and their leases to stdout, and booting from same state in different browsers was giving same mac address to the relay side, which was not what I wanted.

So I figured out when mac addr translation is enabled it seems to rewrite a known mac addr to randomly generated one, however "known" mac addr picked from outside ne2k interface was wrong (in guest vm it was old mac from the snapshot state).

The solution I've found is to forcefully rewrite any mac address that comes out of the VM. 

In theory that may be not very good if we're running router or nested vm, but in that case people won't be using mac addr translation feature anyway!

A bit more clarification: in my initial post I mention

emu.v86.cpu.devices.net.mac
emu.v86.cpu.devices.net.mac_address_in_state

emu is emulation instance, I made it global like so: window.emu = new V86(...). Then put those two lines into inspector js console and it shows those macs. 

Inside the VM I do "ipconfig /all" to see mac addr which was different from both of them.

PR #1565, opened 2026-05-18
[sb16] correctly handle DSP commands that raise interrupts

tmenari:
Correctly handle Sound Blaster DSP commands that raise the interrupt.

This now means the interrupt can get cleared by later DSP commands and allows the Sound Blaster to be detected and work in e.g. Windows 3.1.


copy:
Thanks!

PR #1566, opened 2026-05-18
[devcontainer] set devcontainer build platform

tmenari:
By forcing the devcontainer to build a linux/amd64 image, this allows to be able to build the project on macOS Apple Silicon under Podman Desktop (applehv machine type) or Docker Desktop w/ Rosetta 2.

Additionally remove the curl feature as `curl` is installed by the Dockerfile already, and add the lock file to .gitignore.

NB: the reason it doesn't build on arm64 is because of the dependency on libc-i386 for tests - it may be possible to replace this with a full cross-compiler chain.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -4,10 +4,12 @@
 		"dockerfile": "Dockerfile",
 		"args": {
 			"VARIANT": "ubuntu"
-		}
+		},
+		"options": [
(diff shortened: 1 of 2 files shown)


copy:
> NB: the reason it doesn't build on arm64 is because of the dependency on libc-i386 for tests - it may be possible to replace this with a full cross-compiler chain.

This is only needed for tests, might be better to skip those on arm

PR #1573, opened 2026-06-02
fix: binary-safe HTTP body buffering for multi-segment POST/PUT

muralidhar-challa:
When a POST or PUT request body spans multiple TCP segments (~1460 bytes each), only the first segment's body bytes were used — subsequent segments were silently dropped because `this.read` was cleared after the first `\r\n\r\n` delimiter match, and no more headers trigger re-processing.

**Fix:**
- Buffer incoming TCP data as raw bytes (`Uint8Array`) instead of text-decoding everything. Search for `\r\n\r\n` in binary, text-decode only the header portion. Body stays as binary.
- For PUT/POST: if Content-Length is present and the body is incomplete, store the partial body in `this._xb` with a deferred callback. Subsequent segments accumulate until Content-Length is satisfied, then dispatch.
- Replaced ad-hoc `new TextEncoder()`/`new TextDecoder()` with module-scoped singletons.
- Added comment noting Transfer-Encoding: chunked is unsupported.
- Extracted fetch+response logic into standalone `dispatch_fetch(conn, ...)` to avoid `this` binding issues.

**Tests:**
- 5 unit tests in `tests/devices/fetch_network_post.js` (small POST, large split POST, binary body with embedded CRLFCRLF, GET, split headers)
- 1 integration test in `tests/devices/fetch_network.js` (POSTs 3000-byte body from buildroot guest, verifies full body arrives)

Build: 0 errors, 0 warnings. ESLint: clean.


src/browser/fetch_network.js
```diff
@@ -129,57 +146,84 @@ async function on_data_http(data)
...
         };
         if(["put", "post"].indexOf(opts.method.toLowerCase()) !== -1) {
+            // Check if the body might be split across multiple TCP segments.
+            // If Content-Length is present and larger than what we have,
+            // buffer and wait for the remaining chunks.
+            const content_length = parseInt(req_headers.get("content-length") || "0", 10);
+            const body_bytes = (data instanceof Uint8Array) ? data : new TextEncoder().encode(data);
+            if(content_length > 0 && body_bytes.length < content_length) {
+                const fetch_url = this.net.cors_proxy ? this.net.cors_proxy + encodeURIComponent(target.href) : target.href;
+                this._xb = {
```
copy: We don't usually use `_` prefix in this project.

src/browser/fetch_network.js
```diff
@@ -13,6 +13,28 @@ import {
 // For Types Only
 import { BusConnector } from "../bus.js";
 
+// Module-scoped encoder/decoder singletons (avoids repeated allocations).
+const textEncoder = new TextEncoder();
+const textDecoder = new TextDecoder();
```
copy: I don't think these need to be reused

src/browser/fetch_network.js
```diff
@@ -13,6 +13,28 @@ import {
...
+    for(let i = 0; i <= haystack.length - needle.length; i++)
+    {
+        for(let j = 0; j < needle.length; j++)
+        {
+            if(haystack[i + j] !== needle[j]) continue outer;
+        }
+        return i;
+    }
+    return -1;
+}
```
copy, line 30:
```javascript
function find_crlfcrlf(haystack)
{
    for (let i = 0; i + 3 < haystack.length; i++)
    {
        if (haystack[i] === CR && haystack[i + 1] === LF && haystack[i + 2] === CR && haystack[i + 3] === LF)
        {
            return i;
        }
    }
    return -1;
}
```

muralidhar-challa: Done — replaced `indexOfBytes` with `findCrLfCrLf` using `CR`/`LF` constants as  suggested. Simpler for this single use case.

src/browser/fetch_network.js
muralidhar-challa: Fixed — renamed all underscore-prefixed identifiers throughout:
> - `_xb` → `pendingBody`
> - `_raw` → `rawBuffer

src/browser/fetch_network.js
muralidhar-challa: Fair point, reverted back to inline `new TextEncoder()` / `new TextDecoder()`.

tests/devices/fetch_network_post.js
```diff
@@ -0,0 +1,189 @@
+#!/usr/bin/env node
```
copy, line 1: Let's add this test to devices-test.

copy:
Happy to merge if CI is green and I get a thumbs up from @basicer @chschnell, @SuperMaxusa or @ProgrammerIn-wonderland

tests/devices/fetch_network_post.js
muralidhar-challa: Added to devices-test.

muralidhar-challa:
@copy  Added the test to `make devices-test` and fixed all ESLint `id-match` warnings (snake_case). CI should go green now.

copy:
Thanks!

PR #1576, opened 2026-06-04
Fix graphical vga mode scaling on loading states

JoeOsborn:
I encountered an issue similar to the one described in [PR 1123](https://github.com/copy/v86/pull/1123#discussion_r1784364390). While using VGA text mode, if I made  a save state during text mode and started a fullscreen game that was at a different resolution, then when I loaded the state the canvas size (notably its actual size, not its style) would still be set to a smaller value.  This is because the number of rows and columns did not actually change, so the text mode display was not refreshed.  I fixed this by adding a `force` parameter to the involved functions to give a way to ensure that the settings were refreshed even if it seemed locally to the screen code like a no-op.

### DIFF
--- a/src/browser/dummy_screen.js
+++ b/src/browser/dummy_screen.js
@@ -48,6 +48,12 @@ export function DummyScreenAdapter(options)
     this.pause = function() {};
     this.continue = function() {};
 
+    this.clear_text_state = function()
+    {
+        text_mode_width = null;
+        text_mode_height = null;
+    };
+
     this.set_mode = function(graphical)
     {
         is_graphical = graphical;
--- a/src/browser/screen.js
+++ b/src/browser/screen.js
@@ -337,7 +337,7 @@ export function ScreenAdapter(options, screen_fill_buffer)
 
         // initialize display mode and size to 80x25 text with 9x16 font
         this.set_mode(false);
-        this.set_size_text(80, 25);
+        this.set_size_text(80, 25, true);
         if(mode === MODE_GRAPHICAL_TEXT)
         {
             this.set_size_graphical(720, 400, 720, 400);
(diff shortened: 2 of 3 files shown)


copy:
Threading a boolean flag through a bunch of functions doesn't look like a very readable solution. Can you try to find something nicer?

JoeOsborn:
Would you prefer having restore first call `set_size_text`/ `set_size_graphical` with zeroes or ones, and then call them shortly afterwards again with the real values?  Or some explicit `reset` method on `Screen` to be called at the beginning of a restore?  Another idea is to remove the optimizations where `Screen` checks that the new values differ from the old ones.

JoeOsborn:
I just filed a different bug about graphical text mode, #1579 , and if it is related to initialization somehow maybe the solution is connected to this issue also (but the patch in this PR does not fix it on its own).  @chschnell any thoughts?  Sorry for the double-ping but I realized you weren't addressed on this PR.

copy:
> Would you prefer having restore first call set_size_text/ set_size_graphical with zeroes or ones, and then call them shortly afterwards again with the real values? Or some explicit reset method on Screen to be called at the beginning of a restore? Another idea is to remove the optimizations where Screen checks that the new values differ from the old ones.

It seems that `set_size_graphical_text` and `set_size_text` do things other than just changing the size, so I'd move their body to another function that does all the actual invalidation, and then the `set_size` functions only implement the check, change the size and call the invalidation function.

JoeOsborn:
So I did something slightly different, instead adding a method to invalidate the screen's internal state.  I tried to do as you suggested first but it looked like I'd need to duplicate some logic around what exact calls needed to be made during `set_state`.  Instead this roughly matches how the graphical state is invalidated during a state load in graphical mode.

JoeOsborn:
Oh, and I'm also happy to combine these into one commit and force push a cleaner history after review but I wanted to give you a chance to review commit-by-commit first.

copy:
Thanks! Squashed and merged.

PR #1584, opened 2026-06-10
Disable browser's context menu on screen canvas

chschnell:
When absolute mouse positioning is enabled we no longer lock the mouse pointer to the screen canvas. This leads to the browser's context menu now popping up when right-clicking the canvas, hiding the guest's context menu underneath.

This PR fixes that by disabling the browser's context menu entirely on the canvas. I figured that we don't ever want to see it on the canvas (dosen't matter if relative or absolute positioning is enabled), is that ok or are there any exceptions I should consider?

### DIFF
--- a/src/browser/screen.js
+++ b/src/browser/screen.js
@@ -39,6 +39,7 @@ export function ScreenAdapter(options, screen_fill_buffer)
         graphic_screen = document.createElement("canvas");
         screen_container.appendChild(graphic_screen);
     }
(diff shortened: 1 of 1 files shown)


SuperMaxusa:
Wasn't this fixed in the `wip` branch? https://github.com/copy/v86/commit/7919dfbc815051501c1422da89af248e67f840c7

chschnell:
I didn't know, I'll close this then.

PR #1585, opened 2026-06-10
Fix multi-page JIT module invalidation

blixt:
When running Windows 98 I would get occasional blue screens, but only when JIT was enabled. I tracked it down (disclosure: with the help of Claude Fable 5, but this PR has been tested and written by me) to cases where a JIT module was spread across more than one page.

This PR fixes the issue by tracking all pages and cleaning up appropriately.

There's also a drive-by fix for a line that probably should have been targeting the virtual page slot, not the physical slot, but happy to remove that if it's wrong.

### DIFF
--- a/src/rust/jit.rs
+++ b/src/rust/jit.rs
@@ -137,6 +137,12 @@ struct JitState {
     pages: HashMap<Page, PageInfo>,
     wasm_table_index_free_list: Vec<WasmTableIndex>,
     compiling: Option<(WasmTableIndex, CompilingPageState)>,
+
+    // pages each module was compiled from + reverse index. used to clean up JIT
+    // code from written pages: `pages` alone loses track of multi-page modules
+    // when a sibling page is invalidated, which can leave stale modules executable
+    module_coverage: HashMap<WasmTableIndex, HashSet<Page>>,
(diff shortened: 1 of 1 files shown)


src/rust/jit.rs
```diff
@@ -1139,7 +1157,8 @@ pub fn update_tlb_code(virt_page: Page, phys_page: Page) {
             state_flags,
             hidden_wasm_table_indices: _,
         }) => set_tlb_code(virt_page, *wasm_table_index, entry_points, *state_flags),
-        None => cpu::clear_tlb_code(phys_page.to_u32() as i32),
+        // Note: tlb_code is indexed by virtual page
+        None => cpu::clear_tlb_code(virt_page.to_u32() as i32),
```
blixt, line 1160: Drive-by fix: this seems like it should use `virt_page`, not `phys_page`

copy:
I recently started looking in this independently, and intend to propose a different fix for this.

The drive-by fix looks good, feel free to only fix that and remove the other commit.

copy:
Follow-up: 14c822342cae169d07bcc20ff2b8182b9652a48c a8486f1c45daae48c0d952589b770a8b881decd8 928e6a6fc04691c627b053128a8d3079914cefb1

Issue #1556, opened 2026-05-10
Windows 9x CD-ROM device support fails with disks >= 256 MB


chschnell:
While trying to build a new Win95 image for v86 I noticed that the ATAPI CD-ROM support doesn't work.

As soon as the CD-ROM device is accessed I get blue-screens (recoverable) under Windows 95 like these:

<details>
<summary>Windows 95 example blue-screen screebshots</summary>

<img width="720" height="400" alt="Image" src="https://github.com/user-attachments/assets/419f9970-7720-4274-9496-ad71cb95ea19" />

<hr>

<img width="720" height="400" alt="Image" src="https://github.com/user-attachments/assets/d6ca9fbb-73fb-403c-b4a5-e944e5b9c13d" />

</details>

Windows 98 fails silently without blue-screens, but in both cases the D: folder is completely empty.

I tried using:

* the NEC CD-ROM driver from the Windows 95 boot floppy
* the OAK CD-ROM driver from the Windows 98 boot floppy
* the built-in CD-ROM driver of Windows 95 (which works under qemu, never tried it under v86)

Nothing helped, so I ran some more tests:

* happens both at the v86 main site and with a local build of current v86 HEAD
* happens both under SeaBIOS and Boch's BIOS
* happens both in v86 debug- and release-mode
* happens both using the Win95 and the Win98 image at the v86 main site (Win2K works)
* happens using my own Win95 image with OAK CD-ROM driver, from [here](https://github.com/copy/v86/pull/1325#issuecomment-2906349028) where I posted about having it working

I tried to fall back to the v86 repository state #1345 right when ATAPI CD-ROM support should be working:

```
git checkout d9088888e477663f0bf627a3e262d00af2849109
```

The problem was STILL there, that makes me doubt whether the released Win9x ATAPI support ever worked (what puzzles me is that I know that I had it working when I posted about it but now I can't reproduce it).

I can't remember if or when I last tested this after I had posted about it. Does it work for anybody else? Did it ever?

To summarize, CD-ROM support for Windows 9x appears currently broken, and it looks to me like it never worked.

**EDIT:** What does work is to boot from any boot floppy (Windows 95, Windows 98 or patcher9x) with a mounted CD-ROM image, I can access the CD just fine and run software from it.

SuperMaxusa:
I think it might be caused by Autorun or something like that.
When I insert an empty CD, created in UltraISO - it opens properly, but with a Windows 95 Install CD - it crashes with VxD blue-screen.

[...]

PR #1586, opened 2026-06-11
Support large CD-ROMs in Windows 9x (MSF)

blixt:
I had issues mounting a 300 MB ISO in Windows 98, and after a bit of digging it turned out it was because Windows wants the leadout in minutes:seconds:frames (MSF) format but we were providing it as blocks, so the larger the image the more wrong it was, causing misbehavior from the OS. (Sidenote: Windows NT/2000 will work with LBA and does not have this bug.)

For reference [QEMU also does this conversion](https://github.com/qemu/qemu/blob/master/hw/ide/atapi.c#L50-L56).

This should improve #1556, but I'm not sure if it fixes all the issues there.

As a "bonus" this PR also makes the device report "not present" when remounting an image since this seems to improve Windows 98 behavior.

### DIFF
--- a/src/ide.js
+++ b/src/ide.js
@@ -239,6 +239,11 @@ const ATAPI_ASC_INV_FIELD_IN_CMD_PACKET = 0x24;
 const ATAPI_ASC_MEDIUM_MAY_HAVE_CHANGED = 0x28;
 const ATAPI_ASC_MEDIUM_NOT_PRESENT = 0x3A;
 
+// Pending media-change state, reported to the guest in order on a swap.
+const MEDIUM_CHANGED_NONE = 0;
+const MEDIUM_CHANGED_UNIT_ATTENTION = 1;
+const MEDIUM_CHANGED_NOT_PRESENT = 2;
+
 // Debug log detail bits (internal to this module)
 const LOG_DETAIL_NONE = 0x00;   // disable debug logging of details
 const LOG_DETAIL_REG_IO = 0x01; // log register read/write access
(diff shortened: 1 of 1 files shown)


chschnell:
I tried this branch with Windows95b and it doesn't fix the problem described in #1556.

Could you please upload a short video demonstrating that you have a CD-ROM image > 255M working under Windows 98 (or is it 98 SE)?

We had already tried MSF encoding in #1556, and it didn't fix it for Win95 (I still have a branch [atapi-toc-cmd](https://github.com/chschnell/v86/tree/atapi-toc-cmd)). So it DOES fix it under Windows 98?

blixt:
I tested many times with mounting the Office 97 ISO for Windows 98 (using your disk/state snapshot) and every time the OS didn't seem to recognize even the name of the disk and  just became unresponsive when I tried to open D:, until I made the changes in this PR.

I have a some minor aesthetic custom HTML/CSS/JS on top of v86 running with the code in this PR in this recording:

https://github.com/user-attachments/assets/191f86d2-85ec-495a-ab89-724ad9507565

The mount code I run is this:

```typescript
async function mountCdrom(file: File) {
  const buffer = await file.arrayBuffer();
  emulator.eject_cdrom();
  await delay(1500);
  await emulator.set_cdrom({ buffer } as unknown as V86Image);
}
```

(It's an `arrayBuffer()` because I had some issues with lazy reads.)

blixt:
And yes this is Windows 98 SE:
image

copy:
Happy to merge if @chschnell gives this a green checkbox.

chschnell:
Yes, this looks good to merge!

copy:
(CI needs to be green though)

copy:
I pushed a small fix, besides that it looks good. Thanks!

PR #1588, opened 2026-06-11
Report signal processor chip as absent for virtual SoundBlaster 16

blixt:
Mounting the Microsoft Office 97 disk on Windows 98 SE and closing its install wizard (which plays a sound) consistently crashes EXPLORER.EXE with:

Screenshot 2026-06-11 at 16 29 03

I tracked this down to the virtual SoundBlaster 16 audio device saying it supported CSP (advanced/creative signal processor) when it doesn't. Since CSP is not required to play sound, this PR just reports it as absent. I verified the same flow no longer crashes.

### DIFF
--- a/src/sb16.js
+++ b/src/sb16.js
@@ -784,11 +784,28 @@ function any_first_digit(base)
     return commands;
 }
 
-// ASP set register
-register_dsp_command([0x0E], 2, function()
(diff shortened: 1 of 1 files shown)


copy:
I wasn't able to reproduce this, but it looks relatively safe so good merge. Thanks!

PR #1589, opened 2026-06-14
vmware: add legacy text-clipboard backdoor commands (6-9)

felixrieseberg:
Just like the other VMWare-related PRs, no hard feelings if you don't want them - I just want to upstream stuff I've done for Windows. This one implements the four legacy text-clipboard backdoor commands on VMware port 0x5658:

| Cmd | Name | Direction | Description |
|-----|------|-----------|-------------|
| 6 | `GETSELLENGTH` | host→guest | Returns byte-length of staged text, `0xFFFFFFFF` if nothing fresh |
| 7 | `GETNEXTPIECE` | host→guest | Reads next 4 bytes of staged text (LE dword) |
| 8 | `SETSELLENGTH` | guest→host | Allocates a receive buffer for an incoming transfer |
| 9 | `SETNEXTPIECE` | guest→host | Writes next 4 bytes into the receive buffer; emits `vmware-clipboard-guest` when full |

Bus events added:
- **`vmware-clipboard-host`** – caller stages a `Uint8Array`; the device makes it readable via GETSELLENGTH/GETNEXTPIECE
- **`vmware-clipboard-guest`** – fired when the guest finishes writing, carries the received `Uint8Array`

This is the pre-RPC clipboard protocol that open-vm-tools and its predecessors use. Because the guest reads and writes directly via `IN`/`OUT` instructions on port 0x5658, a minimal Win9x agent can sync `CF_TEXT` with the host from ring 3 without any kernel driver — just a `MOV EAX, 564D5868h` / `IN EAX, DX` loop.

The command numbers and wire format match VMware Workstation's implementation as documented in the open-vm-tools source (`lib/backdoor/backdoorGcc32.c`, `vmtools/src/vmbackdoor.c`).

## Implementation notes

- `CLIP_MAX` (64 KiB) caps both directions to prevent unbounded allocation.
- `clip_out_fresh` is a one-shot flag: `GETSELLENGTH` arms it (sets `fresh = false`) so a second poll before new data arrives returns `0xFFFFFFFF`.
- On the guest→host side, `SETNEXTPIECE` auto-fires `vmware-clipboard-guest` once the cursor reaches the declared length. A zero-length `SETSELLENGTH` fires immediately.
- Clipboard state is ephemeral and not included in `get_state`/`set_state`; a save/restore discards any in-flight transfer, matching the behaviour of real VMware Tools.

### DIFF
--- a/src/vmware.js
+++ b/src/vmware.js
@@ -8,6 +8,10 @@ import { BusConnector } from "./bus.js";
 const VMWARE_PORT = 0x5658;
 const VMWARE_MAGIC = 0x564D5868;
 
+const CMD_GETSELLENGTH = 6;
+const CMD_GETNEXTPIECE = 7;
+const CMD_SETSELLENGTH = 8;
+const CMD_SETNEXTPIECE = 9;
 const CMD_GETVERSION = 10;
 const CMD_ABSPOINTER_DATA = 39;
 const CMD_ABSPOINTER_STATUS = 40;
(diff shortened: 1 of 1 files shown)


chschnell:
Nice PR (same goes for #1542)!

Do you have some experience with vmware backdoor support for older guests? I was looking for Windows and found these:

* https://archive.org/details/winPreVista
* https://archive.org/details/VMwareToolsforWin95RTM_OSR1

I have the absolute mouse positioning working under Windows 3.1, I wonder if I could get the clipboard working, too.

copy:
Could you show the code to integrate this with the clipboard api?

copy:
This could use some documentation and integration into the UI, but since it's clearly correct and to avoid stale PRs, let's merge this.

Thanks!

PR #1590, opened 2026-06-14
vmware: add GETTIME backdoor command (23)

felixrieseberg:
Just like the other VMWare-related PRs, no hard feelings if you don't want them - I just want to upstream stuff I've done for Windows. This one implements implements VMware backdoor command 23 (`GETTIME`) on port 0x5658.

Return values in registers after the `IN EAX, DX`:

| Register | Value |
|----------|-------|
| EAX | Seconds since the Unix epoch (UTC), unsigned |
| EBX | Remaining microseconds (0–999 999) |
| ECX | Maximum time lag in microseconds (hardcoded 1 000 000) |
| EDX | Host's UTC offset in minutes, east positive |

Guests read the RTC only at boot. A VM resumed from a saved state continues running with whatever clock it was saved with, and the guest has no way to learn the correct time without rebooting.

With GETTIME, a small guest agent can issue a single backdoor call on a timer and compare its own `time()` against the host — no kernel driver, no full VMware Tools stack. The protocol is simple enough for a hand-rolled 16/32-bit DOS or Win9x TSR.

## Notes on GETTIMEFULL

VMware deprecated GETTIME in favour of command 35 (`GETTIMEFULL`) because EAX wraps as a *signed* value in 2038. Read as unsigned (which is what `>>> 0` gives JS, and what 32-bit C gives with `unsigned long`), EAX is valid until 2106. GETTIME is simpler for the guest to consume and is still the first thing open-vm-tools' `Backdoor_GetTime` wrapper tries. I think for our purposes her eis is probably fine :)

### DIFF
--- a/src/vmware.js
+++ b/src/vmware.js
@@ -9,6 +9,7 @@ const VMWARE_PORT = 0x5658;
 const VMWARE_MAGIC = 0x564D5868;
 
 const CMD_GETVERSION = 10;
+const CMD_GETTIME = 23;
 const CMD_ABSPOINTER_DATA = 39;
 const CMD_ABSPOINTER_STATUS = 40;
 const CMD_ABSPOINTER_COMMAND = 41;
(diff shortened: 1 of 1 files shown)


copy:
Thanks!

PR #1594, opened 2026-06-23
Add debian-xfce docker setup tools

wangyenshu:
This pull request adds tools for automated debian-xfce setup in two flavours: 9p and split raw disk. A web demo is https://wangyenshu.github.io/v86-debian-xfce-example/ with firefox, gimp, libreoffice ...

xfce desktop is launched automatically and saved in build-state process. Memory state is compressed then splitted into 20MB chunks, the logics for merging is in [debian-9p.html](https://github.com/wangyenshu/v86/blob/debian-example/tools/docker/debian/web-example/debian-9p.html) and [debian-raw-disk.html](https://github.com/wangyenshu/v86/blob/debian-example/tools/docker/debian/web-example/debian-raw-disk.html).

Remark: 
I think that split raw disk have better performance than 9p because the average size of files in 9p filesystem setup is much smaller than split raw disk chunks. And loading too much small files over https is inefficient. You can compare the speed of toggling xfce Application menu in https://wangyenshu.github.io/Scientific-Lab-in-Browser/ (9p filesystem) and https://wangyenshu.github.io/v86-debian-xfce-example/ (split raw disk). But even if using split raw disk, it takes about 10 minutes to launch firefox and the majority of time is used in loading small disk chunks over https. 

I am not a professional programmer. Please point it out if there is anything to be improved.


copy:
Thanks!

> I think that split raw disk have better performance than 9p because the average size of files in 9p filesystem setup is much smaller than split raw disk chunks.

Yeah, that's also my observation. It can be somewhat mitigated with a fast CDN.
You should also test chunk sizes bigger than 128 kB. I have found 512-1024 kB to be the sweet spot for modern OSes.

I pushed some improvements, in particular:

> Memory state is compressed then splitted into 20MB chunks

This didn't look very useful, so I removed it. It's an example anyway, so if it's needed any user can add it themselves.

PR #1597, opened 2026-06-23
Mask VGA DAC indices to fix colors on some demoscene demos

### LINKED ISSUE #1596: Palette not getting set for some demoscene demos
A lot of demoscene demos have the default palette in V86, while having a more appropriate palette in QEMU with the same VGA BIOS and same disk image:

Image
Image

floppy image (DOS boot disk with the affected demo above):
[demo.img.zip](https://github.com/user-attachments/files/29258606/demo.img.zip)

Observed with Chrome and the June 23rd, 2026 trunk. Running FreeDOS 1.3.
[console-log-palette-issue.txt](https://github.com/user-attachments/files/29258864/console-log-palette-issue.txt)

parkertomatoes:
Fixes https://github.com/copy/v86/issues/1596 by wrapping VGA DAC indices around at 255. 

I'm not sure any normal 13h application depends on this behavior, but I've noticed several tiny demoscene demos on Pouet triggering it. After wrapping the indices, colors match QEMU's output for me.

### DIFF
--- a/src/vga.js
+++ b/src/vga.js
@@ -1641,7 +1641,7 @@ VGAScreen.prototype.port3C7_read = function()
 
 VGAScreen.prototype.port3C8_write = function(index)
 {
-    this.dac_color_index_write = index * 3;
+    this.dac_color_index_write = (index & 0xFF) * 3;
     this.dac_state |= 0x3;
 };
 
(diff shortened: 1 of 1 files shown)


src/vga.js
```diff
@@ -1641,10 +1641,16 @@ VGAScreen.prototype.port3C7_read = function()
...
+    this.dac_color_index_write = (index & 0xFF) * 3;
     this.dac_state |= 0x3;
 };
 
+VGAScreen.prototype.port3C8_write16 = function(data)
+{
+    this.port3C8_write(data & 0xFF);
+    this.port3C9_write(data >> 8 & 0xFF);
+};
+
```
parkertomatoes, line 1653: Added a second corner case - this demo uses a 16-bit write on 0xC38 to write to both 0xC38 and 0xC39. 
https://parkertomatoes.github.io/demo-parade/?which=49438

copy:
Thanks!

PR #1600, opened 2026-06-24
Populate tag word after FST instruction

### LINKED ISSUE #1599: bad x87 FPU emulation causes DOS "3D effect" demo to render as a flat rectangle.
This demo uses floating point instructions to generate a 3D-ish effect:
https://www.youtube.com/watch?v=Y8rKpN_3Sj0
(apologies for the rude title, I didn't write it)
Image

V86 renders this demo as a flat rectangle:
Image

The source code for the demo isn't available, but from the disassembly, it's using floating point trigonometry  functions to warp the texture. V86 renders the texture (seemingly correctly) so everything but that floating point transformation is working.

Using latest trunk, Chrome. Floppy image and debug log is attached.

[demo-1843.img.zip](https://github.com/user-attachments/files/29296109/demo-1843.img.zip)
[demo-843-debug.log](https://github.com/user-attachments/files/29296110/demo-843-debug.log)

parkertomatoes:
Fixes https://github.com/copy/v86/issues/1599 .

The demo in question used FST to store a temporary variable, and FISTP to pop the value off the stack. The emulator copied the value after the FST instruction, but didn't update the tag word so nothing came back when it popped.

This commit updates "fpu_stack_empty" in the FST handler, which fixed the demo for me:
Screenshot 2026-06-24 at 7 11 54 PM
https://parkertomatoes.github.io/demo-parade/?which=1843 (again, apologies for the rude title)

Also added an llm-written regression test that uses FST/FISTP to put the FPU in a bad state if FST doesn't set the tag word.

### DIFF
--- a/src/rust/cpu/fpu.rs
+++ b/src/rust/cpu/fpu.rs
@@ -605,7 +605,13 @@ pub unsafe fn fpu_load_tag_word() -> i32 {
     return tag_word;
 }
 #[no_mangle]
-pub unsafe fn fpu_fst(r: i32) { fpu_write_st(*fpu_stack_ptr as i32 + r & 7, fpu_get_st0()); }
+pub unsafe fn fpu_fst(r: i32) {
+    let index: i32 = *fpu_stack_ptr as i32 + r & 7;
+    fpu_write_st(index, fpu_get_st0());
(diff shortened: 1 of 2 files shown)


copy:
Thanks!

PR #1602, opened 2026-06-26
Use rounded quotient for remainder for FPREM1

### LINKED ISSUE #1601: Demoscene demo "remnants" doesn't match QEMU (and PC) output
Testing another "demoscene" case - "remnants":
https://www.pouet.net/prod.php?which=96536

V86 renders something, but it fades to black instead of white, and the procedurally generated image looks different. QEMU isn't perfect but it's a lot closer to the demo's intended visuals. 

how it's supposed to look:
https://www.youtube.com/watch?v=2p5HCTUNgfU&t=535s

QEMU:
Image

V86:
Image
(boot floppy works on main v86 site but tinted red because of a recently fixed palette issue)

boot floppy with demo:
[demo-96536.img.zip](https://github.com/user-attachments/files/29361393/demo-96536.img.zip)

testing on Chrome in MacOS.
[copy.sh-1782432971099.log](https://github.com/user-attachments/files/29361642/copy.sh-1782432971099.log)

parkertomatoes:
Fixes https://github.com/copy/v86/issues/1601 .

Issue turned out to be FPREM1, which the demo uses for its graphics generation. The current implementation correctly rounds the quotient for FPREM1. But it uses an integer modulus for the remainder, effectively truncating the operands for both FPREM and FPREM1. This was enough to cause the effects in the demo. After applying this fix, it "remnants" renders correctly in V86 (albeit very slowly):

Screenshot 2026-06-25 at 8 02 19 PM

### DIFF
--- a/src/rust/cpu/fpu.rs
+++ b/src/rust/cpu/fpu.rs
@@ -498,9 +498,9 @@ pub unsafe fn fpu_fprem(ieee: bool) {
     let exp1 = st1.log2();
     let d = (exp0 - exp1).abs();
     if !intel_compatibility || d < F80::of_f64(f64::to_bits(64.0)) {
-        let fprem_quotient =
-            (if ieee { (st0 / st1).round() } else { (st0 / st1).trunc() }).to_i32();
-        fpu_write_st(*fpu_stack_ptr as i32, st0 % st1);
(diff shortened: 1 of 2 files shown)


copy:
Thanks!

copy:
> albeit very slowly

The demo mostly runs in interpreted (non-jit) mode. Most likely because the demo regularly writes to the code page.

PR #1603, opened 2026-06-30
fix: load task LDT before validating task CS

Yunus4832:
Fix potential panic in TSS hardware task switching support


copy:
Thanks!

Issue #1401, opened 2025-09-10
Large saved state files


alejandrocobo:
Hi. I've been a v86 user since 2021. One of my favorites projects!

I've been using the same version since then and now I've updated. Back then, a saved state file size of Windows 95 was around 40 Mb. You needed both, the hda image and the initial_state file. I see now you just need the saved state file, but its size is more or less the sum of the hda plus the former saved state file size; in my case is 280MB.

This makes way slower to start the virtual machine since I have to download a larger file. With the old version, I could enable async for the hda and I had to download just the saved state 40MB file while the rest was served on demand.

Is there a way to save just the state like is was done back then?

Thanks.

SuperMaxusa:
>  I see now you just need the saved state file

No, you still need an HDD image, the state file only contains the changed HDD sectors.

You can compress the state file using `zstd`, see https://github.com/copy/v86/discussions/1242#discussioncomment-12029670.

alejandrocobo:
Ok, since I cannot restore states (https://github.com/copy/v86/issues/1400) and the file size is the sum of both, I assumed otherwise.



> the state file only contains the changed HDD sectors.

That's weird. I save the state right after Windows 95 has started in both the old and the current v86 versions with very different results.

copy:
Async images should only store the sectors that were written to.

Try setting a breakpoint before saving the state here: https://github.com/copy/v86/blob/4251ac1a9e29b14d681475e802c39378c4c22626/src/buffer.js#L71
And here: https://github.com/copy/v86/blob/4251ac1a9e29b14d681475e802c39378c4c22626/src/buffer.js#L334

If the first breakpoint gets hit, it means you're accidentally loading the whole disk image.

If the second breakpoint gets hits, inspect the `block_cache` variable. Maybe Windows is touching more sectors than expected.

alejandrocobo:
> Async images should only store the sectors that were written to.

Well, the problem was between the chair and the keyboard. I was not using the async configuration while creating the state image. Now it works fine and, in fact, the state image is smaller than before.

This fixes https://github.com/copy/v86/issues/1400 too.

It works too when both are not-async.

Sorry guys.

trcrsired:
> Async images should only store the sectors that were written to.
> 
> Try setting a breakpoint before saving the state here:
> 
> [v86/src/buffer.js](https://github.com/copy/v86/blob/4251ac1a9e29b14d681475e802c39378c4c22626/src/buffer.js#L71)
> 
[...]

Issue #759, opened 2022-11-05
Feature Request


Atybot:
I request a feature for the node.js version for the modified hdd data to be able to save+load, that is store, on the physical maschine as a file. Is this possible to implement? Thank you.

5UP3R-D1N:
There is a way to do this, also doing it in buildroot is useless, i have tried it

So you might wanna look deep in the tutorial code of save and load

PR #1607, opened 2026-07-08
buffer: bound AsyncXHRBuffer block_cache with LRU eviction + write-back (opt-in)

semidark:
## Problem

`AsyncXHRBuffer`'s `block_cache` (used for `async: true` `hda`/`hdb`/`cdrom` images backed by a `url`) has no eviction of any kind. Every 256-byte block ever read or written is kept in `block_cache` (a `Map`) for the lifetime of the buffer instance — the only thing that ever clears it is `set_state()` during a full snapshot restore.

For a long-running host that backs a disk image with a real local file (e.g. a Node.js/Electron process, as opposed to a static/remote browser asset), this causes host memory usage to grow unboundedly with **total cumulative guest disk I/O**, not with the configured disk size or guest RAM. A guest that writes/rewrites a few hundred MB over its lifetime (e.g. syncing a project workspace, running builds) can balloon host RSS by that same amount, permanently, with no way to reclaim it short of restarting the buffer.

This came up debugging a real-world case: a v86-based sandbox app backing `hdb` with a writable local file saw host RSS climb from ~550MB to 2.5GB+ after syncing a single project workspace into the guest, entirely due to unbounded `block_cache` growth. #1401 shows the maintainer independently walking a user through `block_cache`-driven state-file bloat from a related angle, so this appears to be a known-but-previously-unaddressed pressure point in this file.

## Fix

Add an **opt-in** `max_cache_bytes` constructor/config option, via `buffer_from_object({ ..., max_cache_bytes })`. Left `undefined` (the default), behavior is **100% unchanged** — this matters for existing callers where `filename` may be a remote URL/browser `File` and writes have nowhere else to durably live except memory until `save_state()`.

When `max_cache_bytes` is set:
- `block_cache`'s `Map` iteration order is used as LRU order: every cache hit deletes+re-inserts the entry, so the front of the map is always the least-recently-used block.
- Once the cache exceeds the cap, a background eviction pass runs: clean (never-written) entries are dropped for free; dirty entries are first **written back to the backing file** (new `write_file_ranges` in `lib.js`, Node-only — uses `fs.promises.open(filename, "r+")` + positional `write()`) so a later cache miss reads correct data instead of stale bytes from an otherwise never-updated backing file.
- Dirty writes are **coalesced into contiguous ranges** before hitting the filesystem (`coalesce_writes`/`finish_run`): adjacent block indices are merged into a single `{ start, data }` write. This matters at scale — writing one 256-byte block per `fs.write` syscall does not scale when a single burst touches hundreds of thousands of blocks (e.g. syncing a large workspace); uncoalesced, flushing a 200MB write burst took ~22s in testing, versus well under a second once coalesced.
- A new `AsyncXHRBuffer.prototype.flush()` proactively writes back all dirty blocks (also coalesced) and drops the entire cache, independent of the size threshold — useful for reclaiming memory at a known idle point (e.g. after a bulk write) instead of waiting for size-triggered eviction to catch up.
- `V86.prototype.flush_disks()` exposes `flush()` for `hda`/`hdb` (all four possible IDE interfaces) through the public API.

### Concurrency

At most one eviction/flush write batch runs per buffer at a time. A block that gets re-dirtied by a concurrent `set()` while its *previous* value is mid-write-back is correctly left cached/dirty afterward (`block_cache_is_write` is only cleared for an index once its in-flight write completes **and** nothing re-added that index to the dirty set in the meantime).

### `AsyncFileBuffer` / `AsyncXHRPartfileBuffer`

Both classes reuse `AsyncXHRBuffer.prototype.set`/`handle_read` via direct prototype assignment (the existing pattern also used for `get_from_cache`/`get_state`/`set_state`). Since `set`/`handle_read` now call the new `touch_cache_entry`/`maybe_schedule_eviction` methods, those had to be mirrored onto both classes' prototypes too, or calling `set()`/`handle_read()` on them throws. Neither class ever sets `max_cache_bytes` on its own instances (only `AsyncXHRBuffer`'s constructor accepts it via `buffer_from_object`), so these are safe no-ops there — see the comments at both prototype-assignment sites in `buffer.js`.

## Testing

Added `tests/api/bounded-disk-cache.js` — operates directly on `buffer.js` (no VM boot required, so it's fast and has no image/kernel dependency), wired into `make api-tests`. Covers:
- default (`max_cache_bytes` unset) behavior is unchanged: cache grows to hold every block.
- a bounded cache stays near its configured cap after writing well beyond it.
- every byte ever written is correctly readable **through the same buffer** after eviction (falls through to a corrected on-disk read).
- every byte ever written is correctly readable via a **brand new buffer instance** pointed at the same file after eviction + `flush()` — proving durability (data relocated to disk, not lost), not just in-memory cache consistency.
- a performance regression guard: writing + evicting + flushing an 8MB burst under a tight cache bound completes in well under 5s (empirically ~40ms), guarding against a regression back to per-block syscalls.
- `AsyncFileBuffer` and `AsyncXHRPartfileBuffer` `set()`/`handle_read()` don't crash (regression coverage for the prototype-sharing issue above).

Also verified:
- Full `make all` build (Closure Compiler + Rust/wasm32) succeeds with 0 errors/0 warnings.
- Manual load test: a 200MB sequential write + flush drops from ~22s to well under a second with coalescing; a fully scattered (no adjacency) 50MB/8MB-cap stress scenario still completes in well under a second, confirming graceful degradation when coalescing can't help.
- `eslint` clean on all changed/added files.

## Compatibility

No changes to existing public API signatures; `max_cache_bytes` is a new optional field on the `hda`/`hdb`/`cdrom` config objects and is `undefined`-safe throughout. This is purely additive and defaults to the current unbounded behavior for every existing caller.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -374,6 +374,7 @@ api-tests: build/v86-debug.wasm
 	./tests/api/reboot.js
 	#./tests/api/reboot-buildroot.js # https://github.com/copy/v86/issues/636
 	./tests/api/pic.js
+	./tests/api/bounded-disk-cache.js
 
 all-tests: eslint kvm-unit-test qemutests qemutests-release jitpagingtests api-tests nasmtests nasmtests-force-jit rust-test tests expect-tests
 	# Skipping:
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -900,6 +900,40 @@ V86.prototype.save_state = async function()
 };
 
 /**
+ * Flush any pending write-back for hda/hdb's disk-image block cache (see
+ * `max_cache_bytes` in the hda/hdb options and AsyncXHRBuffer.flush in
+ * buffer.js). Only meaningful for disks backed by a local, writable file
+ * (i.e. `{ url, async: true, max_cache_bytes }` on a non-browser/Node
+ * host) — a no-op otherwise, including if `max_cache_bytes` was never
+ * configured (in which case there is nothing pending to flush) or the
+ * drive has no attached buffer.
+ *
+ * Bounded caches also evict on their own as they grow past
+ * max_cache_bytes, so calling this is optional; it's useful to proactively
+ * reclaim memory at a known idle point (e.g. after a bulk write/hydration
+ * completes) instead of waiting for the next write to trigger eviction.
+ *
+ * @return {!Promise}
+ */
+V86.prototype.flush_disks = async function()
+{
+    const ide = this.v86 && this.v86.cpu.devices.ide;
+    if(!ide)
+    {
+        return;
+    }
+
+    const buffers = [
+        ide.primary?.master?.buffer,
+        ide.primary?.slave?.buffer,
+        ide.secondary?.master?.buffer,
+        ide.secondary?.slave?.buffer,
+    ].filter((buffer) => buffer && typeof buffer.flush === "function");
+
+    await Promise.all(buffers.map((buffer) => buffer.flush()));
+};
+
+/**
  * @return {number}
  * @ignore
  */
--- a/src/buffer.js
+++ b/src/buffer.js
@@ -1,6 +1,7 @@
 import { CPU } from "./cpu.js";
-import { load_file, get_file_size } from "./lib.js";
+import { load_file, get_file_size, write_file_ranges } from "./lib.js";
 import { dbg_assert, dbg_log } from "./log.js";
+import { LOG_DISK } from "./const.js";
 
 // The smallest size the emulated hardware can emit
 const BLOCK_SIZE = 256;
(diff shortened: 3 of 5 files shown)


copy:
Could you disclose to which degree generative AI was used in the writing of the code, commit messages and pull request?

semidark:
Sorry my fault. Should have been clearer on that in the first place.

This resulted from a test where I am trying to build an v86 and electron backed Linux sandbox. It was designed and implemented manly by fable and opus 5.8 running in a opencode harness. If this PR is against your contribution guidelines I am sorry and I will of course close it. Don't want to be part of this growing problem for Open source project like yours.

Thnx for the effort

copy:
No worries, let's close this.

Regarding the original issue, I think there are easier fixes:

- A flag to turn of the read cache (primarily targeted at node, where the OS has its own file cache)
- Writeback to the disk image for nodejs (#759) (plus a flag that disabled write cache)
- An API that inspects/clears the cache, so that embedders can do their own thing

PR #1609, opened 2026-07-10
Optionally simulate IDE read latency

### LINKED ISSUE #1608: Print Shop Deluxe 1.0 for DOS hangs at main menu, ok in QEMU and DOSBOX
Using latest V86, FreeDOS 1.2, Chrome. 

Print Shot Deluxe 1.0 for DOS consistently hangs in V86 at the same point, right when it hits the main menu. Interestingly, it consistently _doesn't_ hang if dev tools are open, but it will reproduce if you use debug WASM and leave the dev tools closed. And it works fine in other x86 emulators, including QEMU.

[psd.img.zip](https://github.com/user-attachments/files/29848466/psd.img.zip)
[debug.log](https://github.com/user-attachments/files/29848472/debug.log)
Image

parkertomatoes:
Fixes https://github.com/copy/v86/issues/1608, kind of.

After doing a LOT of tracing, I think I understand the problem at least. I got lucky because PSD.EXE leaks some symbol names through a table in an error formatter.

The hang I'm observing comes from an unchecked null pointer returned by an allocator function. Each allocation, it measures the time from start and tags a struct with it. After X allocations, it does a garbage collection and appears to use that measured age to prioritize. 

In the failing path, IDE.js loads data _so fast_ that this age tag is always zero. Having all the allocations be the same rank puts enough pressure on the garbage collection that it fails to reclaim any memory and returns a null pointer. The null pointer is unchecked, data is written to it, and the program hangs.

It's not 100% clear why this doesn't happen to QEMU. Both bochs and seabios reproduce the error on v86 and not on qemu. My best guess is v86 is just quicker about it when the disk is a memory buffer. 

This "fix" adds an option to put of a bit of latency on IDE reads if the backing buffer is synchronous. This spreads the allocations out enough to let the cleanup succeed. Most software works find without this unnecessary latency, though, so it's turned off by default.

### DIFF
--- a/src/browser/starter.js
+++ b/src/browser/starter.js
@@ -243,6 +243,7 @@ V86.prototype.continue_init = async function(emulator, options)
     settings.preserve_mac_from_state_image = options.preserve_mac_from_state_image;
     settings.mac_address_translation = options.mac_address_translation;
     settings.cpuid_level = options.cpuid_level;
+    settings.ide_read_async = !!options.ide_read_async;
     settings.virtio_balloon = options.virtio_balloon;
     settings.virtio_console = !!options.virtio_console;
 
--- a/src/cpu.js
+++ b/src/cpu.js
@@ -144,6 +144,7 @@ export function CPU(bus, wm, stop_idling)
 
     /** @type {!Object} */
     this.devices = {};
+    this.ide_read_async = false;
 
     this.instruction_pointer = view(Int32Array, memory, 556, 1);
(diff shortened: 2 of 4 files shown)


parkertomatoes:
I'm gonna close this one - it's not clearly fixing bad emulation, it's not slowing emulation down in any realistic way, and the footprint of creating an option is bigger than I'm comfortable with for this. Plus, I was able to work around the original issue.

PR #1613, opened 2026-07-23
Expose a parallel port

parkertomatoes:
This adds a "device" that only emits bus events for the LPT1 registers. The goal is to make it possible to add a virtual printer from outside the emulator. 

Parallel port logic is timing based, but this hacks around it by latching falling ACK events until the guest reads the status register. For both bochs and seabios that's enough to keep int 17h synchronized while the bus blasts status and control register updates at whatever speed.

I would have also included a printer, but the best library for parsing printer protocols, GhostPDL, has a copyleft license that isn't compatible with v86. So I made a separate GhostPDL-based PCL5 printer that can be wired to this interface but has an AGPL3 license:
https://github.com/parkertomatoes/inkjet-emu

And some demos with DOS and Windows 3.x printing with v86 and inkjet-emu:
https://inkjet-emu-demo.swannerfly42.workers.dev/


src/parallel.js
```diff
@@ -0,0 +1,185 @@
...
+
+    this.data = state[0];
+    this.status = state[1];
+    this.control = state[2];
+    this.status_latched = state[3];
+};
+
+// For Types Only
+import { CPU } from "./cpu.js";
+import { BusConnector } from "./bus.js";
```
SuperMaxusa: nit-pick: move this to the top of file

SuperMaxusa:
It would be good if you could create multiple LPT ports like in UART:

https://github.com/copy/v86/blob/2f1346b0e7d88d4cbbbcc05fe15b4e369c3de23f/src/cpu.js#L1184-L1195

parkertomatoes:
Thanks for the review. Added LPT2, but I don't think seabios or boch's support LPT3/LPT4.

tests/devices/parallel.js
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env node
```
copy: To be honest, I don't find this test very useful: It tests mostly internal implementation details, not whether this presents a working parallel port to the guest OS.

A more useful test would be to boot Linux or DOS, run some commands that interact with the parallel port and test that the corresponding events get called.

parkertomatoes: I see what you mean. test/api/floppy.js seems to do what you're asking so I tweaked it a bit to make a higher-level parallel port test.

copy:
Thank you!

PR #1622, opened 2026-08-21
Update vga.js

Godlance:
fix: prevent image freezes caused by svga_offset

This PR fixes a screen freeze in some svga dependant games. 

The SVGA 8bpp fast path in `screen_fill_buffer` ignores `svga_offset`, causing permanent display freezes with page-flipped double buffering.

`VGAScreen.prototype.screen_fill_buffer`'s fast path for 8bpp SVGA modes reads the displayed frame starting at byte `0` of `svga_memory` always, which ignores `svga_offset` that `port1CF_write` sets from `VBE_DISPI_INDEX_X_OFFSET`/`VBE_DISPI_INDEX_Y_OFFSET`.

This is the standard VBE "Set Display Start" call that real VESA-aware software uses for page-flipped double buffering, it renders the next frame into a back buffer elsewhere in `svga_memory` and then flips the display-start offset to it instead of copying pixels.

Since this fast path always paints from offset `0`, any program that flips the display start away from `0` gets a display that silently freezes on whatever was last drawn at offset `0` while the program keeps rendering correctly into the buffer v86 no longer looks at. 

Given that there is no error, it was really hard to debug from my side.

The existing code already has a comment to this fact:

```js
// XXX: Slow, should be ported to rust, but it doesn't have access to vga256_palette
// XXX: Doesn't take svga_offset into account
```

This PR fixes the second XXX.

I've tested it with Master of Orion 2 and display freeze is fully resolved.

This will help any 90s era game that uses page flipping.

Edit: Formatting.

### DIFF
--- a/src/vga.js
+++ b/src/vga.js
@@ -2536,10 +2536,14 @@ VGAScreen.prototype.screen_fill_buffer = function()
             // XXX: Doesn't take svga_offset into account
             const buffer = new Int32Array(this.cpu.wasm_memory.buffer, this.dest_buffet_offset, this.screen_width * this.screen_height);
             const svga_memory = new Uint8Array(this.cpu.wasm_memory.buffer, this.svga_memory.byteOffset, this.vga_memory_size);
+            // svga_offset selects which region of svga_memory is actually being shown, which programs use for double buffering / page flipping 
+            // This has to be applied here too to avoid the displayed image freezing on whatever was last written at offset 0 when a program flips away
+            // from it. (Master of Orion 2 640x480x256 VESA mode).
+            const base = this.svga_offset % this.vga_memory_size;
 
             for(var i = 0; i < buffer.length; i++)
             {
-                var color = this.vga256_palette[svga_memory[i]];
+                var color = this.vga256_palette[svga_memory[(base + i) % this.vga_memory_size]];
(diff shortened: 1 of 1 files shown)


copy:
From your description, it's not clear to me that this game actually uses the wraparound behavior. I have pushed an alternative fix that clamps instead (like the other bpps). Could you test that this fixed the issue?

Godlance:
> From your description, it's not clear to me that this game actually uses the wraparound behavior. I have pushed an alternative fix that clamps instead (like the other bpps). Could you test that this fixed the issue?

Yeah, sorry for taking so long. 

<del>I just tested the latest release (https://github.com/copy/v86/commit/132da7b6549911c77d95cb895e772b69a2251b29) and it still has the same problem. So, either MoO2 is doing something different altogether or the fix is breaking something else.</del>

<del>If I manage to find more games or other software of another kind that shows a similar behavior, we could find out what's going on. For now, I'll maintain the fix in my setup.</del>

<del>Edit: If I could find a MoO2 demo, I could also put up a website for you to test this behavior, but it seems I can only find full game releases.</del>

I was using a release version of libv86.js instead of compiling one with your changes, sorry. After compiling and testing I can assure that your fix works just as well, thank you!

copy:
Thanks for diagnosing the issue and testing!

PR #1625, opened 2026-08-24
boot menu option

SuperMaxusa:
Adds a boot menu option for seabios, might be useful for debug purposes.


copy:
Thanks!

(I reverted the bios file update, I don't think it's worth adding 400kB to every git clone for a minor update)

PR #1632, opened 2026-09-10
ide: do not raise an IRQ when ready to receive an ATAPI packet

jlxip:
[Information Technology - AT Attachment with Packet Interface - 6 (ATA/ATAPI-6)](https://pdos.csail.mit.edu/6.828/2018/readings/hardware/ATA-d1410r3a.pdf), `DP1: Receive_Packet State` says:

_When the device is ready to receive the command packet, [...] INTRQ is negated._

With asynchronous CD-ROM reads, the premature interrupt causes Windows 98 SE to reset the device before the requested data arrives, preventing Setup from copying the driver files from the CD in the last stage.


jlxip:
Sorry I'm an idiot

copy:
Thanks!

PR #1636, opened 2026-09-10
dma: Correct 16-bit channel addressing

jlxip:
Regarding 16-bit ISA DMA, osdev.org ([the link in the source](https://github.com/copy/v86/blob/d96be774e549a83371b038b86e819804c96b921f/src/dma.js#L371)) says:

_It is important to note that the upper bit of the internal address is *lost* when it is shifted left -- it is not ORed into the page register byte._

It's wrong, according to:
- [Intel 82801DB ICH4 Datasheet](https://www.intel.com/content/dam/www/public/us/en/documents/datasheets/82801db-io-controller-hub-4-datasheet.pdf), sections 5.4.3.1 and 9.2.1
- [IBM Personal Computer AT Technical Reference (March 1984)](https://www.minuszerodegrees.net/manuals/IBM_5170_Technical_Reference_1502243_MAR84.pdf), figure "Address Generation for DMA Channels 7 through 5"
- [Intel 82371AB (PIIX4) Datasheet](https://0x04.net/~mwk/doc/intel/82371ab.pdf), section 8.4.9.1 (actually the same table as ICH4)

Intel says:

_For transfers to/from a 16-bit slave (channel’s 5–7), the address is shifted left one bit location. Bit 15 will be shifted into Bit 16._

The page register's lowest bit must also be ignored. This PR fixes both things.

For reproducing the bug, I've prepared [this floppy](https://mega.nz/file/CdUi1Kxa#jBW-Z-wG72scqazjs056nBarE-fyFSeRSDNGb9ZQlKE). It has two files:
- `TEST.EXE`, extracted from [the latest doslib build](https://github.com/joncampbell123/doslib/releases/tag/doslib-20250729), `hw/sndsb/dos86l/test.exe`
- `RUN.BAT`, which runs `TEST.EXE` on `The Microsoft Sound.wav` (copies it first to C:)

You will hear the sound glitched before the fix, and fine after. This is only a minor inconvenience, but I was playing a game in which I had no sound at all before the fix.


copy:
Thanks!

